PR #13, opened 2018-01-16
fix faulty urls that were too deeply nested

jenweber:
Warning, trying to view the diff on GitHub might crash your browser.

I only found one type of url that had a line break between the markdown for `]` and `(`

I did a crazy find/replace to move the url nesting up a level. I expect some links to be broken because a few were working earlier, but I'll find them and fix them before removing the WIP label. Edit: I wrote a script to find the broken links and there were a gazillion. They are fixed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+node_modules
--- a/guides/v1.10.0/application/index.md
+++ b/guides/v1.10.0/application/index.md
@@ -18,10 +18,8 @@ What does creating an `Ember.Application` instance get you?
    `App.PostsView` and `App.PostsController`). This helps to prevent
    polluting the global scope.
 2. It adds event listeners to the document and is responsible for
-   delegating events to your views. (See [The View
-   Layer](../understanding-ember/the-view-layer)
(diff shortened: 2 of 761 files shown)


jenweber:
@mansona @locks can you try to run this locally using `guides-app` and spot check links/images? Make sure that you are using my work as the source and not what's deployed to npm. I believe I have fixed all but 15 links (and they are easy enough that someone else can do them in a later PR). 

Use npm link or make the following edit in guides-app to point to files on your hard drive:
```
const jsonTrees = versions.map((version) => new StaticSiteJson(`../guides-source/guides/${version}`, {
  contentFolder: `content/${version}`
}));
```

I also could use eyes on the link-checker file to make sure it's intelligible to other humans. Try running it too. We can decide what to do with it later...

jenweber:
Closed. Discovered that urls in production are  different than urls locally 😭  

Locally, md files are found at `v2.16.0/subgroup/content.md`

In production, they are at `v2.16.0/subgroup/content/index.md`

Where it appeared that links were too deeply nested, what was actually the problem was mishandled slashes. A correct link to a md file should have a trailing slash, like `v2.16.0/subgroup/content/`

PR #37, opened 2018-05-11
Standardize docs

sivakumar-kailasam:
This fixes a bunch of inconsistencies in our code blocks that needed to be fixed before we could attempt to make code highlighting work for code blocks with diffs. I've taken utmost care to review each commit made in this branch & its all done with help of IDE & custom code so it requires a quick glance at most from a reviewer :)


jenweber:
I skimmed all the diffs and checked through each rendered page of the tutorial for 3.1. This PR only changes the code block declarations and not the code inside them. lgtm

PR #39, opened 2018-05-11
Update index.md

ajcolter:
These links are broken on the site.


mansona:
Hey @ajcolter thanks for the contribution 🎉 we're currently in the process of deploying the new infrastructure for the Ember Guides and we should have it live in a few days.

As it turns out this isn't an issue on the new guides, you can see it in action on a temporary URL [here](https://ember-guides-prod.herokuapp.com/v3.0.0/testing/) and as you can see those links are working.

Thanks again for the contribution! If you want to help out some more you can reach us on the Community Slack channel in the #-team-learning channel 👋

PR #41, opened 2018-05-16
Improvements in code snippets for autocomplete-component tutorial

jessieay:
**WHY**: the existing code diff highlights were confusing when going
through the tutorial.

**WHAT**:
- update diff to highlight *all* added lines in each section
- add diff highlighting and more lines to snippet where `triggerKeyEvent` and `fillIn` are added so provide more context on where in the file this happens.
- update snippet where 2nd integration test for component was added to
make it clearer that it is an entirely new test. Previously just one line was highlighted but the entire test is new when you are in this step of the tutorial.

Pic of BEFORE:

screen shot 2018-05-16 at 3 41 29 pm

Pic of AFTER:

screen shot 2018-05-16 at 3 44 07 pm

### DIFF
--- a/guides/v3.1.0/tutorial/autocomplete-component.md
+++ b/guides/v3.1.0/tutorial/autocomplete-component.md
@@ -371,7 +371,7 @@ Since our component is expecting the filter process to be asynchronous, we retur
 
 Next, we'll add the call to render the component to show the cities we've provided above.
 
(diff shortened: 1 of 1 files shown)


jenweber:
This makes a lot of sense to me. Thank you @jessieay! Let us know if you spot anything else that could be done better.

Issue #34, opened 2018-05-09
Make the Controllers section more helpful


jenweber:
There's not much on https://guides.emberjs.com/current/controllers - they are not distinguishable from routes or components in this explanation.

Try to help someone new understand, “where and when should I use a Controller?”

You can base content off [this blog post](https://medium.com/@jenweber/what-are-controllers-used-for-in-ember-js-cba712bf3b3e?source=linkShare-bbfa056dba1d-1525867885). The relevant sections are “the essentials” and “when you should make a controller”. I wrote a lot about Query Parameters but we can just link to the Routing section of the guides instead.

The end result should be at maximum twice the current length.

jenweber:
@chrisrng is working on this.

I encourage you to open a PR once you have a first draft in order to get feedback. Thank you!

PR #46, opened 2018-05-20
Update Controller Index

chrisrng:
First draft in tackling #34 

[rendered](https://github.com/chrisrng/guides-source/blob/c32e24ae5a2e689e80ecfbd84c18e09c1dfc3cf1/guides/v3.1.0/controllers/index.md)

### DIFF
--- a/guides/v3.1.0/controllers/index.md
+++ b/guides/v3.1.0/controllers/index.md
@@ -1,27 +1,35 @@
-## Controllers
+### What is a Controller?
 
-Controllers behave like a specialized type of Component that is rendered by
-the router when entering a Route.
+A [Controller](https://emberjs.com/api/ember/release/classes/Controller) is routable object which receives a single property from the Route – `model` – which is the return value of the Route's [`model()`](https://emberjs.com/api/ember/3.1/classes/Route/methods?anchor=model) method.
 
-The controller receives a single property from the Route – `model` – which is
-the return value of the Route's `model()` method.
+The model is passed from the Route to the Controller by default using the [`setupController()`](https://www.emberjs.com/api/ember/3.1/classes/Route/methods/setupController?anchor=setupController) function. The Controller is then often used to decorate the model with display properties such as retrieving the full name from a name model.
 
-To define a Controller, run:
+A Controller is usually paired with an individual Route of the same name.
 
+### Defining a Controller
+
+We only need to generate a Controller file if we want to customize the properties or provide any actions to the Route. If we have no customizations, Ember will provide a default Controller instance for us at run time.
+
+To generate a controller, run
 ```bash
 ember generate controller my-controller-name
 ```
 
-The value of `my-controller-name` must match the name of the Route that renders
-it. So a Route named `blog-post` would have a matching Controller named
-`blog-post`.
+This creates a controller file at `app/controllers/my-controller-name.js`, and a unit test file at `tests/unit/controllers/my-controller-name-test.js`.
 
-You only need to generate a Controller if you want to customize its
-properties or provide any `actions`. If you have no customizations, Ember will
-provide a Controller instance for you at run time.
+The controller name `my-controller-name` must match the name of the Route that renders it. So if the Route is named `blog-post`, it should have a matching Controller named `blog-post`. The matching file names of the Controller and the Route signals to Ember that this Controller must be used when landing on the `blog-post` Route.
 
-Let's explore these concepts using an example of a route displaying a blog
-post. Presume a `BlogPost` model that is presented in a `blog-post` template.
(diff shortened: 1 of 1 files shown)


guides/v3.1.0/controllers/index.md
```diff
@@ -1,27 +1,31 @@
-## Controllers
+### What is a Controller?
 
-Controllers behave like a specialized type of Component that is rendered by
-the router when entering a Route.
+[Controllers](https://emberjs.com/api/ember/release/classes/Controller) are a routable object that is used to decorate the model returned by the router when entering a Route. A Controller is usually paired with an individual Route of the same name.
```
jenweber: Can you make this singular? A controller is a routable object...

guides/v3.1.0/controllers/index.md
```diff
@@ -1,27 +1,31 @@
...
 
-The controller receives a single property from the Route – `model` – which is
-the return value of the Route's `model()` method.
```
jenweber: I think we need to mention these two lines above somewhere explicitly

guides/v3.1.0/controllers/index.md
```diff
@@ -75,27 +86,16 @@ new `isExpanded` property is true.
...
 
-<div class="common-question">
-  <h4>Should I use controllers in my application? I've heard they're going away!</h4>
+###### Should we use controllers in my application? I've heard they're going away!
   
-  <p>Yes! Controllers are still an integral part of an Ember application architecture, and generated by the framework even if you don't declare a Controller module explicitly.</p>
-</div>
+Yes! Controllers are still an integral part of an Ember application architecture, and generated by the framework even if you don't declare a Controller module explicitly.
+
+###### When should we create a Controller?
```
jenweber: Can the content of this section be rolled into "When and where to use controllers"?

jenweber:
Thank you @chrisrng! I added some comments. I really like the the "where & when" section and the places you've simplified some of the language.

I am curious about this sentence:

> A controller is a routable object that is used to decorate the model returned by the router when entering a Route.

I don't know enough about controllers myself to know if this is accurate. Can you tell me more about it?

mansona:
@jenweber I quite like @chrisrng's line about decorating a model, but I would be a tiny bit concerned about the wording 🤔 

> is a routable object that is used to decorate the model 

that suggests that its only purpose is to decorate the model, which is not true. You also can make use of controllers if you're not using models in some cases, even if it's not recommended 😂 

@chrisrng how about: 

> A controller is a routable object that can be used to decorate the model returned by the router when entering a Route.

or

> A controller is a routable object that is mostly used to decorate the model returned by the router when entering a Route.

Again this might need a look over from someone who knows more about controllers than I do 😂

guides/v3.1.0/controllers/index.md
chrisrng: This is what I mean by routable :)

chrisrng:
@jenweber @mansona 

Rewrote the first part into the following:

> ### What is a Controller?
> A [Controller](https://emberjs.com/api/ember/release/classes/Controller) is routable object which receives a single property from the Route – `model` – which is the return value of the Route's `model()` method.
> 
> The model is passed from the Route to the Controller by default using the [setupController](https://www.emberjs.com/api/ember/3.1/classes/Route/methods/setupController?anchor=setupController) function. The Controller is then often used to decorate the model with display properties such as retrieving the full name from a name model.
> 
> A Controller is usually paired with an individual Route of the same name.

guides/v3.1.0/controllers/index.md
chrisrng: Thoughts on leaving this here? I think it might be quoted a bit and wanted it in the same place for people to see

chrisrng:
One more update to adhere to Contributing guide (fixed links and commit message)

[rendered](https://github.com/chrisrng/guides-source/blob/d58c64ffcc10ba6107d8d5e20fd3778484d04c70/guides/v3.1.0/controllers/index.md)

guides/v3.1.0/controllers/index.md
jenweber: I'm fine with either

jenweber:
Thank you for the edits! I want to get one more person to sign off on the "routable object" definition and we'll be good to merge :D I ran the app locally and everything renders great.

jenweber:
@mansona I have received one comment that "routable" is incorrect, and a few other people saying it's fine, so I asked in #dev-ember and didn't hear back yet.

I'm going to merge and if I get more clarification, we can correct it. It's related to the route, so it's routable, ok :P

jenweber:
Merged, thank you @chrisrng!

PR #49, opened 2018-05-21
v3.2.0: Added docs on let helper

kennethlarsen:


guides/v3.2.0/templates/built-in-helpers.md
```diff
@@ -36,3 +36,32 @@ format of a concatenated string.
 
 This will display the result of `this.get('foo.item1')` when index is 1,
 and `this.get('foo.item2')` when index is 2, etc.
+
+### Built-in block helpers
+Now let's say your template is starting to get a bit cluttered and you now want to clean up the logic in your templates. This can be achieved with the `let` block helper. The [`{{let}}`](#) helper lets you create new bindings in your template.
```
sivakumar-kailasam: @kennethlarsen assuming the link is set to `#` because it needs to be updated later? In that case can we create an issue for that? 

Or if you know what the link is going to be in api docs when it releases, then let's add it now.

kennethlarsen: @sivakumar-kailasam My guess is that it would be https://www.emberjs.com/api/ember/release/classes/Ember.Templates.helpers/methods/get?anchor=let but someone would need to confirm that for me.

sivakumar-kailasam: Seems right, https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/syntax/let.ts#L10. Can you please update it with the 
https://www.emberjs.com/api/ember/release/classes/Ember.Templates.helpers/methods/get?anchor=let so that we can merge this?

sivakumar-kailasam:
Thanks @kennethlarsen 👍

PR #58, opened 2018-05-26
Fixed component lifecycle anchor links

chrismou:
Anchor links to methods within the component lifecycle pages all contained an unneeded `-code` in the url

I'd guess from the consistently wrong urls that they were auto-generated from the anchor titles, which contained a word surrounded by backticks that looks to have been converted to the word `code`.

Fixed and tested from 2.30 through to 3.2.0


wayne-o:
👾

PR #59, opened 2018-05-27
Where to add attribute bindings

pablobm:
Previously submitted as https://github.com/emberjs/guides/pull/2132

The Guides currently advise to use `Component.reopen` to add attribute bindings to bundled components, such as `LinkComponent` and `TextField`. However this practise is now discouraged.

This PR updates this advice to recommend that users override these components instead.

Note that I have had to remove the text field example, as it requires overriding the undocumented `-text-field` component.

Here's a Twiddle with an example: https://ember-twiddle.com/0feff0577066c1414747aef04202aa9d?openFiles=templates.application.hbs%2C

### DIFF
--- a/guides/v3.1.0/templates/binding-element-attributes.md
+++ b/guides/v3.1.0/templates/binding-element-attributes.md
@@ -40,46 +40,37 @@ If `isAdministrator` is `false`, Handlebars will produce the following:
 
 ### Adding Other Attributes (Including Data Attributes)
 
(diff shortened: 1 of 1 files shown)


jenweber:
P.S. Sorry this took so long. Thank you for being persistent! 

This will show up in the live guides the next time we update the deployment (it's not CI/CD yet). Changes have been pulled into 3.2 and 3.3 as well: https://github.com/ember-learn/guides-source/pull/110

Issue #79, opened 2018-06-05
invoking-actions-directly-on-component-collaborators bad example


mattmarcum:
https://guides.emberjs.com/release/components/triggering-changes-with-actions/#toc_invoking-actions-directly-on-component-collaborators

Hi hope this is the right place for this.  At the bottom of the section is a service `app/services/messaging.js` that has an `actions` hash with a method, `sendMessage` on it.  `sendMessage` should not be in an actions hash, it should just be directly accessible on the service itself.  I tried implementing this as is and it didn't work.  I moved my method out of the `actions` hash and it started working.

<img width="745" alt="screen shot 2018-06-05 at 5 50 44 pm" src="https://user-images.githubusercontent.com/4745044/41006903-02e18c4e-68e9-11e8-8f62-0efb3a839e1e.png">

acorncom:
Thanks much @mattmarcum, any chance you'd be able to send in a PR to address the issue?

mattmarcum:
https://ember-twiddle.com/3e57cb6d449c1e66ac758d721f3d0fe3

PR #89, opened 2018-06-15
Updated examples for invoking actions on component collaborators (#79)

mattmarcum:
Resolves https://github.com/ember-learn/guides-source/issues/79


jenweber:
Hiya @mattmarcum! Thanks for the Twiddle and PR. I think I found how that example was supposed to work ([modified Twiddle](https://ember-twiddle.com/3e62224faeca71fc464c68442dba5e8e?openFiles=templates.components.my-component.hbs%2Ctemplates.components.button-with-confirmation.hbs)). Could you take a look?

I think maybe the example in the guides worked as-is, however it's very unclear so maybe we need to add the non-block form service example too. It's odd that they are different.

mattmarcum:
Not sure I understand, in the modified twiddle you posted the someAction button doesn't work for me.  Is that on purpose?  As far as I can tell service don't support an actions hash, it doesn't have the action mixin - its just an ember object basically.  https://github.com/emberjs/ember.js/blob/v3.2.2/packages/ember-runtime/lib/system/service.js#L43

Are you wanting some specific changes to the pr?

jenweber:
Hey, thanks for your patience - it's been a while since I have been able to look into this. I would like to keep your suggestion, and I agree that having actions on a service is weird, but we need to modify the other code block too where it's using the service. Otherwise, the examples won't work at all.

The problem with the current guides content is that the `actions` hash on a service works _sometimes_. See [new twiddle demonstrating this](https://ember-twiddle.com/190c45aa9e6d4e696251bb7dd45849ce?openFiles=templates.components.parent-component.hbs%2Ctemplates.components.child-component.hbs)

So, we have a service with an action and a normal method:
``` javascript
// alert.js
import Service from '@ember/service';

export default Ember.Service.extend({
  actions: {
    thisIsAServiceAction() {
      alert('service action')
    }
  },
  thisIsANormalMethod() {
      alert('service method, not an action')
  }
});
```

And here's the `child-component`
```
<button {{action "onConfirm"}}></button>
```

## Things that work

We can pass the service's actions method to a block form child component and it will work:
```
<!-- parent-component.hbs -->
{{#child-component onConfirm=(action "thisIsAServiceAction" target=alert)}}
yielded content
{{/child-component}}
```

We can pass the service action to a non block component and it will work:
```
<!-- parent-component.hbs -->
{{child-component onConfirm=(action "thisIsAServiceAction" target=alert)}}
```

We can pass a _normal service method_ (non-action) as a variable:
```
{{child-component onConfirm=alert.thisIsANormalMethod}}
<br>
```

We can call a _normal service method_ (non-action) directly in the parent/home component:
```
<!-- parent-component.hbs -->
<button {{action "thisIsANormalMethod" target=alert}}>
   parent component button
</button>
```

## Things that don't work:

We cannot call a service action directly on the parent component. This does not work:
```
<button {{action "thisIsAServiceAction" target=alert}}>
    parent component button
</button>
```

We cannot pass a regular method to a block child component. This will not render:
```
<!-- parent-component.hbs -->
{{#child-component onConfirm=(action "thisIsANormalMethod" target=alert)}}
yielded content
{{/child-component}}
```

We cannot pass a regular method to a normal style component. This will not render:
```
<!-- parent-component.hbs -->
{{child-component onConfirm=(action "thisIsANormalMethod" target=alert)}}
```

mattmarcum:
Wow, great work!

It looks like there's a lot of inconsistencies in how ember components handle `actions` and `targets`.  My gut instinct is telling me that some of these are just unintended consequences of expecting that everything passed into the `action` helper as a `target` would have the [action-handler mixin](https://github.com/emberjs/ember.js/blob/v3.3.1/packages/ember-runtime/lib/mixins/action_handler.js) mixed in.  

Here's a [twiddle](https://ember-twiddle.com/432e7f72d6b8a43365acd4c6b02d7b5c?openFiles=templates.components.parent-component.hbs%2Ctemplates.components.child-component.hbs) showing that mixing in the `ActionHandler` to a service fixes one of the cases that was broke before.  Unfortunately it also breaks one of the examples that you listed as working...

I get the feeling that some of this behavior is just wrong and I'm wondering which should get documented as the `ember way`.  Since the code up on the guides just flat doesn't work maybe we should get the bare minimum update(whatever that is) merged so that guides at least have working/usable code?

jenweber:
@mattmarcum agreed on the minimal working example.

How about something [like this](https://ember-twiddle.com/7452c4afbbd5019baf9be8b9870f73dd?openFiles=templates.components.child-component.hbs%2Ctemplates.components.child-component.hbs)? I have a feeling that the target stuff pre-dates closure actions.

Btw it is ok to just PR for the latest version. We do not try to maintain past versions.

mattmarcum:
I think something like that would be better than what's up there now.  I'll close this pr.

PR #101, opened 2018-06-30
document injecting services into ES6 components

BryanCrotaz:
Injecting using `{inject as service}` doesn't work with ES6 classes in Ember 3.x.

Documenting usage of `ember-decorators` which do work.

[ember-decorators issue 249](https://github.com/ember-decorators/ember-decorators/issues/249#issue)

### DIFF
--- a/guides/v3.0.0/applications/dependency-injection.md
+++ b/guides/v3.0.0/applications/dependency-injection.md
@@ -197,6 +197,8 @@ export default Component.extend({
 
 #### Injection into ES6 classes
 
+**NOTE** ES6 classes in Ember are bleeding edge, so your mileage will vary elsewhere, but here's how to get injection working with them.
+
 The above `inject` method with ES6 classes works (accidentally) in Ember 2.x but not at all in Ember 3.x.
 
 > If you get the compile error `Assertion Failed: InjectedProperties should be defined with the inject computed property macros.` then you're using the above `inject` helper on at least one ES6 class in Ember 3.x. Replace it with the `@service` decorator instead as below.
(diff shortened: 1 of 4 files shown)


sivakumar-kailasam:
@BryanCrotaz Thanks for the PR but official support for es6 classes is still [a work in progress](https://github.com/emberjs/ember.js/issues/16134). 

Ember decorators is a great addon, but let's discuss first how it could be included in the guides via an issue.

//cc @ember-learn/learning-team

acorncom:
@BryanCrotaz sorry to hear about all the snags you hit, sounds quite frustrating ...

A little more context on our thinking here (which fairly closely matches what folks on that ember-decorators thread had to say). Our current approach with the guides is to document Ember best practices as recommended by the core teams). In the case of ES6 classes / decorators, we’re all eagerly looking forward to when they are the default option we can all get behind. But at the moment, there are some rough edges (as you sadly hit).

Our goal with the guides and other base learning resources is to recommend the smoothest/cleanest learning path for new folks and to leave experimentation to the folks who may be better setup to handle the frustrations and gashes inherent in living on the cutting (or bleeding?) edge.

If you’re passionate about about getting ES6 module syntax documented though, we’ve been talking about having all our code syntax in both the older Ember object format as well as in the bleeding edge ES6 object model format. Part of the reason we’ve been reworking the guides into an Ember app is to help with handling the complexities of documenting both the old and the new formats (which we may have to do for a time). If you’d love to help us tackle the technical headaches we’re going to hit there, we’d love to get a hand there. And it’d probably help us ship things a bit faster. 

Hope the context helps!

PR #105, opened 2018-07-04
Fix 'display rental details' test

ThiefZero:
In recap of the final test file, .`textContent` should be `.text()` to correctly run the test

### DIFF
--- a/guides/v3.2.0/tutorial/simple-component.md
+++ b/guides/v3.2.0/tutorial/simple-component.md
@@ -374,8 +374,8 @@ module('Integration | Component | rental listing', function (hooks) {
 
   test('should display rental details', async function(assert) {
     await render(hbs`{{rental-listing rental=rental}}`);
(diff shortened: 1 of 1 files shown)


acorncom:
Hey @ThiefZero, thanks for the feedback. Are you sure this is accurate though? We merged https://github.com/ember-learn/guides-source/pull/105 a few weeks ago and those tests all pass without needing the changes you're suggesting ...

ThiefZero:
Hello @acorncom thanks for responding.
Upon further investigation you are correct, my suggested changes would break the tests.

The confusion lies within the tutorial itself, currently the instructions for writing this test differs from the final recap of the entire test file.

At first the following is stated:

image


However in the recap the following is stated:
image

As the first one also causes an error building due to the fuction missing the `async` keyword, I believe it is best to change that part to the recap part. 

I'll close this pull request and create a new one reflecting above changes.

PR #106, opened 2018-07-04
Differentiate when a promise/record is returned

jpadilla:
This change adds `.then()` to `findRecord` and `findAll` examples and moves the comment on the end of the line to the previous line which should also help by removing the need to scroll towards the right to see that lil tidbit.

screenshot


jpadilla:
@acorncom should I do the same update for v3.3?

acorncom:
@jpadilla that'd be much appreciated ...

PR #107, opened 2018-07-06
Differentiate when a promise/record is returned

jpadilla:
Following up on #106 by applying same change for v3.3.0 docs


jenweber:
Thank you!!!

PR #108, opened 2018-07-08
Add missing 'async' before function

ThiefZero:
As thoroughly described in #105 this is the only change left 😄


jenweber:
Thank you @ThiefZero ! Merged.

I checked and this is [already correct in Super Rentals](https://github.com/ember-learn/super-rentals/blob/ca6e0567e0b87403d07ab3b5d274d74cda53e19b/tests/integration/components/rental-listing-test.js#L20)

PR #112, opened 2018-07-20
Fix examples bug on writing-helpers.md

hakilebara:
image


hakilebara:
Thanks for the tip 👍 I updated the PR to include 3.3, 3.1 & 3.0 docs as well.

PR #116, opened 2018-07-26
Update computed-properties.md

melaniespath:
Add missing word.


jenweber:
Thanks a bunch! Good catch. It reads better this way.

jenweber:
Merged.

PR #87, opened 2018-06-10
Replaces `this.get with ES5 getter where possible (#2303) in v3.2.0

cspanring:
Attempt to check off task 1 in https://github.com/emberjs/guides/issues/2303

@jenweber please have a look and/or pull in more eyes to make sure the changes are OK.

PR #122, opened 2018-07-29
apply cspanring's getter work to 3.3 #87

jenweber:
See #87


jenweber:
@cspanring will you have a look and let me know if this is good to go?

jenweber:
I need 1 reviewer from the learning team to push the green button. Then we are good to go!

PR #124, opened 2018-07-30
updates more `get` instances and language in object model

cspanring:
Attempt to check off task 2 in https://github.com/emberjs/guides/issues/2303 and update some of the language around `get` and `set`, including when to and when not to use it.

@jenweber please have a look. might conflict slightly with #122

### DIFF
--- a/guides/v3.3.0/object-model/classes-and-instances.md
+++ b/guides/v3.3.0/object-model/classes-and-instances.md
@@ -226,11 +226,27 @@ Person.create({
 
 ### Accessing Object Properties
 
-When accessing the properties of an object, use the [`get()`][7]
-and [`set()`][8] accessor methods:
+When reading a property value of an object, you can in most cases use the common Javascript dot notation, e.g. `myObject.myProperty`. 
 
-[7]: https://www.emberjs.com/api/ember/release/classes/@ember%2Fobject/methods/get?anchor=get
-[8]: https://www.emberjs.com/api/ember/release/classes/@ember%2Fobject/methods/set?anchor=set
+[Ember proxy objects][9] are the one big exception to this rule. If you're working with Ember proxy objects, including promise proxies for async relationships in Ember Data, you have to use Ember's [`get()`][7] accessor method to read values.
+
+Let's look at the following `blogPost` Ember Data model:
+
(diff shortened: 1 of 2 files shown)


guides/v3.3.0/object-model/classes-and-instances.md
```diff
@@ -226,11 +226,15 @@ Person.create({
 
 ### Accessing Object Properties
 
-When accessing the properties of an object, use the [`get()`][7]
-and [`set()`][8] accessor methods:
+When reading a property value of an object, you can in most cases use the common Javascript dot notation, e.g. `myObject.myProperty`. 
+
+The one big exception to this rule are [Ember proxy objects][9]: if you're working with Ember proxy objects, including promise proxies for async relationships in Ember Data, you have to use Ember's [`get()`][7] accessor method to read values.
```
jenweber: Slight edit:

[Ember proxy objects][9] are the one big exception to this rule. If you're working with Ember proxy objects, including promise proxies for async relationships in Ember Data, you have to use Ember's [`get()`][7] accessor method to read values.

jenweber: I also think maybe we need a small example here.

For example, say there is an Ember Data model, `blogPost`, which has many `comment` records associated with it. We could access attributes of the `blogPost` itself using dot notation, like `blogPost.title`. However, to show the comments that are part of the relationship, we would need to use `blogPost.get('comments')`.

guides/v3.3.0/object-model/classes-and-instances.md
```diff
@@ -226,11 +226,15 @@ Person.create({
...
+
+For updating property values use Ember's [`set()`][8] method. It will propagate the value change to computed properties, observers, templates, etc. If you "just" use Javascript's dot notation to update a property value, computed properties won't recalculate, observers won't fire and templates won't update.
```
jenweber: Likewise, I suggest that you flip the order of this sentence. "Always use Ember's [`set()`][8] method to update property values."

cspanring:
@jenweber thanks for your feedback! rebased, updated the language and added a simple Ember Data example to that section. please have a look again.

jenweber:
Thanks Christian!

PR #126, opened 2018-08-01
improves description on objects as computed property dependents keys

### LINKED ISSUE #82: Describe how computed props works with objects in more detail
When we get to https://guides.emberjs.com/release/object-model/computed-properties/#toc_multiple-dependents-on-the-same-object we don’t seem to specify that computed doesn’t work directly on an object but has to be pointed to props inside that object. To newcomers this can be trippy since they might wonder why they can't compute on the object itself.

cspanring:
* Attempt to improve the language around the details of using object references and properties as dependents keys on computed property.
* Updates the example code to use something slightly more interesting than `foo bar`.

@jenweber could use some stylistic proof reading.

resolves #82

### DIFF
--- a/guides/v3.3.0/object-model/computed-properties.md
+++ b/guides/v3.3.0/object-model/computed-properties.md
@@ -80,31 +80,68 @@ ironMan.fullName; // no console output since dependencies have not changed
 
 In the previous example, the `fullName` computed property depends on two other properties of the same object.  
 However, you may find that you have to observe the properties of a different object.
+
 For example, look at this computed property:
 
 ```javascript
 import EmberObject, { computed } from '@ember/object';
 
-let obj = EmberObject.extend({
-  baz: { foo: 'BLAMMO', bar: 'BLAZORZ' },
+let home = EmberObject.extend({
+  location: {
+    streetName: 'Evergreen Terrace',
+    streetNumber: 742
+  }
 
-  something: computed('baz.foo', 'baz.bar', function() {
-    return `${this.baz.foo} ${this.baz.bar}`;
+  address: computed('location.streetName', 'location.streetNumber', function() {
+    return `${this.location.streetNumber} ${this.location.streetName}`;
   })
 });
+
+home.address // 742 Evergreen Terrace
+home.set('location.streetNumber', 744)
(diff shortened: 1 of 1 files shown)


guides/v3.3.0/object-model/computed-properties.md
```diff
@@ -80,31 +80,68 @@ ironMan.fullName; // no console output since dependencies have not changed
...
 
 ```javascript
 import EmberObject, { computed } from '@ember/object';
 
-let obj = EmberObject.extend({
-  baz: { foo: 'BLAMMO', bar: 'BLAZORZ' },
+let home = EmberObject.extend({
+  location: {
+    street: 'Evergreen Terrace',
+    nr: 742
```
jenweber: I much prefer a real example like this over `foo bar`! Thank you! Maybe this object could be `streetName` and `streetNumber` for the keys. That might be clearer to beginners than an abbreviation.

guides/v3.3.0/object-model/computed-properties.md
```diff
@@ -80,31 +80,68 @@ ironMan.fullName; // no console output since dependencies have not changed
...
+    return `${this.location.nr} ${this.location.street}`;
+  })
+});
+
+home.address // 742 Evergreen Terrace
+home.set('location.nr', 744)
+home.address // 744 Evergreen Terrace
+```
+
+Important to remember here is that you have to observe the object properties. Adding the object reference `'location'` as dependents key will not recalculate the computed property `address` when the `street` or `nr` properties change. It will only recalculate when the object reference `location` is updated.
```
jenweber: "It is important to observe an object's properties, not the object itself that has properties nested inside. If the object reference `'location'` is used as a dependent key, the computed property will not recalculate when the `street` or `nr` properties change."

I would remove "It will only recalculate when the object reference `location` is updated." I think this would just confuse beginners.

jenweber: I might come back to this phrasing later. I feel like my revision is still too complex.

cspanring:
@jenweber thanks! made the requested changes.

jenweber:
Awesome, thank you!

PR #135, opened 2018-08-19
The `context` parameter is no more (+formatting)

pablobm:
The `context` parameter was removed from `find` and `findAll` in `ember-test-helpers`. Currently, trying to use it will get you an exception, as can be seen in this PR: https://github.com/emberjs/ember-test-helpers/pull/358 That's the first of these two commits.

Since I was at it, I added a second commit to try make the formatting less awkward. Currently it tries to be a definition list formatted as nested lists, but doesn't really achieve that. From the Showdown documentation[1], I see that the correct way to do that would be with 4 spaces, rather than the current 2. However I went to see what this would look like if instead I put both term+definition in a single list item:

alternative-formatting

Thoughts?

[1] Nested lists with Showdown: https://github.com/showdownjs/showdown/wiki/Showdown's-Markdown-syntax#nested-lists

### DIFF
--- a/guides/v3.3.0/testing/acceptance.md
+++ b/guides/v3.3.0/testing/acceptance.md
@@ -134,15 +134,14 @@ Synchronous helpers are performed immediately when triggered.
   - Returns the currently active route name.
 * [`currentURL()`][7]
   - Returns the current URL.
-* [`find(selector, context)`][8]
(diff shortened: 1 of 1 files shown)


jenweber:
Thanks! This is an improvement on all fronts. I prefer the simpler markdown too.

PR #143, opened 2018-09-04
Adding note for transitionning to a wildcard route

sly7-7:
I just found myself stuck because I forgot this empty argument (seems like the router interprets the path as a dynamic segment). The error seems intentionnally not thrown, and hard to track down: https://github.com/tildeio/router.js/blob/73d6a7f5ca6fba3cfff8460245fb5a3ceef8a2b5/lib/router/transition-intent/named-transition-intent.ts#L214)


jenweber:
Hello @sly7-7! Thanks for adding this to help other people who might get stuck otherwise. I made a couple teeny edits: replaced "ie" with "for example" for our international readers, added whitespace before the code block, and changed `route` to `routes` in the fake file codepath.  

Congrats on your first PR to the guides! If you spot anything else that could be improved, please let us know! Merged.

PR #149, opened 2018-09-14
Update docs to use prefixed `this.` in templates

chadhietala:
This likely needs to be read through end-to-end to ensure we are not missing anything. This is part of the migration plan for [RFC#308](https://github.com/emberjs/rfcs/blob/master/text/0308-deprecate-property-lookup-fallback.md)


jenweber:
I have questions about whether changes like these should merge now or be part of an "ember editions" push. @ember-learn/learning-team and @chadhietala, what are your thoughts?

jenweber:
Hi! This PR was still open as of creating the 3.5 release. This means before this PR to be merged, any changes requested should be made, and then the same additions should be applied to the upcoming 3.5 version files. They will be available later this week. Sorry about the extra step!

chadhietala:
I'm not sure as these changes are related to an RFC that pre-dates Editions. In the RFC I outlined the [transition path](https://github.com/emberjs/rfcs/blob/master/text/0308-deprecate-property-lookup-fallback.md#transition-path). The only parts that aren't done are updating the docs and moving forward on making this a default template lint sort of requires the docs to be updated. I would like to hear what @tomdale has to say since there have been talks about grouping behavior like this as part of Octane edition.

jenweber:
Unless someone has a specific objection to merging this, let's do it at the end of this week.

I get the sense that nobody actually thinks this needs to be deferred, we all just think that other people think it needs to be deferred.

jenweber:
To dos:
- [ ] Someone needs to review the content itself
- [ ] The changes need to be applied to the 3.5 files (or 3.6 if it gets released soon)
- [ ] Put the 3.4 files back to their prior state (i.e. make this PR target a later version)

PR #168, opened 2018-09-30
routes-and-templates: Fix anchor parameters in documentation links.

ybakos:
The URLs here have a trailing slash, considered part of the anchor, which results
in the link taking a user to the top of the doc page, rather than the specific
anchored method within the doc page.

Remove the trailing slash, so the anchor is accurate, and the user is taken
directly to the appropriate method in the doc page.


jenweber:
Awesome, thanks for this fix! And congrats on your first PR to the guides 🎉 

If you'd like to participate in [Hacktoberfest](https://hacktoberfest.digitalocean.com/), you just need to register and make 4 more PRs to any public open source project during October to get some swag.

Issue #162, opened 2018-09-29
Explain how to pass actions and then use actions that have been passed


jenweber:
There are a few different ways to pass actions in Ember. We don't realllly teach them in the Guides, though it's covered in the Tutorial.

To do:
- [ ] link to [this section](https://guides.emberjs.com/release/components/triggering-changes-with-actions/#toc_passing-the-action-to-the-component) on passing actions to components from [here](https://guides.emberjs.com/release/components/passing-properties-to-a-component/)
- [ ] include a note that actions can't be passed down from a route, only from a controller or another component

Changes should only be made to the markdown files for the latest version of Ember.

Parrryy:
I'd be happy to pick this up :)

jenweber:
Closed by https://github.com/ember-learn/guides-source/pull/177, thank you!

PR #177, opened 2018-10-01
Improve explanation of passing actions (#162)

Parrryy:
First draft for [#162](https://github.com/ember-learn/guides-source/issues/162)

### DIFF
--- a/guides/v3.4.0/components/passing-properties-to-a-component.md
+++ b/guides/v3.4.0/components/passing-properties-to-a-component.md
@@ -54,6 +54,21 @@ known as being "bound"). That is, if the value of `componentProperty`
 changes in the component, `outerProperty` will be updated to reflect that
 change. The reverse is true as well.
 
+In addition to making properties available, actions can be made available
+to components. This allows data to flow back up to its parent. You pass actions
+like this.
+
+```handlebars {data-filename=app/templates/index.hbs}
+  {{button-with-confirmation
+    text="Click here to unsubscribe."
+    onConfirm=(action 'unsubscribe')
+  }}
+```
+
+It is important to note that actions can only be passed from a controller or another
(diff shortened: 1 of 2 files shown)


guides/v3.4.0/components/passing-properties-to-a-component.md
```diff
@@ -54,6 +54,21 @@ known as being "bound"). That is, if the value of `componentProperty`
...
 change. The reverse is true as well.
 
+In addition to making properties available, actions can be made available
+to components. This allows data to flow back up to its parent. You pass actions
+like this.
+
+```handlebars {data-filename=app/templates/index.hbs}
+  {{button-with-confirmation
+    text="Click here to unsubscribe."
+    onConfirm=(action 'unsubscribe')
```
jessica-jordan: I think it would be good to update the quotes here in the `action` helper, too, to also use the double quoted version as the other string expression in the template examples:

```
(action "unsubscribe")
```

Parrryy:
I've updated the formatting of the quotes as instructed ☕️ 😄

guides/v3.4.0/components/passing-properties-to-a-component.md
```diff
@@ -54,6 +54,21 @@ known as being "bound"). That is, if the value of `componentProperty`
...
+
+```handlebars {data-filename=app/templates/index.hbs}
+  {{button-with-confirmation
+    text="Click here to unsubscribe."
+    onConfirm=(action "unsubscribe")
+  }}
+```
+
+It is important to note that actions can only be passed from a controller or another
+component. They cannot be passed from a route. See [passing an action to the component](/triggering-changes-with-actions/#toc_passing-the-action-to-the-component)
```
jessica-jordan: I just had another look locally and this already looks great! One more thing I noticed is, that the link here doesn't point to the correct page, yet:

linktoactions

Can you check what is missing for the link? I assume there are other relative links in the guides source which you can use as a reference, too.

PR #186, opened 2018-10-04
docs(Objects in Ember): update intro to objects

### LINKED ISSUE #155: Write a friendlier introduction to the Objects guide
When someone is done with the tutorial, the first thing they read is [this](https://guides.emberjs.com/release/object-model/). It's very confusing to beginners and right away, makes us look dated, because ES2015 is hardly new 😆 

To-do:
- [ ] take out the reference to ES2015 being "new"
- [ ] write a better introduction. Imagine someone is reading this on day-one of Ember! New learners are probably the only people who look at this page.
- [ ] only make changes to the markdown file for the latest version of the guides

A more friendly introduction could hit these points, but these are just suggestions, so feel free to take a different direction:
- One of the first things you'll notice when you generate JavaScript files in Ember is that most of the code you will write goes inside of an object.
- While an Ember Object might look a lot like an ES2015 JavaScript class, it has some special powers.
- The most important is that an Ember Object can be watched for changes
- (description of what is meant here)

chrisrng:
Fixes #155 

[rendered](https://github.com/ember-learn/guides-source/blob/e655e9abc7b3720e6964560c79eb8a5fcb29c01d/guides/v3.4.0/object-model/index.md)

### DIFF
--- a/guides/v3.4.0/object-model/index.md
+++ b/guides/v3.4.0/object-model/index.md
@@ -2,13 +2,13 @@ One of the first things you'll notice when you generate JavaScript files in Embe
 
 ### Introducing: Ember Objects
 
-The [Ember Object](https://www.emberjs.com/api/ember/release/classes/EmberObject) class extends plain JavaScript objects to provide core framework functions such as participating in Ember's [binding system](../bindings/) or how changes to an object automatically trigger updates to the user interface.
+The [Ember Object](https://www.emberjs.com/api/ember/release/classes/EmberObject) class extends plain JavaScript objects to provide core framework functions such as participating in Ember's [binding system](../object-model/bindings/) or how changes to an object automatically trigger updates to the user interface.
 
 Most objects in Ember, including Routes, Models, Services, Mixins, Controllers, and Components extend the `EmberObject` class.
 
 ### Why Ember Objects?
 
-The most important reason is that an Ember Object can be watched for changes – or _observed_. For example, being [observable](https://www.emberjs.com/api/ember/release/classes/Observable) is important for [computed properties](../computed-properties/). It is one of the fundamental ways that models, controllers and views communicate with each other in an Ember application.
+The most important reason is that an Ember Object can be watched for changes – or _observed_. For example, being [observable](https://www.emberjs.com/api/ember/release/classes/Observable) is important for [computed properties](../object-model/computed-properties/). It is one of the fundamental ways that models, controllers and views communicate with each other in an Ember application.
 
 ### More on Ember Objects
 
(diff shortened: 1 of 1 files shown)


guides/v3.4.0/object-model/index.md
```diff
@@ -1,18 +1,21 @@
...
 
-Ember also extends the JavaScript `Array` prototype with its
-[Ember.Enumerable](https://emberjs.com/api/ember/2.15/classes/Ember.Enumerable) interface to provide change observation for arrays.
+Most objects in Ember, including Routes, Models, Services, Mixins, Controllers, and Components extend the `EmberObject` class.
 
-Finally, Ember extends the `String` prototype with a few [formatting and
-localization methods](https://www.emberjs.com/api/ember/release/classes/String).
+### Why Ember Objects?
+
+The most important reason is that an Ember Object can be watched for changes – or _observed_. Ember Objects, being [observable](https://www.emberjs.com/api/ember/release/classes/Observable), supports observing any of its properties including [computed properties](https://guides.emberjs.com/release/object-model/computed-properties/). It is one of the fundamental ways that models, controllers and views communicate with each other in an Ember application.
```
jenweber: Possible simplification of "Ember Objects, being observable...."

For example, being [observable](https://www.emberjs.com/api/ember/release/classes/Observable) is important for [computed properties](https://guides.emberjs.com/release/object-model/computed-properties/).

guides/v3.4.0/object-model/index.md
```diff
@@ -1,18 +1,21 @@
...
-JavaScript objects don't support the observation of property value changes.
-Consequently, if an object is going to participate in Ember's binding
-system you may see an `Ember.Object` instead of a plain object.
+### Introducing: Ember Objects
 
-[Ember.Object](https://www.emberjs.com/api/ember/release/modules/@ember%2Fobject) also provides a class system, supporting features like mixins
-and constructor methods. Some features in Ember's object model are not present in
-JavaScript classes or common patterns, but all are aligned as much as possible
-with the language and proposed additions.
+The [Ember Object](https://www.emberjs.com/api/ember/release/classes/EmberObject) class extends plain JavaScript objects to provide core framework functions such as participating in Ember's binding system.
```
jenweber: Can we add a very short definition of "binding system" here? Something like "... binding system, where changes to an object automatically trigger updates to the user interface."

guides/v3.4.0/object-model/index.md
chrisrng: Fixed!

guides/v3.4.0/object-model/index.md
chrisrng: Added this as well

jenweber:
Hey, I missed one thing in my review. Links between topics within this app should use relative links, I.e. `../../topic/subtopic` rather than linking to URLs like guides.emberjs.com. Can you correct that? Thanks!

PR #189, opened 2018-10-06
autocomplete-component.md: Fix test module string description.

ybakos:
The tutorial mistakenly uses 'rental-listing' in the 'list-filter-test.js'
test module description string. Correct it, to match what the end user
would see after generating the test for this component.


sivakumar-kailasam:
Thanks @ybakos

PR #192, opened 2018-10-06
removes extra info for Ember.run wrapping

gokatz:


gokatz:
We can even club these two section by mentioning priority and purpose
* https://guides.emberjs.com/release/applications/run-loop/#toc_how-does-the-run-loop-work-in-ember
* https://guides.emberjs.com/release/applications/run-loop/#toc_what-happens-in-these-queues

jenweber:
Hey @sivakumar-kailasam can you take a look at this PR and let me know your thoughts on removing vs replacing this material?

jenweber:
Hi! This PR was still open as of creating the 3.5 release. This means before this PR to be merged, any changes requested should be made, and then the same additions should be applied to the upcoming 3.5 version files. They will be available later this week. Sorry about the extra step!

gokatz:
No issue. will accommodate these changes to the newer versions too.

gokatz:
But, when I tested in a 2.18 app, I don't get any run loop assertion during the test. Am I missing anything?

sivakumar-kailasam:
afaict, https://github.com/BackburnerJS/backburner.js/pull/306 got merged early in feb & released as 2.2.0. It made its way into ember in [3.2.0](https://github.com/emberjs/ember.js/blob/master/CHANGELOG.md#v320-may-31-2018). So unsure how that'd work on a prior version ¯\_(ツ)_/¯

gokatz:
will re-check and update

jenweber:
Hi @gokatz, I've chatted with a couple people like Mike North and have learned that the run guides are more useful than we thought. Could you take a second pass at this and see what sections we can salvage that explain the run loop, but maybe without the part where it makes it look like people should wrap code in a run?

gokatz:
Hi @jenweber, Sorry for the delay 😬 
Yes. I think we can only trim off the [section](https://guides.emberjs.com/release/applications/run-loop/#toc_how-is-run-loop-behaviour-different-when-testing) that talks about the issue in testing (autorun assertion). 

Also, we [mentioned](https://github.com/ember-learn/guides-source/commit/7e0b67e811a89c84f8abde4ad35dfa3c389a97c0#diff-f51ca89a10fe6d4d1423d42997fba6c4R242) autoruns are **suboptimal**. We can even rephrase those as per [this RFC](https://github.com/rwjblue/rfcs/commit/d5f8a31de3dbd4024100be063d420f8949469d66#diff-9e299cf1de89e9a906a5d0a1c1a4e901R159).

jenweber:
Also @gokatz I do think it would be good to refactor how we talk about runs. We should check to see what from that RFC has actually been implemented.

Issue #157, opened 2018-09-29
Explain how to transition between Routes


jenweber:
We should add a section to [Defining your routes](https://guides.emberjs.com/release/routing/defining-your-routes/) that very briefly explains how to move between routes, and link to the API docs

There are 3 main ways:
- `this.transitionTo` or `transitionToRoute` from a route
- using the Router Service
- link to in a template

ynotdraw:
Hey @jenweber, I'd love to take a stab at this!

jenweber:
Closed with merging #196 🎉

PR #196, opened 2018-10-08
Explain how to transition between Routes (#157)

ynotdraw:
RE: #157

I was struggling a bit finding a good spot for this section to fit in, so I ended up settling with the bottom.  I didn't want to tamper with the existing section orders, because it felt like the content flowed really well. At any rate, let me know if I'm missing anything here or if I should dig more into the concepts!


jenweber:
@ynotdraw thanks for this PR! I agree that's the best spot for the content.

PR #212, opened 2018-10-15
v3.6.0 with array helper

locks:


jenweber:
@locks @mansona we need a plan to make sure that the work done on 3.5 in the meantime doesn't get dropped. Can you look into exactly how we should do that? Maybe some git diff magic?

jenweber:
I tried changing the base to the v3.6-staging branch and GitHub didn't like it. I will merge/rebase and then open a new PR to see if that's easier to review.

jenweber:
Merged into the 3.6 staging branch via https://github.com/ember-learn/guides-source/pull/287

PR #213, opened 2018-10-15
Fixed a typo for the handle bars template

LalithaRajanala:
Fixed a typo for the handle bars template. It should be {{body}} instead of {{{body}}}


locks:
@LalithaRajanala thanks for the help :) If you have any more corrections, send them our way!

PR #215, opened 2018-10-17
Updated documentation for query-param scope

LalithaRajanala:
While working with {{link-to}} helper and query param today, I hit issue https://github.com/emberjs/ember.js/issues/16867.
I couldn't find a sample example to see why query param was getting set as undefined. Looks like that is the expected behavior based on the comment in the thread.
Hence updated the documentation to make it more clearer.

### DIFF
--- a/guides/v3.5.0/routing/query-params.md
+++ b/guides/v3.5.0/routing/query-params.md
@@ -255,6 +255,13 @@ This affects query param behavior in two ways:
    will become `/articles?page=2`.
 
 ### Sticky Query Param Values
+The query params are defined per route/controller. They are not global to the app. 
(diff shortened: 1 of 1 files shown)


guides/v3.5.0/routing/query-params.md
```diff
@@ -255,6 +255,13 @@ This affects query param behavior in two ways:
    will become `/articles?page=2`.
 
 ### Sticky Query Param Values
+The query params are defined per route/controller. They are not global to the app. 
+For example, if a route `first-route` has a query param `firstParam` associated with it and we try to navigate to `first-route` by using `link-to` helper from a different route `second-route`, like in the following handlebar template, the `firstParam` will be omitted.
```
Gaurav0: Does it matters whether 'first-route' is an ancestor of 'second-route' or not?

jenweber:
Thanks @LalithaRajanala for the PR and @Gaurav0 for reviewing!

PR #216, opened 2018-10-17
Slack -> Discord link in readme

hakubo:


mansona:
@MelSumner can we use this link like this? or do we need to add the invite link to this too?

hakubo:
@MelSumner ?

jenweber:
Thanks for this PR! I am not sure how stable that URL is so I added the invite link too, just in case.

jenweber:
Merged

PR #220, opened 2018-10-18
Clarify loading search within a route (#35)

### LINKED ISSUE #35: Nested routes can't find top level loading substates
Seems that the documentation is missed an important part of the bubble mechanism in loading substate.

It says that a loading substate is bubble up to root level loading.hbs, but really only render the route-level loading template.

Here is the original issue:
emberjs/ember.js#12367

theroncross:
closes #35 

This is as much a sanity check of my understanding as it is an attempt to help the community :grin: 

Please let me know if I've missed anything.

### DIFF
--- a/guides/v3.5.0/routing/loading-and-error-substates.md
+++ b/guides/v3.5.0/routing/loading-and-error-substates.md
@@ -63,6 +63,8 @@ Router.map(function() {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


jenweber:
Thanks @theroncross! I'm asking another person to look this over for accuracy, since I don't use loading templates much.

PR #221, opened 2018-10-18
fixed referencing old Ember.Enumerable

skyboyer:
It seems for me that Ember.Enumerable was used in older releases of Ember. Currently link to https://emberjs.com/api/ember/release/classes/Ember.Enumerable just returns 404.


skyboyer:
in case this change is valid: should I also check retrospectively other versions' guides? or must it be separated PR?

jenweber:
Hi @skyboyer! Thanks for this PR and catching it.

Our rule of thumb is that broken links and harmful, blatantly incorrect info are worth fixing in earlier versions. If you're willing to do it, it would be great to have this change in earlier versions too. I will merge this now to get the improvement in for users ASAP, but a second PR would be awesome!

PR #224, opened 2018-10-25
Object.freeze prototype properties

locks:


jenweber:
I agree that we should modify the lint rule.

rwjblue:
AFAIK they already are special cases?

krisselden:
frozen arrays have a non trivial performance cost in v8, it would be nicer if this could leverage the debug stripping of the babel-plugin-debug-macros

Issue #166, opened 2018-09-30
Show how to do arrays/objects in an Ember Data model


jenweber:
[This page](https://guides.emberjs.com/release/models/defining-models/) shows how to do type transforms. But what about if a record has a deeply nested read-only object or array? New learners might think they need to make relationships and child models, and this gets difficult quickly, however there is a simple solution.

Add a note and an example of read-only arrays and nested objects in Ember Data:

```
location: DS.attr() // a read-only object
tags: DS.attr() // a read-only array
```

```
{{model.location.latitude}}
```

joshmcrty:
This is mentioned on the Relationships page already: https://guides.emberjs.com/release/models/relationships/

Should it be moved to the Defining Models page instead?

spruce:
This should be closed as of #225

PR #225, opened 2018-10-26
Added read-only attributes to defining models (#166)

Parrryy:
Added to defining models page.


jenweber:
This is great, thank you! If I had a dollar for every time someone had no idea you could do this and tried to make some really intense relationship data...

Issue #191, opened 2018-10-06
Flag All Advanced Topics in the Guides Source


jessica-jordan:
Part of https://github.com/ember-learn/guides-app/issues/259

We need help to flag all issues that can be considered as advanced in the Guides Source! It would be useful to add a flag `isAdvanced: true` to all the topics which are most useful to experienced users.

For more context, see also the [related Quest Issue](https://github.com/ember-learn/guides-app/issues/259).

spruce:
I'd like to take a stab at this.

PR #234, opened 2018-10-31
Adding advanced flags to pages.yml

spruce:
I took the liberty to create a proposal of topics which should be advanced.

My method is mostly relying on my experience of using ember and which topics I used in a rather simple app. This is only the first stab. I have no hard opinion which should be advanced and which not. I hope I did understand correct where it should be added.

Issue link: https://github.com/ember-learn/guides-source/issues/191

### DIFF
--- a/guides/v3.5.0/pages.yml
+++ b/guides/v3.5.0/pages.yml
@@ -54,14 +54,17 @@
       url: "classes-and-instances"
     - title: "Reopening Classes and Instances"
       url: "reopening-classes-and-instances"
+      isAdvanced: true
     - title: "Computed Properties"
       url: "computed-properties"
     - title: "Computed Properties and Aggregate Data"
       url: "computed-properties-and-aggregate-data"
     - title: "Observers"
       url: "observers"
+      isAdvanced: true
     - title: "Bindings"
       url: "bindings"
+      isAdvanced: true
     - title: "Enumerables"
       url: 'enumerables'
 
@@ -80,12 +83,15 @@
       url: "redirection"
     - title: "Preventing and Retrying Transitions"
       url: "preventing-and-retrying-transitions"
+      isAdvanced: true
     - title: "Loading / Error Substates"
       url: "loading-and-error-substates"
+      isAdvanced: true
     - title: "Query Parameters"
       url: "query-params"
     - title: "Asynchronous Routing"
       url: "asynchronous-routing"
+      isAdvanced: true
 
 - title: "Templates"
   url: 'templates'
@@ -110,8 +116,10 @@
       url: "input-helpers"
     - title: "Development Helpers"
       url: "development-helpers"
+      isAdvanced: true
     - title: "Writing Helpers"
       url: "writing-helpers"
+      isAdvanced: true
 
 - title: "Components"
   url: 'components'
@@ -126,6 +134,7 @@
       url: "wrapping-content-in-a-component"
     - title: "Customizing a Component's Element"
       url: "customizing-a-components-element"
+      isAdvanced: true
     - title: "Using Block Params"
       url: "block-params"
     - title: "Handling Events"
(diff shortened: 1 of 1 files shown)


mansona:
Hey @spruce 👋 

Thanks a million for taking the time to look into this 🎉 It's a great place to start the conversation, even if it is only a "first stab" 😉 

Just to clarify on your thoughts for these sections. By marking the *section* as advanced do you think that 100% of the content on those pages should be considered advanced? Or did you mark those pages thinking that they contained at least some advanced topics?

One of the things that came out of the conversation about this was that some pages would have some sub-headings that might be considered "advanced" and we might need to not only mark sections/pages but also subheadings within a page. Do you have any thoughts on this since you've done a pass over the documentation with this in mind?

spruce:
Hi @mansona 

I tried to mark the pages where I think that the overwhelming majority of topics are advanced. Some pages that I did not mark have some advanced content in them but not enough for me to cross my "threshold". I couldn't think of a simple way to mark only some parts of a topic as advanced. That's why I only marked whole pages. If I'm wrong, please enlighten me. 😄 

Lets go over one for one. If I got anything wrong in your opinion, please correct me.
- `reopening-classes-and-instances`: In my point of view can be abused and not a necessity for basic ember usage
- `computed-properties-and-aggregate-data`: I was on the fence about that one. CPs on Arrays are quite basic but the cp macros are not always needed.
- `observers`: Is stated on the page
- `bindings`: Normally not used
- `preventing-and-retrying-transitions`: This one is also on the line. For users to get started, in my opinion, it is not needed for them to retry transitions. When using something like `ember-simple-auth` you basically get this without completely understanding how the transition hooks work.
- `loading-and-error-substates`: I know they exists but have only used them twice. But one could argue that this should be basic because if I don't know it I can't use it and I will find another solution to this problem. 
- `asynchronous-routing`: stated by the page.
- `development-helpers`: I can imagine that this is a little strange to new users.
- `writing-helpers`: I think that in the beginning users use helpers provided by others and only then start to write their own.
- `customizing-a-components-element`: With the outlook onto possibly outer-html components soon I think this will get obsolete that's way I thought it could be advanced
- `pushing-records-into-the-store`: For the happy path of ember with ember-data one doesn't need these methods, right? That means for me that they are advanced.
- `handling-metadata`: For the first basic application metadata is not needed. It may be strange for a new learner why there is metadata.
- `customizing-adapters`: In my opinion not on the happy simple path.
- `customizing-serializers`: Same as adapters.
- `applications-and-instances`: Is not really needed to use ember.
- `dependency-injection`: Only the ad-hoc injection header is basic the rest is nice-to-know but not necessary to work with ember. Maybe later in tests to understand what is looked up.
- `initializers`: The page mostly talks about how to make DI work. As I think this is advanced so is this.
- `run-loop`: Says so on the page in a note.
- `feature-flags`: New Learners will not try out new features from the get-go.
- `optional-features`: In my opinion this only allows to go of the paved way to use ember and should only be used by advanced users.
- `contributing`: As sad as it is I think new ember users will not immediately contribute to ember or need to understand how the project is structured.

mansona:
Wow! thanks for the **extremely** detailed breakdown 😍 This is really quite amazing!

I only have a few things that I don't fully agree with but overall you're completely on the right track 🎉 (I think anyway)

As for marking a particular section as being "advanced" I have had a bit of a "pet project" that I'm trying to make a reality for the Guides. We *could* mark sections as advanced using something like this: 

screenshot 2018-11-01 at 09 30 27

Obviously the copy needs a bit of work but we could use this idea to marks segments if the rest of the Learning team liked it 👍

spruce:
I also would like to collapse the section and provide, in the note by Tomster, a link to expand the section. So one is not inclined to ignore the message and keep reading.

Getting back to the technical side of things. How could this be achieved? 
Would you "simply" add this note and not have some kind of configuration for the section that marks it as advanced? How would the isAdvanced flags from the pages.yml be used? This is going into the territory of https://github.com/ember-learn/guides-app/issues/260 , right?

I think the copy of the note is alright 👍 . I must say I like the friendlieness of Tomster in general and the note in particular.

mansona:
So right now we're not on the stage of making any major changes to the functionality, like collapsing sections or filtering things out. We decided in the meeting that we would **markup** things that are advanced and then once we have done that we could look at what we decided on and decide what needed to be done 👍 

For the technical side of things, to achieve this I literally just added the HTML to the Markdown directly: 

```
### Computed Property Macros

<div class="cta">
  <div class="cta-note advanced">
    <div class="cta-note-body">
      <div class="cta-note-heading">Tomster says...</div>
        This section is considered advanced.
      </div>
    </div>
    <img src="/images/mascots/tomster.png" role="presentation" alt="Ember Mascot">
  </div>
</div>
```

It's not glamorous but it does the job 👍 and we can consider the presence of this as enough meta-data to say that the section is to be considered advanced for now (if we're happy with this sort of thing in the guides that is)

btw. do you know about the weekly meeting @spruce? It's 12:30 EST every Thursday, and we usually post the link in the #dev-ember-learning channel in discord

acorncom:
This looks like a good breakdown to me overall. Tom Dale brought up a good point in the Learn Team meeting today about there not necessarily being a distinction in a learners mind between "simple" and "advanced" when working on learning something (he was specifically addressing this in regards to Ember Data serializers and adapters). We'd have to be careful if we segment those off, but I think we could add "tie-in" links toward the bottom of a page (something along the lines of "if your backend doesn't automatically behave using JSON-API standards, then head over to the "Customizing Serializers" page for details on how to handle that")

jenweber:
Given the approved review and since this doesn't change behavior, I'm merging it. Thank you @spruce for your detailed analysis!

Overall I think this will be helpful in our quest to help construct a simpler learning path for Ember.

Issue #20, opened 2018-04-22
Epic - improving Ember Data sections


jenweber:
This issue is a reference for future Ember Data section improvements. Following discussion and consensus, these items should be broken into separate issues and small PRs.

- [ ] sample req and res objects/spec for different adapters/serializers
- [ ] Replace "customizing" language for serializers. I think we could simplify the mental model for everyone by teaching more about what a serializer is and how to write one instead of focusing on “customizing”
- [ ] Focus serializer explanations around most useful API methods. There are only 3 necessary methods, only 2 of which are “happy path”. Those methods are `normalizeResponse` `serialize` and `pushPayload`, and each can be described in an ideal world as “a pure function that converts from non-json-api to json-api or vice-versa”. The 2 happy path methods are `normalizeResponse` and `serialize`
- [ ] make sure distinction between `model` and `record` is clear. `record` for instances and `ModelClass` for the class
- [ ] better intro explaining what adapters are
- [ ] overall, gradually align on json-api terminology within ember-data, e.g. instead of saying `record` say `resource` and `resource-identifier`

Helpful terminology definitions:
- `store`: a cache for `records`
- `Model`: a schema class used when instantiating a `record`
- `Adapter`: a request-manager: it takes instructions on what to find and manages the process of making an API request and returning the response. Essentially an abstraction over fetch.
- `Serializer`: a formatter used to convert your API responses into json-api format if they are not already (caveat, needs to be a subset of json-api (camelCase member names and dasherized, singularized types)
- `finders`: a utility method for chaining together `request => format data => update cache`
- `records` (e.g. instances of Models) - think of them as “remote state”. A `record` is some data that lives outside of the lifecycle of the application, e.g. on your server

jenweber:
cc @runspired - please edit however you want ^^^

lifeart:
In serializers: `_normalizeDocumentHelper`, `extractAttributes`, `extractRelationships` - most used methods for me, it will be nice to cover it for newbies

ppcano:
Related [discuss thread](https://discuss.emberjs.com/t/when-to-add-ember-data/15050)

zachgarwood:
[...]

PR #240, opened 2018-11-02
Add `Ember Data flexibility` section

ppcano:
The change splits the section `What are Ember Data models?`  in two parts:

- `What are Ember Data models?`
- `Ember Data flexibility`

I described the reasons for this change on the [`When to add Ember Data` discussion](https://discuss.emberjs.com/t/when-to-add-ember-data/15050/14).

> Some folks decided to provide a JSON-API backend because Ember Data uses it as a default format, and this added more work for them. I think extending the REST adapter and use a conventional REST backend would have been a better choice for them. 

This made me think how the documentation of the default format of Ember Data  is communicated or may be interpreted. 

@jenweber suggested creating small incremental PRs, so this is a small change to highlight more the existing content around the "Flexibility" capabilities that was included in the `What are Ember Data models?`.

Additionally, I also thought about:

1) extend a little more the `Ember Data flexibility` section.

2) Tweak the content of the `Convention Over Configuration with JSON API` section.

But I didn't want to include more changes without validating if there is a shared opinion on the topic. 

Relates to #20

### DIFF
--- a/guides/v3.5.0/models/index.md
+++ b/guides/v3.5.0/models/index.md
@@ -54,9 +54,25 @@ more information about how components get model data, see the
 [Specifying a Route's Model](../routing/specifying-a-routes-model/)
 guide.
 
+At first, using Ember Data may feel different than the way you're used
+to writing JavaScript applications. Many developers are familiar with
+using AJAX to fetch raw JSON data from an endpoint, which may appear
+easy at first. Over time, however, complexity leaks out into your
+application code, making it hard to maintain.
+
+With Ember Data, managing models as your application grows becomes both
+simple _and_ easy.
+
+Once you have an understanding of Ember Data, you will have a much
+better way to manage the complexity of data loading in your application.
+This will allow your code to evolve without becoming a mess.
+
+## Ember Data flexibility
+
 Thanks to its use of the _adapter pattern_, Ember Data can be configured
-to work with many different kinds of backends. There is [an entire
-ecosystem of adapters][adapters] that allow your Ember app to talk to different
+to work with many different kinds of backends. There is [an entire ecosystem of adapters][adapters]
+and several [built-in adapters](./customizing-adapters/)
+that allow your Ember app to talk to different
 types of servers without you writing any networking code.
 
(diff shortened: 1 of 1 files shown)


guides/v3.5.0/models/index.md
```diff
@@ -54,9 +54,25 @@ more information about how components get model data, see the
...
+using AJAX to fetch raw JSON data from an endpoint, which may appear
+easy at first. Over time, however, complexity leaks out into your
+application code, making it hard to maintain.
+
+With Ember Data, managing models as your application grows becomes both
+simple _and_ easy.
+
+Once you have an understanding of Ember Data, you will have a much
+better way to manage the complexity of data loading in your application.
+This will allow your code to evolve without becoming a mess.
```
jenweber:
```suggestion
This will allow your code to evolve and grow, with better maintainability.
```

Agreed, if we can not use the "mess" language, it's better. Some of my past writing drew criticism for a similar statement and I think people had good points.

guides/v3.5.0/models/index.md
```diff
@@ -54,9 +54,25 @@ more information about how components get model data, see the
...
 guide.
 
+At first, using Ember Data may feel different than the way you're used
+to writing JavaScript applications. Many developers are familiar with
+using AJAX to fetch raw JSON data from an endpoint, which may appear
+easy at first. Over time, however, complexity leaks out into your
+application code, making it hard to maintain.
+
+With Ember Data, managing models as your application grows becomes both
+simple _and_ easy.
```
jenweber:
```suggestion
simpler _and_ easier.
```
It can still be hard to use Ember Data, sometimes. I think maybe the -er is a bit more realistic and kinder to new learners.

jenweber:
@ppcano I like both of these ideas. Want to give a rough draft a try and PR it? We can also run some questions by the Ember Data team to get their vision on these points.

Also, we have been kinda moving away from the "Convention over configuration" type language because it doesn't really seem to resonate anymore, so maybe you could help improve that too.

The JSONAPI section was probably written when the project was brand-new. I have a feeling that we can keep the same content level with fewer words, now that the project is better established and known.

Issue #230, opened 2018-10-28
Move away from google maps in super rentals


toddjordan:
google maps is more complicated to use (for free) and we've gotten feedback of its issues from several learners.  It also doesn't work unpaid once you deploy it to a server. We'd like to move to a more friendly to OSS library, such as leaflet js.

There's an issue representing this in the super rentals example repo.  Opening here to track against the tutorial-proper.
https://github.com/ember-learn/super-rentals/issues/90

toddjordan:
Here are some relevant PRs and Repos that are a part of this fix:
https://github.com/ember-learn/super-rentals/pull/110/files
https://github.com/ember-learn/ember-simple-leaflet-maps

toddjordan:
Merged #242 which implements a google maps replacement

Issue #228, opened 2018-10-28
Create an async/await version of the complex component sections of Super Rentals


toddjordan:
Currently the [complex component](https://guides.emberjs.com/release/tutorial/autocomplete-component/) section of Super Rentals features an asynchronous search.  

There is some code there to ensure search requests and responses do not get out of sync, and it is not ideal for new learners.

What I'd like to do is take advantage of async await in this code, which would simplify it, as well as show learners that its a valid option within ember.

locks:
No longer applies due to re-written Super Rentals!

PR #242, opened 2018-11-05
Replace google maps with leaflet

toddjordan:
Addresses #230


guides/v3.5.0/tutorial/service.md
```diff
@@ -1,99 +1,52 @@
...
+For the purpose of this tutorial we will take advantage of the [Leaflet Maps API](https://leafletjs.com/)
+The [Leaflet Maps API](https://leafletjs.com/) requires us to reference its library from a remote script.
+In this case we'll provide this script to our Ember app via an addon called `ember-simple-leaflet-maps`.
 
 ```bash
-GOOGLE_MAPS_API_KEY=<your key here> ember s
+ember install ember-simple-leaflet-maps
 ```
 
-### Accessing the Google Maps API with a Utility
```
toddjordan: Removing the Utility part of this guides section.  I thought it wasn't necessary because:

* It was too many things introduced at the same time
* It was dealing directly with the maps API
* Removing it reduces the size/complexity of the section.

guides/v3.5.0/tutorial/service.md
```diff
@@ -108,56 +61,59 @@ Let's get started creating our service by generating it through Ember CLI,
 which will create the service file, as well as a unit test for it.
 
 ```bash
-ember g service maps
+ember g service map-element
```
toddjordan: Changed the service name to `map-element` because

* The new service I created that interacts with leaflet is called `map`
* This service actually creates an HTML element with a rendered map

guides/v3.5.0/tutorial/service.md
```diff
@@ -108,56 +61,59 @@ Let's get started creating our service by generating it through Ember CLI,
...
 ```
 
 Now implement the service as follows.
+The main API will be an [async function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) called `getMapElement`,
```
toddjordan: Introducing async/await here.  We may want to explain it some more.  Also linking to an explanation.  I've actually created in issue for introducing it earlier, #228

guides/v3.5.0/tutorial/service.md
```diff
@@ -108,56 +61,59 @@ Let's get started creating our service by generating it through Ember CLI,
...
+      set(this, 'cachedMaps', {});
     }
+    this._super(...arguments);
   },
 
-  getMapElement(location) {
+  async getMapElement(location) {
     let camelizedLocation = camelize(location);
-    let element = this.get(`cachedMaps.${camelizedLocation}`);
+    let element = get(this, `cachedMaps.${camelizedLocation}`);
```
toddjordan: I think I can get away with accessing it through `this.chachedMaps[camelizedLocation]` instead.  What do you think?

guides/v3.5.0/tutorial/service.md
```diff
@@ -167,38 +123,30 @@ ember g component location-map
 
 Running this command generates three files: a component JavaScript file, a template, and a test file.
 
-Let's start by adding a `div` element to the component template.
-This `div` will act as a place for the 3rd party map API to render the map to.
-
-```handlebars {data-filename="app/templates/components/location-map.hbs" data-diff="-1,+2"}
-{{yield}}
-<div class="map-container"></div>
```
toddjordan: since this div was only to add a class name and append the map, I figured I could just do this to the component top level element instead.

guides/v3.5.0/tutorial/service.md
```diff
@@ -167,38 +123,30 @@ ember g component location-map
...
+  classNames: ['map-container'],
+  mapElement: service(),
 
   didInsertElement() {
     this._super(...arguments);
-    let mapElement = this.maps.getMapElement(this.location);
-    let mapContainerElement = this.element.querySelector('.map-container');
-    mapContainerElement.appendChild(mapElement);
+    this.mapElement.getMapElement(this.location).then((mapElement) => {
+      this.$().append(mapElement);
```
toddjordan: @sivakumar-kailasam made a good point that we should refer to this as `this.element` instead.

guides/v3.5.0/tutorial/service.md
```diff
@@ -208,25 +156,30 @@ This property will be passed in to the component by its parent template below.
...
-  <div class="detail type">
-    <span>Type:</span> {{rental-property-type rental.category}}
-      - {{rental.category}}
-  </div>
-  <div class="detail location">
-    <span>Location:</span> {{rental.city}}
-  </div>
-  <div class="detail bedrooms">
-    <span>Number of bedrooms:</span> {{rental.bedrooms}}
+  <div class="details">
```
toddjordan: I changed the DOM of this component to make formatting this with flexbox easier.  I had to make some tutorial css addon changes because the leaflet map required different styling. I still need to ripple this back through the component parts for the tutorial.

guides/v3.5.0/tutorial/service.md
jenweber: This is a good move! Totally agree

jenweber:
Can be merged when https://github.com/ember-learn/super-rentals/pull/110 is done!

guides/v3.5.0/tutorial/service.md
```diff
@@ -208,25 +168,30 @@ This property will be passed in to the component by its parent template below.
...
+    <h3>{{link-to rental.title "rentals.show" rental class=rental.id}}</h3>
```
jenweber:
```suggestion
    <h3>{{rental.title}}</h3>
```

The `rentals.show` route doesn't exist at this point in the tutorial, so the `{{link-to}}` breaks the app at this stage of the walkthrough

jenweber:
If I copy-paste things as they are, I get a weird layout problem that doesn't show up in the super-rentals repo. It looks like an override of `position: relative` is needed in the stylesheets from the addon:

screen shot 2018-11-10 at 5 18 15 pm

toddjordan:
@jenweber the CSS issues you see is because I haven't published the new version of ember-cli-tutorial style.

You can check out the PR on the super-rentals repo for a view of what the layout looks like with the new CSS.  When we merge this pr, I will publish the update to the tutorial css.

I also made additional fixes this weekend so things should be good to go.

toddjordan:
Currently my view (from a new project just using tutorial examples (with ember-cli-tutorial-sytle@3.x)
image

toddjordan:
Also,
image

toddjordan:
Also, 
image

guides/v3.5.0/tutorial/service.md
toddjordan: Fixt

PR #279, opened 2018-11-18
Add missing namespace.

enderahmetyurt:


jenweber:
Hi @enderahmetyurt! Thanks for dropping by and helping out.

The `namespace: 'api'` attribute goes in an adapter, not a model, so I am closing this PR. However, as I read that section over, it seems that maybe it's confusing because we talk about the Model and namespace but without showing the adapter that was made. Would you like to help add that adapter code block? If so, please open an issue and a PR.
 
This is the section that could have the adapter example go after it:
> When we call findAll, Ember Data will attempt to fetch rentals from /api/rentals. If you recall, in the section titled Installing Addons we set up an adapter to route data requests through /api.

The code block is at the bottom of this page: https://guides.emberjs.com/release/tutorial/installing-addons/

Issue #245, opened 2018-11-10
List of tiny tutorial fixes that should happen


jenweber:
To help with one of these issues, please open a separate issue for it and put the changes in its own PR. This is a list of things I noticed while reading the tutorial from start to finish:

## Volunteer needed
- [ ] "Creating your app" should have a quick note about the kind of knowledge prerequisites that are helpful (not required)
- [x] backslash typo, `for short/` is in `routes-and-templates.md`
- [ ] "functions that are invoked `automatically when a route is rendered`"
- [ ] "To keep things simple at first, we'll populate the model for our rental listing page to use a hard-coded array of JavaScript objects." is wordy
- [ ] "client HTTP stubbing library" will be confusing to new devs. Could be improved by saying immediately that "What this means is Mirage will allow us to create fake data..."
- [ ] "In order for this to work, we need our application to default to making requests to the namespace of /api" could be stated more plainly
- [ ] "Let's call this action toggleImageSize" block breaks the app, since the action is not defined. Needs a note that for the next step to work, we need to make changes in both the js and hbs file
- [ ] needs an explanation of when to make a helper instead of a component
- [ ] needs a quick explanation of the difference between guides and API docs
- [ ] "using ember data" should briefly explain how the ember data model and the model hook are different, or maybe link to that explanation in the guides
- [ ] "An Ember Service is an object that can be made available for different parts of an application." does not differentiate from components. Should indicate that it's special because there is only one instance of it, and it can be used almost anywhere to store shared information or methods.
- [ ] "3rd party map API" should prob be "third-party" and we should explain what is meant by that term

## Done
- [x] linting -  Unnecessary string concatenation. Use {{rental.image}} instead of "{{rental.image}}"
- [x] "Building a Simple Component" page needs a definition of what a component is, and its relationship to routes #simple
- [x] Needs plainer words - "Here we invoke the rental-listing component by name, and assign each rentalUnit as the rental attribute of the component."
- [x] "The value of isWide comes from our component's JavaScript file" should be followed with a statement like "any attributes in the component's JavaScript file can be used in its template"
- [x] "Be sure to restart your server after the addon has installed successfully." should tell them how to stop the server, control-c #282 
- [x] Handlebars are not explained, in general https://github.com/ember-learn/guides-source/pull/291#pullrequestreview-180530621
- [x] "All we want to do when a user..." section provides too much info. Should just introduce the hook and replaceWith. 
[...]

PR #283, opened 2018-11-19
Defines what a component is and is used for

zachgarwood:
On the Building a Simple Component page, we jump right into building a
component without defining what a component is or what they are use for.
This change adds a brief definition (the same one used in
https://guides.emberjs.com/v3.5.0/components/defining-a-component/) and
a reason why the user might want to use them.

This addresses one of the "tiny tutorial fixes" @jenweber listed
in #245.


BradenLawrence:
This change looks good to me!

jenweber:
Awesome, thank you!

PR #293, opened 2018-11-28
Updates the explanation of template helpers

zachgarwood:
On the Creating a Handlebars Helper page, we understate the purpose and
usage of template helpers. This change modifies the opening paragraphs to
go into more details.

This addresses one of the "tiny tutorial fixes" @jenweber listed
in #245.


zachgarwood:
I'm worried I might have used too much jargon with terms like "encapsulate," "declarative," and "presentation logic."

PR #302, opened 2018-12-06
Remove no-unnecessary-concat jslint violation

maxcal:
This removes the following JSLint violation `Unnecessary string concatenation. Use {{rental.image}} instead of "{{rental.image}}".  no-unnecessary-concat`.


jenweber:
Awesome, thank you for this fix!

PR #307, opened 2018-12-10
 ReferenceError: click is not defined

rimian:
Add missing `click` import


jenweber:
@locks this one needs to be added to 3.6 as well

Issue #310, opened 2018-12-12
It's not obvious how to use query-params from the template/links section


sonic1981:
The [template/links](https://guides.emberjs.com/v3.4.0/templates/links/)/query-params section is a bit cryptic and doesn't show you how to set up query params. it needs a link to [query params section of route](https://guides.emberjs.com/v3.5.0/routing/query-params/)

I'd suggest adding something along the lines of:

`For more information on how to use query parameters see the [query parameters](../../routing/query-params/) section in Routing.`

I'm happy to do this if people agree?

jenweber:
Sorry to not respond sooner. We only manage to review Issues/PRs every few days but this is a great idea and we appreciate the help! 

Closed by #311, thanks!

sonic1981:
@jenweber  No worries! This was plenty fast enough! :)

PR #311, opened 2018-12-12
Add link to the query parameters section in routing from links

sonic1981:
https://github.com/ember-learn/guides-source/issues/310


jenweber:
Awesome, thank you @sonic1981 !

For future PRs, you can submit just changes to the latest version that is live and the upcoming version (if those files exist). Generally, we don't try to update older versions unless there is actively incorrect info or a broken link.

Merged 🎉

PR #316, opened 2018-12-12
Remove Contributing to Ember.js

locks:
It is constantly getting out of date, and it feels insufficient and confusing more than helpful.


jenweber:
I don't think we can just remove these. We'll kill links. I was under the impression that URLs in the guides are semVer, and we either need to redirect or keep the page and link to the best content.

Issue #320, opened 2018-12-12
Update "Managing Dependencies"


locks:
## Goal
Update the Addons and Dependencies/Managing Dependencies sub-section of the Ember Guides. (https://guides.emberjs.com/release/addons-and-dependencies/managing-dependencies/) 

## To Do
- [x] Expand the Addons section by adding 2-3 paragraphs that explain more about the addon ecosystem, and maybe expound on what ember observer is, and what you can expect to find there. 
- [x] Update the explanation for `ember-auto-import` to the "regular npm packages" section

Turbo87:
@locks with all this keep in mind that `ember-auto-import` does not compile dependencies with Babel, so if they are incompatible with your browser targets you're out of luck (see https://github.com/ef4/ember-auto-import/issues/41)

pablobm:
@locks - I believe the first of the two checkboxes can be ticked as per https://github.com/ember-learn/guides-source/pull/326

acorncom:
thanks @pablobm. @locks anything else we want to do here?

PR #326, opened 2018-12-13
Expand addons section

pablobm:
My attempt at expanding the addons section, as per https://github.com/ember-learn/guides-source/issues/320.

This only covers the first of the two acceptance criteria on the ticket ("Expand the Addons section..."), as I don't have the expertise to address the second one.


pablobm:
Oh, I wasn't even aware of the section at the Ember CLI docs! I agree there's a balance between de-duplication and immediacy that is required here. De-duplicate the examples sounds good, as they are more likely to go out of date.

Any thoughts on capitalisation of the word "addon"? I'm using "Ember Addons" and then standalone "addons", and I'm not sure if there's a convention I should be following.

jenweber:
Thanks for this PR! That's the same pattern I followed for the cli docs.

PR #344, opened 2018-12-27
[WIP] Fixing index.md files for 2.3

mansona:
This is to fix issues where you get 404s on some of the pages of the guides. Every folder needs an index.md file and I have implemented the redirects necessary for any pages that have been moved.

This PR is part of a series of PRs to fix each of the versions in turn. I am doing it like this to make it easier to review. The best way to review these PRs is to look at each of the commits individually so that you can see that there has been no change to the files (just moves) in the first commit and then just adding the redirects in the second commit 👍


mansona:
You may have seen my last few PRs that are very much like this one 😂 but this one has thrown up an issue 😭 

Essentially the first page in the "Getting Started" link isn't the index

image 2018-12-27 at 5 04 01 p m

image 2018-12-27 at 5 06 01 p m

To be honest I don't really know what to do here. We could correct it but I don't know if it makes sense to fix it because we can't "redirect" to the new place, it is more like changing what the link means 🤔 

Also we don't actually need to fix it. The point of these PRs is to make sure there is an index file, it doesn't actually need to be the first one 🤔

MelSumner:
@mansona - we also need this for SEO

I think we need to figure out the specific instances where this is occurring and write issues for each. In cases where someone might have a link we can't break, and it's reasonable, we can add a link to the server-side re-directs.

jenweber:
@mansona is this still WIP or should it be merged?

PR #346, opened 2018-12-29
Fixing index.md files for 1.13

mansona:
This supersedes #340


jenweber:
Ran this locally and it doesn't explode, so trusting that the server behavior will be what you intend.

PR #355, opened 2019-01-09
Fixes 'customize a components element' link

scottswezey:
The previous version linked to `https://guides.emberjs.com/release/components/defining-a-component/customizing-a-components-element/`, which showed a 404 error instead of `https://guides.emberjs.com/release/components/customizing-a-components-element/`, which is the intended page.

I see from the contribution guidelines that links should be in another format, but have retained the current markdown style to be consistent with the current document.

### DIFF
--- a/guides/v3.6.0/components/defining-a-component.md
+++ b/guides/v3.6.0/components/defining-a-component.md
@@ -49,7 +49,7 @@ Each component is backed by an element under the hood. By default,
 Ember will use a `<div>` element to contain your component's template.
 To learn how to change the element Ember uses for your component, see
 [Customizing a Component's
-Element](./customizing-a-components-element/).
+Element](../customizing-a-components-element/).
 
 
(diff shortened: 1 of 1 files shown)


scottswezey:
If needed, the live version of this link can be found by visiting 
https://guides.emberjs.com/release/components/defining-a-component/#toc_defining-a-component-subclass -- It's the first line/link directly above the linked header/section.

mansona:
As I mentioned in https://github.com/ember-learn/guides-app/issues/299 I'm more concerned why this wasn't picked up by the tests. I would prefer this fix to also accompany a fix to the tests that shows it was failing before and passing after 👍

locks:
Thanks for sending the PR @scottswezey. As mentioned by Chris, this exposed a flaw in our testing facilities that we want to address so we have some assurance that we properly fix the issue and avoid it in the future.
I am closing this and opening an issue to move it forward!

PR #356, opened 2019-01-09
Update CONTRIBUTING.md

locks:


locks:
Confirm! We added imports a while back and should generally strive to provide as complete a code sample as it makes sense so it is easier for people who are unfamiliar with app code.

mansona:
@locks also it makes it easier for tooling to lint the code examples 😂

PR #365, opened 2019-01-21
Single repo implementation

mansona:
This PR moves us into a "single repo" mode just like the new [Ember Cli Guides](https://github.com/ember-learn/cli-guides) project. This has a number of benefits over the two repo process we have been doing for the last year: 

- We now have the ability to preview each PR and see how it would look before we merge (you can see this with the Netlify test app below)
- removes the need to deploy the guides-source to npm and then merge that change into guides-app
- any time we merge a PR into master it will automatically be deployed
- vastly simplifies the contributing guide for anyone who wants to run the guides locally while editing (no need to do any repo linking)
- we are now using actually shared implementations for the Guides and the Cli Guides, and will be using the same for the upcoming Tutorials app too 👍 

It's also worth noting that merging this doesn't make it **go live** right away. I wouldn't merge this PR unless we have decided that we intended to go live realatively soon, but we have time to figure things out for a little while before going live.


jenweber:
Note that we will need to take care of a couple odds and ends before we turn on Continuous Deployment (which will happen when we merge this?)

MelSumner:
can you clarify what this is combining? I have strong opinions against merging the guides-source and guides-app- for maintenance reasons I think it's strategically important for us to separate the content from the presentation layer.

MelSumner:
@mansona looks like there's some conflicts to resolve first.

PR #390, opened 2019-02-05
Update tutorial to be more consistent with itself and the super-rentals application

dayton-bobbitt:
Make use of async / await when using the `settled` helper to be consistent with the implementation other asynchronous test helpers.


guides/v3.7.0/tutorial/autocomplete-component.md
```diff
@@ -512,11 +507,10 @@ test('should update with matching listings', async function (assert) {
...
   // keyup event to invoke an action that will cause the list to be filtered
   await triggerKeyEvent(this.element.querySelector('.list-filter input'), "keyup", 83);
+  await settled();
 
-  return settled().then(() => {
-    assert.equal(this.element.querySelectorAll('.city').length, 1, 'One result returned');
-    assert.equal(this.element.querySelector('.city').textContent.trim(), 'San Francisco');
-  });
+  assert.equal(this.element.querySelectorAll('.city').length, 1, 'One result returned');
+  assert.equal(this.element.querySelector('.city').textContent.trim(), 'San Francisco');
```
Turbo87: the code change here is correct, though I'm wondering why we even use `settled()` here. `triggerKeyEvent()` already calls `settled()` before returning, so this seems redundant.

guides/v3.7.0/tutorial/autocomplete-component.md
```diff
@@ -413,15 +413,10 @@ module('Integration | Component | list-filter', function(hooks) {
...
+waits for all asynchronous tasks to complete.
 
-Ember's [settled helper](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#settled)
-waits for all asynchronous tasks to complete before running the given function callback.
-It returns a promise that we also return from the test.
-
-If you return a promise from a QUnit test, the test will wait to finish until that promise is resolved.
-In this case our test completes when the `settled` helper decides that processing is finished,
-and the function we provide that asserts the resulting state is completed.
+In this case our assertions will not be made until the `settled` helper decides that processing is finished.
```
Turbo87: I think I prefer the old phrasing of this paragraph 🤔

dayton-bobbitt: I agree but the old content seemed to no longer apply directly to the example because we are no longer returning a promise from the QUnit tests. I wasn't sure quite how to reword it though.

guides/v3.7.0/tutorial/autocomplete-component.md
dayton-bobbitt: I believe this example is the introduction of the `settled` helper. Since it’s redundant, maybe it would make since to remove it and move the introduction of `settled` to a test where it’s needed?

guides/v3.7.0/tutorial/autocomplete-component.md
Turbo87: that's not entirely true. we *are* still returning a Promise, it's just not directly visible because we're using an async function.

guides/v3.7.0/tutorial/autocomplete-component.md
Turbo87: @locks ☝️ ?

guides/v3.7.0/tutorial/autocomplete-component.md
dayton-bobbitt: @Turbo87 I reverted the content back to its original wording.

guides/v3.7.0/tutorial/autocomplete-component.md
dayton-bobbitt: @locks @Turbo87 If you think it makes sense, I can remove the `settled` helper that is used after `tiggerKeyEvent()`

jenweber:
Before this PR can be merged, we will also need a PR to update the sample app: https://github.com/ember-learn/super-rentals/blob/master/tests/integration/components/list-filter-test.js

Is that something you can help with?

Otherwise it looks good to me! Thanks!

dayton-bobbitt:
Sure, I can do that!

dayton-bobbitt:
@jenweber Just waiting on the super-rentals PR to be approved and merged https://github.com/ember-learn/super-rentals/pull/117

guides/v3.7.0/tutorial/autocomplete-component.md
rwjblue: Turns out I commented on the same issue over in https://github.com/ember-learn/super-rentals/pull/117. IMHO, we should remove these superfluous `await settled()` calls (any `await settled()` after another async test helper from `@ember/test-helpers` are generally not needed).

dayton-bobbitt: I'm fine removing `await settled()`, but as I mentioned above, I don't believe there is another time in the tutorial where that test helper is introduced.

Previously, were waiting on a response from @locks about his opinion on removing `await settled()`, so in the meantime, I updated the super-rentals application to be consistent with this PR as @jenweber suggested.

jenweber:
Somehow, the Super Rentals demo app PR got merged a long time ago and this didn't. Our bad. Thanks again for this work! I added a fix for missing code highlighting and will merge shortly.

PR #392, opened 2019-02-07
TOC | Re-order and re-title

locks:
Closes https://github.com/ember-learn/guides-source/issues/418.
Closes https://github.com/ember-learn/guides-source/issues/402.

What it says on the tin.

### DIFF
--- a/guides/release/pages.yml
+++ b/guides/release/pages.yml
@@ -68,32 +68,7 @@
     - title: "Enumerables"
       url: 'enumerables'
 
-- title: "Routing"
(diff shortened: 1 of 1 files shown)


jenweber:
Looks good, many thanks for putting this together!

I have a little bit of content that might help tie these sections together in this ordering. Up to you if it's useful - I'll make some PRs. I spiked a reordering (in a non-mergeable way) to test out the flow while I was writing the RFC, then I wrote a couple transition sections to see how it would work out.

PR #400, opened 2019-02-07
models: add note about the store injection

knownasilya:
Basically that it's automatically available on routes.

Not sure if this is the best place for it. Was looking for like a blockquote example, but didn't find one, so maybe there is some syntax to make this stand out a bit.

### DIFF
--- a/guides/release/models/index.md
+++ b/guides/release/models/index.md
@@ -175,14 +175,14 @@ Good Ember apps take a different approach. Ember Data gives you a single
 Routes and their corresponding controllers can ask the store for models, and the store is
 responsible for knowing how to fetch them.
 
-Note: The Ember Data store is automatically injected on every Route instance as `this.store`.
-
 It also means that the store can detect that two different components
 are asking for the same model, allowing your app to only fetch the data
 from the server once. You can think of the store as a read-through cache
 for your app's models. Both routes and their corresponding controllers have access to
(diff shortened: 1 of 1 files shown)


guides/v3.7.0/models/index.md
```diff
@@ -182,6 +182,8 @@ for your app's models. Both routes and their corresponding controllers have acce
 this shared store; when they need to display or modify a model, they
 first ask the store for it.
 
+Note: The Ember Data store is automatically injected on every Route instance as `this.store`.
```
locks:
```suggestion

<div class="cta">
  <div class="cta-note">
    <div class="cta-note-body">
      <div class="cta-note-heading">Zoey says...</div>
      <div class="cta-note-message">
        Ember Data injects the store service in every route and controller,
        so you can access it as `this.store`!
      </div>
    </div>
    <img src="/images/mascots/zoey.png" role="presentation" alt="Ember Mascot">
  </div>
</div>
```

I got the markup from Chris 😁 I made a tweak to the prose, what do you think?

knownasilya: Wait, controllers get the store too?

locks: I believe so: https://github.com/emberjs/data/blob/master/addon/setup-container.js#L68

knownasilya:
@locks

PR #401, opened 2019-02-08
Angle bracket invocation guide

locks:


guides/v3.7.0/reference/syntax-conversion-guide.md
```diff
@@ -1,24 +1,27 @@
 ## Angle Bracket Syntax
 
-The [Angle Bracket Syntax](https://github.com/emberjs/rfcs/blob/master/text/0311-angle-bracket-invocation.md) is an alternative style of invoking components in templates. The difference between the Angle Bracket Syntax and the Classic Invocation Syntax is purely syntactical and does not affect the semantics of invoking a component.
+The [angle bracket invocation syntax (ABIS)](https://github.com/emberjs/rfcs/blob/master/text/0311-angle-bracket-invocation.md) is an alternative style of invoking components in templates.
```
mansona: I'm pretty sure it's the convention to capitalise the letters of a phrase you're about to create an abbreviation from 🤔 so that would be: 

```suggestion
The [Angle Bracket Invocation Syntax (ABIS)](https://github.com/emberjs/rfcs/blob/master/text/0311-angle-bracket-invocation.md) is an alternative style of invoking components in templates.
```

guides/v3.7.0/reference/syntax-conversion-guide.md
```diff
@@ -1,24 +1,27 @@
...
+The difference between ABIS and classic invocation syntax (CIS) is syntactical and does not affect the semantics of invoking a component.
```
mansona: same as above: 

```suggestion
The difference between ABIS and Classic Invocation Syntax (CIS) is syntactical and does not affect the semantics of invoking a component.
```

guides/v3.7.0/reference/syntax-conversion-guide.md
```diff
@@ -1,24 +1,27 @@
...
+Classic invocation syntax may also be referred to as curly invocation syntax.
 
-**Classical Invocation Syntax:**
+**Classic invocation syntax:**
```
mansona: I would revert this to keep consistent with above 👍

guides/v3.7.0/reference/syntax-conversion-guide.md
```diff
@@ -1,24 +1,27 @@
...
 ```handlebars
 {{site-header user=this.user class=(if this.user.isAdmin "admin")}}
 ```
 
-**Angle Bracket Syntax:**
+**Angle bracket invocation syntax:**
```
mansona:
```suggestion
**Angle Bracket Invocation Syntax:**
```

guides/v3.7.0/reference/syntax-conversion-guide.md
```diff
@@ -52,23 +55,57 @@ You can apply regular HTML attributes like `class`, `id`, `aria-role`, etc. when
...
+});
+```
+
+```handlebars
+<MyGreeting @params={{array "Hello" "World"}}>
+```
+
+### What is happening with the Classic Invocation Syntax?
+
+Classic invocation syntax—which uses “curlies” `{{}}` instead of angle brackets `<>`–is here to stay.
```
MelSumner: is there a reason you have "curlies" here instead of handlebars? handlebars and angle brackets seem more consistent with each other.

guides/v3.7.0/reference/syntax-conversion-guide.md
```diff
@@ -52,23 +55,57 @@ You can apply regular HTML attributes like `class`, `id`, `aria-role`, etc. when
 </SuperSelect>
 ```
 
-### Determining the Argument Scope
+### Positional parameters
 
-The fundamental change is that the scope of arguments passed in and properties local to the component are no longer mashed together. There is a clear boundary between arguments passed down from the parent and arguments that is tracked in the local component.
+The classic invocation syntax supports passing arguments to the component by their position.
+In the following example, `"first"` and `"second"` are positional parameters:
```
chrisrng:
```suggestion
In the following example, `"hello"` and `"world"` are positional parameters:
```
Shouldnt this be hello and world to match the example below?
Should this be

guides/v3.7.0/reference/syntax-conversion-guide.md
locks: I am going the opposite way and not use abbreviations at all. Coming back to the material and reading it, I don't think it is improved by the use of abbvs.

guides/v3.7.0/reference/syntax-conversion-guide.md
locks: I kept it from the original, will update.

guides/v3.7.0/reference/syntax-conversion-guide.md
mansona: Yes I agree 👍

Issue #398, opened 2019-02-07
Update Octane guides sub-section: Components > Adding actions


MelSumner:
## Goal
Update the "adding actions" sub-section of the Components section of the Octane guides so it includes how to do components in a way that is consistent with the Octane edition of Ember. 

## TODO
- [ ] Translate to native classes
  - [ ] class definition
  - [ ] `@action` decorator on actions
- [ ] Update templates not to use string actions
- [ ] Add a header about string actions

## References
This issue is a part of the [Component section tracking issue](https://github.com/ember-learn/guides-source/issues/395)

knownasilya:
@locks for `* Add a header about string actions` what specifically should this cover?

knownasilya:
And making sure that this is the right file: https://github.com/ember-learn/guides-source/blob/octane/guides/release/components/triggering-changes-with-actions.md

locks:
> @locks for `* Add a header about string actions` what specifically should this cover?

@knownasilya what I meant by it is that doing `{{action 'myAction'}}` is considered legacy in the Octane universe, but we still have to document it for interop reasons.

knownasilya:
Sort of a migration guide and warning, got it.

PR #430, opened 2019-02-12
Components: defining a component

locks:
Closes https://github.com/ember-learn/guides-source/issues/396.

PR #432, opened 2019-02-12
components(actions): start converting to class and @action decorator

knownasilya:
Resolves https://github.com/ember-learn/guides-source/issues/398

- [x] Class
- [x] GC/tracked
- [x] onclick vs element modifier
- [x] Legacy String Actions

### DIFF
--- a/guides/release/components/triggering-changes-with-actions.md
+++ b/guides/release/components/triggering-changes-with-actions.md
@@ -311,20 +311,20 @@ we will use a property `confirmValue` to represent that argument and pass it to 
 
 ```javascript {data-filename=app/components/button-with-confirmation.js}
 import Component from '@ember/component';
+import { action } from '@ember/object';
 
-export default Component.extend({
-  actions: {
-    //...
-    submitConfirm() {
-      // call onConfirm with a second argument
-      let promise = this.onConfirm(this.confirmValue);
-      promise.then(() => {
-        this.set('confirmShown', false);
-      });
-    },
-    //...
+export default class ButtonWithConfirmation extends Component {
+  @action
+  submitConfirm() {
+    // call `onConfirm` with a second argument
+    let promise = this.onConfirm(this.confirmValue);
+    promise.then(() => {
+      this.set('confirmShown', false);
+    });
   }
-});
+  
+  //...
+}
 ```
 
 In order for `confirmValue` to take on the value of the message text,
@@ -333,12 +333,13 @@ To accomplish this,
 we'll first modify the component so that it can be used in block form and we will [yield](../wrapping-content-in-a-component/) `confirmValue` to the block within the `"confirmDialog"` element:
 
 ```handlebars {data-filename=app/templates/components/button-with-confirmation.hbs}
-<button {{action "launchConfirmDialog"}}>{{this.text}}</button>
+<button {{action this.launchConfirmDialog}}>{{this.text}}</button>
+
 {{#if this.confirmShown}}
   <div class="confirm-dialog">
     {{yield this.confirmValue}}
-    <button class="confirm-submit" {{action "submitConfirm"}}>OK</button>
-    <button class="confirm-cancel" {{action "cancelConfirm"}}>Cancel</button>
+    <button class="confirm-submit" {{action this.submitConfirm}}>OK</button>
+    <button class="confirm-cancel" {{action this.cancelConfirm}}>Cancel</button>
   </div>
 {{/if}}
 ```
@@ -348,9 +349,9 @@ we can now use the component in `SendMessage` to wrap a text input element whose
 
 ```handlebars {data-filename=app/templates/components/send-message.hbs}
 <ButtonWithConfirmation
-    @text="Click to send your message."
-    @onConfirm={{action "sendMessage" "info"}}
-    as |confirmValue|>
+  @text="Click to send your message."
+  @onConfirm={{action this.sendMessage "info"}}
+as |confirmValue|>
   {{input value=confirmValue}}
 </ButtonWithConfirmation>
 ```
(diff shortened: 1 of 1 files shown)


knownasilya:
cc @locks

guides/release/components/triggering-changes-with-actions.md
```diff
@@ -103,34 +103,37 @@ in the child component we will implement the logic to confirm that the user want
...
 
 ```handlebars {data-filename=app/templates/components/button-with-confirmation.hbs}
-<button {{action "launchConfirmDialog"}}>{{this.text}}</button>
+<button {{action this.launchConfirmDialog}}>{{this.text}}</button>
+
 {{#if this.confirmShown}}
   <div class="confirm-dialog">
-    <button class="confirm-submit" {{action "submitConfirm"}}>OK</button>
-    <button class="confirm-cancel" {{action "cancelConfirm"}}>Cancel</button>
+    <button class="confirm-submit" {{action this.submitConfirm}}>OK</button>
```
knownasilya: @locks @pzuraq should we be using the modifier version of `action` in the examples, since this won't bubble anyway and is less clear then `onclick={{this.submitConfirm}}`.

guides/release/components/triggering-changes-with-actions.md
```diff
@@ -365,37 +370,37 @@ Actions can be invoked on objects other than the component directly from the tem
...
 
 We can tell the action to invoke the `sendMessage` action directly on the messaging service with the `target` attribute.
 
 ```handlebars {data-filename=app/templates/components/send-message.hbs}
 <ButtonWithConfirmation
-    @text="Click to send your message."
-    @onConfirm={{action "sendMessage" "info" target=this.messaging}}
-    as |confirmValue|>
+  @text="Click to send your message."
+  @onConfirm={{action this.messaging.sendMessage "info"}}
```
knownasilya: @pzuraq is this correct, nothing special is required any more?

pzuraq: `sendMessage` needs to be decorated with `@action`, if it is then this should work

guides/release/components/triggering-changes-with-actions.md
pzuraq: I'm not sure about that. I think ideally if we used a modifier version, we would use something like the [on modifier](https://github.com/buschtoens/ember-on-modifier#readme), `{{action}}` has too many quirks and acts like jquery action handlers which are not ideal. Since that's not mainline, I would lean toward `onclick=` (cc @tomdale I believe he has some opinions here too)

guides/release/components/triggering-changes-with-actions.md
```diff
@@ -103,34 +103,37 @@ in the child component we will implement the logic to confirm that the user want
 
 ```javascript {data-filename=app/components/button-with-confirmation.js}
```
knownasilya: @locks @pzuraq if we use GC, should the import be something else? Do we know this yet?

guides/release/components/triggering-changes-with-actions.md
```diff
@@ -103,34 +103,37 @@ in the child component we will implement the logic to confirm that the user want
...
 
-export default Component.extend({
-  actions: {
-    launchConfirmDialog() {
-      this.set('confirmShown', true);
-    },
+export default class ButtonWithConfirmation extends Component {
+  @action
+  launchConfirmDialog() {
+    this.set('confirmShown', true);
```
knownasilya: @locks @pzuraq There is no `this.set` in GCs, correct? Also I'd have to use `@tracked`?

guides/release/components/triggering-changes-with-actions.md
locks: Go with `import Component from '@glimmer/component';` for now!

guides/release/components/triggering-changes-with-actions.md
locks: Confirm to both!

knownasilya: `import Component, { tracked } from '@glimmer/component';`?

locks: I believe so.

guides/release/components/triggering-changes-with-actions.md
knownasilya: I converted to `onclick=...`. Let me know if I should change it to something else.

guides/release/components/triggering-changes-with-actions.md
```diff
@@ -454,39 +465,67 @@ For example, say we want to move account deletion from the `UserProfile` compone
...
+  @onConfirm={{@deleteCurrentUser}}
 />
 ```
 
-Note that `deleteCurrentUser` is no longer in quotes here as opposed to [previously](#toc_passing-the-action-to-the-component). Quotes are used to initially pass the action down the component tree, but at every subsequent level you are instead passing the actual function reference (without quotes) in the action helper.
+Note that `deleteCurrentUser` is now prepended with `@` as opposed to `this.` [previously](#toc_passing-the-action-to-the-component).
+
+Now when you confirm deletion, the action goes straight to the `SystemPreferencesEditor` to be handled in its local context.
+
+## Legacy String Actions
```
knownasilya: @locks @pzuraq needs review

knownasilya:
Needs review

pzuraq:
I think this may be in conflict with the work @locks is doing in #430, I’ll defer to him here

guides/release/components/triggering-changes-with-actions.md
```diff
@@ -504,7 +504,7 @@ Note that `deleteCurrentUser` is now prepended with `@` as opposed to `this.` [p
 
 Now when you confirm deletion, the action goes straight to the `SystemPreferencesEditor` to be handled in its local context.
 
-## Legacy String Actions
+## String Action Syntax
```
knownasilya: 👍 @locks

PR #446, opened 2019-02-13
change header, rephrase ES6 definition

### LINKED ISSUE #301: header uses ES6 to describe modules, description leads with ES2015 modules
ES2015 modules is certainly wordier, but more technically accurate. [The header](https://github.com/ember-learn/guides-source/blob/master/guides/v3.5.0/tutorial/ember-cli.md#es6-modules) says
> ES6 Modules

but the line after says

> ECMAScript 2015 (ES2015 for short or previously known as ES6) modules 

To me, this arrangement of terms is distracting. 

# proposed change
I propose we make the header just `Modules`. This would be similar to [babel's documentation](https://babeljs.io/docs/en/learn#modules).

efx:
Fix #301


efx:
Took the liberty of renaming `previously known` to `popularly known` as I think that better describes how ES6 came about and is used.

locks:
@ember-learn/learning-team are we ok with the anchor changing?

efx:
@jenweber so keep the heading `ES6 Modules`? I agree this will make the resource seem outdated as that is why I changed it.

efx:
I just went aheaded and reverted that part.

jenweber:
Oops, sorry @efx, I read the diff backwards. Changing it to Modules makes sense! It fits the conditions where I think we should go ahead and change anchors. I put it back how your PR originally was before merging.

efx:
ah, no worries. thanks @jenweber !

PR #448, opened 2019-02-13
Update template and Add Analytics (for release)

mansona:


locks:
Seems like a rebase is in order.

PR #465, opened 2019-02-15
[Octane] Revert removal of sections

pzuraq:
This reverts commit c8f68cbb6cf5528c2feb55e776e87128053bf8ed.

### DIFF
--- a/guides/release/pages.yml
+++ b/guides/release/pages.yml
@@ -14,6 +14,8 @@
       url: "index"
     - title: "Core Concepts"
       url: "core-concepts"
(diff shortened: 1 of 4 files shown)


guides/release/pages.yml
```diff
@@ -48,6 +50,10 @@
   pages:
     - title: "Objects in Ember"
```
MelSumner: this can be removed from the Table of Contents, just don't remove the file themselves.

pzuraq: Gotcha

PR #501, opened 2019-02-20
Add dashes to test module names

ondrejsevcik:
I was under the impression that module test names should contain `-` in the component name, shouldn't they?


jenweber:
Hey @ondrejsevcik! Thanks for this! If you mark it as a not-draft, I think this is good to merge. I tested it in a 3.7 app and tests do have a dash in them when generated via the CLI.

ondrejsevcik:
Thanks for the quick response. It's not draft anymore.

jenweber:
Merged, thank you!

PR #509, opened 2019-02-21
Tutorial stub service to prevent ajax

### LINKED ISSUE #347: Missing tutorial information
From https://github.com/emberjs/ember.js/issues/17417:

> When running the tests on the [3.5.0 Services tutorial page](https://guides.emberjs.com/v3.5.0/tutorial/service/) I was getting multiple `Ajax authorization failed` errors on tests I hadn't touched. Naturally this had to do with the Simple Leaflet Maps service. The problem was that while the app was started with `LEAFLET_MAPS_API_KEY=<your key here> ember s` it didn't say anything about this in the test. it just says "restart `ember test --server` whenever you add a new service."
But some of the acceptance tests, since they are running the service (even if they are not testing it), need access to that environment variable to. The tutorial should be clear that the tests should be restarted with `LEAFLET_MAPS_API_KEY=<your key here> ember test --server`
FYI, running Windows 7, Bash command line.
> - @BenReilly

toddjordan:
Fixes #347 

Adds stub service to integration test so that leaflet ajax calls don't happen during test run.

image


jenweber:
It looks like this was already in the super-rentals app, but just not explained before.

Thanks Todd! Merged.

PR #519, opened 2019-02-24
Fix error callback written with then instead of catch

CodingItWrong:
In Defining Your Routes > Wildcard / globbing routes, there is a code sample that's intended to illustrate a promise with a success and error callback, but both are written with `then`:

image

This seems to be a bug, because the second `then` would receive a successful result from the first `then`, not an error. It seems that `catch` is intended instead. This PR corrects that:

image


acorncom:
Good catch!

PR #536, opened 2019-03-01
[OCTANE][FEAT] State Management

pzuraq:
Adds the State Management section. This section definitely will need polish in the future, I imagine more thorough discussions of state could be helpful (tried my best!). We also need to add a section on computed properties, @jenweber this would be great for anyone who has some time to volunteer since they're much more well understood.

Feedback definitely appreciated on the Patterns for State section (and all of it really!)


MelSumner:
@pzuraq https://github.com/ember-learn/guides-source/issues/537

PR #549, opened 2019-03-01
[OCTANE] [CHORE] Quickstart JS to use Class Syntax #541

### LINKED ISSUE #541: Change Quickstart JavaScript code blocks to use Class Syntax
We'd love to have a volunteer for this! Please see [Octane MVP](https://github.com/ember-learn/guides-source/issues/394) for instructions about where to make your PR. Please only volunteer for this issue if you can open a PR by the end of the next day. Thanks for your help!

### To-do
For this issue, we should only make "class syntax" changes to the code blocks within `guides/release/getting-started` markdown files. Later PRs will fix up the written explanations to go with them. Here's an example of converting to Class syntax:

```js
// Before
import Controller from '@ember/controller';

export default Controller.extend({
  title: 'hello-world.app',
});
```

```js
// After
import Controller from '@ember/controller';

  title = 'hello-world.app';
}
```

- [ ] branch from the `octane` branch of the guides
- [ ] See [this guide's Native Classes](https://github.com/ember-learn/guides-source/blob/octane/guides/release/upgrading/editions.md#native-classes) explanation to learn about the changes you will have to make
- [ ] Make changes to the code blocks to use class syntax
- [ ] PR to the `octane` branch

rajasegar:
Fixes #541

1. Change Route example to class syntax in core concepts
screen shot 2019-03-01 at 6 04 44 pm

2. Change PeopleList component to use class syntax
screen shot 2019-03-01 at 6 04 56 pm

3. didRender example to class syntax
screen shot 2019-03-01 at 6 03 25 pm

### DIFF
--- a/guides/release/getting-started/core-concepts.md
+++ b/guides/release/getting-started/core-concepts.md
@@ -74,10 +74,10 @@ In the following example, the [`didRender()`](https://emberjs.com/api/ember/rele
 ```javascript {data-filename=/app/components/foo-did-render-example.js}
 import Component from '@ember/component';
 
-export default Component.extend({
+export default class FooDidRenderExample extends Component {
   didRender() {
-    this._super(...arguments);
+    super(...arguments);
     console.log('I rendered!');
   }
-});
+}
 ```
--- a/guides/release/getting-started/core-concepts.md
+++ b/guides/release/getting-started/core-concepts.md
@@ -76,7 +76,7 @@ import Component from '@ember/component';
 
 export default class FooDidRenderExample extends Component {
   didRender() {
-    super(...arguments);
+    super.didRender(...arguments);
     console.log('I rendered!');
   }
 }
--- a/guides/release/getting-started/quick-start.md
+++ b/guides/release/getting-started/quick-start.md
@@ -131,11 +131,11 @@ We'll take the code created for us by the generator and add a `model()` method t
 ```javascript {data-filename="app/routes/scientists.js" data-diff="+4,+5,+6"}
 import Route from '@ember/routing/route';
 
-export default Route.extend({
+export default class ScientistsRoute extends Route {
   model() {
     return ['Marie Curie', 'Mae Jemison', 'Albert Hofmann'];
   }
-});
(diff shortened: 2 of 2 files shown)


guides/release/getting-started/core-concepts.md
```diff
@@ -74,10 +74,10 @@ In the following example, the [`didRender()`](https://emberjs.com/api/ember/rele
 ```javascript {data-filename=/app/components/foo-did-render-example.js}
 
+export default class FooDidRenderExample extends Component {
   didRender() {
-    this._super(...arguments);
+    super(...arguments);
```
ddoria921: I'm not that familiar with Class syntax, but should this be `super.didRender(...arguments)`? (I think I read that somewhere but I could be mistaken)

rajasegar: But super is a function right, how can you call `super.didRender`, generally to call the base class function we will be using `super()` with whatever arguments from the child class

rajasegar: Found something useful on MDN
Your syntax works on static functions i guess
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super

But from the implementation of didRender, i don't think that is a static function
https://github.com/emberjs/ember.js/blob/30137796af42c63b28ead127cba0e43e45a773c1/packages/%40ember/-internals/glimmer/lib/component.ts#L798

Please let me know whether i am going in the right direction...

jamescdavis: This should definitely be 
```suggestion
    super.didRender(...arguments);
```
This does not only work on static methods. I think the MDN article is just noting that it also works on static methods. It's strange that they don't mention it works for regular class methods. It [definitely works](https://ember-twiddle.com/81eb88e1b46a337d9c9af46428b3467e?openFiles=components.child-component.js%2Ctemplates.components.child-component.hbs).

rajasegar:
@jamescdavis @ddoria921 Thanks guys for the suggestion, committed the change

Issue #540, opened 2019-03-01
Change Tutorial JavaScript code blocks to use Class Syntax


jenweber:
We'd love to have a volunteer for this! Please see [Octane MVP](https://github.com/ember-learn/guides-source/issues/394) for instructions about where to make your PR. Please only volunteer for this issue if you can open a PR with at least part of the work by the end of the next day. Thanks for your help!

### To-do
For this issue, we should only make "class syntax" changes to the code blocks within `guides/release/tutorial` markdown files. Later PRs will fix up the written explanations to go with them. Here's an example of converting to Class syntax:

```js
// Before
import Controller from '@ember/controller';

export default Controller.extend({
  title: 'hello-world.app',
});
```

```js
// After
import Controller from '@ember/controller';

export default class ApplicationController extends Controller {
  title = 'hello-world.app';
}
```

- [ ] branch from the `octane` branch of the guides
- [ ] See [this guide's Native Classes](https://github.com/ember-learn/guides-source/blob/octane/guides/release/upgrading/editions.md#native-classes) explanation to learn about the changes you will have to make
- [ ] Make changes to the code blocks to use class syntax
- [ ] PR to the `octane` branch

muziejus:
I volunteer. Is that sufficient warning to ensure I don't stomp on someone else's work?

jenweber:
Yep! Thanks for the help! Let me know if you have any questions along the way.

jamescdavis:
I volunteer to CR this when the PR is opened. @muziejus, please ping me on the PR.

muziejus:
My one question so far, I think, is about whether I should _also_ be adding decorators?

```javascript
export default class BlahComponent extends Component {
  actions = {
    handleClick() {
     // ...
    }
  }
}
```

looks odd to me.  It wouldn't be much...

rajasegar:
@muziejus Yes i think, should be
```js
export default class BlahComponent extends Component {
  @action
    handleClick() {
     // ...
    }

}
```

muziejus:
OK, I'll add decorators then.

Another question: tests fail on broken links outside of my realm… and the one inside my realm seems unfixable:

```
3260 passing (5s)
  4 failing

  1) check all links in markdown files
       processing guides/release/components/block-params.md:
     AssertionError: [ { fileToFix: 'guides/release/components/block-params.md',
    badLink: '../passing-properties-to-a-component/' } ]: expected [ Array(1) ] to be empty
[...]

Issue #543, opened 2019-03-01
Change Ember Data JavaScript code blocks to use Class Syntax


jenweber:
We'd love to have a volunteer for this! Please see [Octane MVP](https://github.com/ember-learn/guides-source/issues/394) for instructions about where to make your PR. Please only volunteer for this issue if you can open a PR by the end of the next day. Thanks for your help!

### To-do
For this issue, we should only make "class syntax" changes to the code blocks within `guides/release/models` markdown files. Later PRs will fix up the written explanations to go with them. Here's an example of converting to Class syntax:

```js
// Before
import Controller from '@ember/controller';

export default Controller.extend({
  title: 'hello-world.app',
});
```

```js
// After
import Controller from '@ember/controller';

export default class ApplicationController extends Controller {
  title = 'hello-world.app';
}
```

- [ ] branch from the `octane` branch of the guides
- [ ] See [this guide's Native Classes](https://github.com/ember-learn/guides-source/blob/octane/guides/release/upgrading/editions.md#native-classes) explanation to learn about the changes you will have to make
- [ ] Make changes to the code blocks to use class syntax
- [ ] PR to the `octane` branch

efx:
I'll grab this one.

jenweber:
Thanks Eli!!!

acorncom:
@efx anything else needed here?

efx:
@acorncom Tutorials related to ember-data haven't been converted IIRC. I'll check on this today and do a once over of the current status.

efx:
@jenweber @acorncom this should be done now. There are other aspects of decorators and native classes to address as discussed in #603, but they fit better as part of other PR's / contained issues.

jenweber:
Thanks everyone for the help! Closed.

PR #551, opened 2019-03-01
Change Tutorial examples to use Class Syntax (#540)

muziejus:
@jamescdavis


muziejus:
This fixes #540

guides/release/tutorial/service.md
```diff
@@ -147,20 +146,20 @@ This function runs during the component render, after the component's markup get
...
-export default Component.extend({
-  classNames: ['map-container'],
-  mapElement: service(),
+export default class LocationMapComponent extends Component {
+  classNames = ['map-container'];
+  @service mapElement;
 
   didInsertElement() {
-    this._super(...arguments);
+    super();
```
jamescdavis:
```suggestion
    super.didInsertElement();
```

guides/release/tutorial/autocomplete-component.md
```diff
@@ -228,60 +227,59 @@ We will update the results on screen only if the original filter value and the c
...
-  classNames: ['list-filter'],
-  value: '',
+export default class ListFilterComponent extends Component {
+  classNames = ['list-filter'];
+  value = '';
 
-  init() {
-    this._super(...arguments);
+  constructor() {
+    super();
```
jamescdavis:
```suggestion
    super(...arguments);
```

guides/release/tutorial/service.md
```diff
@@ -87,19 +87,19 @@ otherwise we will create a new HTML element and call our Leaflet map service to
...
+  @service map;
 
-  init() {
+  constructor() {
     if (!this.cachedMaps) {
       set(this, 'cachedMaps', {});
     }
-    this._super(...arguments);
-  },
+    super();
```
jamescdavis: Works without passing constructor arguments (presently), but recommend passing them along in case something changes in the future.
```suggestion
    super(...arguments);
```

guides/release/tutorial/autocomplete-component.md
```diff
@@ -79,25 +79,24 @@ The `handleFilterEntry` action will apply the search term filter to the list of
 
 ```javascript {data-filename=app/components/list-filter.js}
+import { action } from '@ember-decorators/object';
 
-  classNames: ['list-filter'],
-  value: '',
+export default class ListFilterComponent extends Component {
+  classNames = ['list-filter'];
```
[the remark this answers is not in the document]
muziejus: I've added them. I forgot about this decorator 😃

guides/release/tutorial/autocomplete-component.md
```diff
@@ -79,25 +79,24 @@ The `handleFilterEntry` action will apply the search term filter to the list of
 
 ```javascript {data-filename=app/components/list-filter.js}
+import { action } from '@ember-decorators/object';
 
-  classNames: ['list-filter'],
-  value: '',
+export default class ListFilterComponent extends Component {
```
efx: Not to muddy the waters here too much, but I thought native class syntax should only apply to Glimmer components: https://github.com/ember-learn/guides-source/blob/octane/guides/release/upgrading/editions.md#glimmer-components. Does this seem correct?

MelSumner: @pzuraq can you review this bit please?

pzuraq: yes, native class syntax applies to everything _except_ classic components. This is because we don't have first class APIs for things like `classNames`.

guides/release/tutorial/autocomplete-component.md
```diff
@@ -79,25 +79,25 @@ The `handleFilterEntry` action will apply the search term filter to the list of
 
 ```javascript {data-filename=app/components/list-filter.js}
+import { action } from '@ember-decorators/object';
+import { classNames } from '@ember-decorators/component';
```
pzuraq: `ember-decorators` is not an official Ember addon, so we aren't going to have it in any examples. Ember itself _does_ support the following decorators:

```js
import { computed, action } from '@ember/object';
import { inject as controller } from '@ember/controller';
import * from '@ember/object/computed';
```

This does mean that we don't have the proper decorators to convert classic components to native class syntax, which is what the core team has agreed on. Everything except class components should always be native classes, and classic components should use classic class syntax.

guides/release/tutorial/autocomplete-component.md
rwjblue: The recent changes to the component blueprints in emberjs/ember.js to support octane were done _before_ we had decided that we didn't want to show native class syntax with `Ember.Component` at all. 

Would someone mind making an issue over there to get it fixed (if you cc @ppcano he will probably be willing to fix it up...)

efx: created that issue: https://github.com/emberjs/ember.js/issues/17701

guides/release/tutorial/autocomplete-component.md
muziejus: Thanks, all. I'll have a look later today and see what I can fix up.

efx:
Thank you @muziejus for taking this on!

guides/release/tutorial/autocomplete-component.md
```diff
@@ -79,25 +79,24 @@ The `handleFilterEntry` action will apply the search term filter to the list of
 
 ```javascript {data-filename=app/components/list-filter.js}
+import { action } from '@ember/object';
 
-  classNames: ['list-filter'],
-  value: '',
+export default class ListFilterComponent extends Component {
+  classNames = 'list-filter';
```
pzuraq: So, we should actually convert this to a GlimmerComponent. You can check out the [edition guide](https://deploy-preview-455--ember-guides.netlify.com/release/upgrading/editions/#toc_glimmer-components) to see what the differences are. In this example, we would update the template like so:


```hbs
<div class="list-filter">
  {{input
    value=this.value
    key-up=this.handleFilterEntry
    class="light"
    placeholder="Filter By City"
  }}
  {{yield this.results}}
</div>
```

This moves the `classNames` to be a part of the template, because Glimmer components don't have a wrapping element. We also remove the `{{action}}` helper from around the action, since it's no longer necessary (`@action` does the same thing). We would then update the class like this:

```js

export default class ListFilterComponent extends Component {
  @tracked value = '';
  @tracked results;

  constructor() {
    super(...arguments);
    this.args.filter('').then((results) => this.results = results);
  }

  @action
  handleFilterEntry() {
    let filterInputValue = this.value;
    let filterAction = this.args.filter;
    filterAction(filterInputValue).then((filterResults) => this.results = filterResults);
  }
}
```

Here we're importing from `@glimmer/component` instead of `@ember/component`, which is the new base class. We also have to use `this.args` to refer to arguments because of this, so we have to update `this.filter` to `this.args.filter` in general. Finally, Glimmer components don't have `this.set`, so we either have to use `Ember.set`, or convert to tracked props. I went for the latter here.

This PR is really looking great, thanks for being understanding, I know there are a lot of changes and they're hard to work through and not all that transparent yet. This is going to be very helpful in getting all this info out there 😄

muziejus: Oh this is great. Thanks. I understand perfectly now, I think. And I'm glad for the clarity about wrapping tags, too, since I'm sick of dealing w/ classNames and tagNames in my component.jses anyway.

muziejus:
OK, between @pzuraq's responses and what @efx did for the component blueprint in emberjs, I think I've got a general sense of what the goals are here:

* Components come from glimmer (gone w/ ad8f9b3 )
* No ember-decorators decorators (those are gone w/ 0d45515 )
* Routes can use native syntax and get that from ember (already in place)
* Ember Data stuff I've more or less reverted, because the model blueprint still uses `export default DS.Model.extend({...` (gone w/ ad8f9b3 )
* Use `@tracked` decorator. (tk)

muziejus:
@pzuraq Pinging you again. I think I've gotten it, and I caught a few overzealous mistakes from elsewhere.

pzuraq:
Awesome job, thanks again for this!

ppcano:
👏@muziejus 

> Ember Data stuff I've more or less reverted, because the model blueprint still uses

Ember Data blueprints will also use Native syntax, check out the open PR https://github.com/emberjs/data/pull/5859

efx:
@ppcano indeed; I think we are tracking that work in #543. I started that PR in #553.

muziejus:
Excellent. Thanks for the support, all. 

@efx I was going to go back through and fix the Ember Data stuff after I saw @ppcano's message, but now I understand I should not and should simply let this ride and help out again in the future when I can.

PR #553, opened 2019-03-01
ember data section class syntax (#543)

efx:
Current progress addressing #543.
See background discussion in the dev-ember-learning channel.

# open questions / related issues
1. https://github.com/ember-learn/guides-source/issues/554
1. ~~standardize decorators and remove usage of ember-decorators~~ 
1. In that discussion, @pzuraq noted we should use `const { Model, attr } = DS;` to then use `@attr` for model attributes. I think this syntax is clearer, but it is slightly inconsistent with example ember-data snippets. (e.g. extending `DS.JSONSerializer`). Should we use destructuring anywhere we import from `DS`?


guides/release/models/defining-models.md
```diff
@@ -51,15 +52,16 @@ properties that combine or transform primitive attributes.
...
+const { Model, attr } = DS;
 
-export default DS.Model.extend({
-  firstName: DS.attr(),
-  lastName: DS.attr(),
+export default class Person extends Model {
+  @attr() firstName;
+  @attr() lastName;
 
   fullName: computed('firstName', 'lastName', function() {
```
efx: This mix of syntaxes has problems. See #554.

guides/release/models/customizing-adapters.md
```diff
@@ -204,22 +208,23 @@ Requests for `user-profile` would now target `/user_profile/1`.
 Some APIs require HTTP headers, e.g. to provide an API key. Arbitrary
 headers can be set as key/value pairs on the `JSONAPIAdapter`'s `headers`
 object and Ember Data will send them along with each ajax request.
-(Note that we set headers in `init()` because default property values
+(Note that we set headers in `constructor()` because default property values
```
efx: Wouldn't using native classes supersede this implementation quirk ? IIRC, this pattern is because of `EmberObject.extend`. Asked in discord dev-ember-learing

guides/release/models/customizing-adapters.md
```diff
@@ -230,17 +235,18 @@ property dependent on the `session` service.
...
 import { inject as service } from '@ember/service';
+const { JSONAPIAdapter } = DS;
 
-
-export default DS.JSONAPIAdapter.extend({
-  session: service('session'),
-  headers: computed('session.authToken', function() {
+export default class ApplicationAdapter extends JSONAPIAdapter {
+  @service session;
+  @computed('session.authToken')
```
pzuraq: I think we can remove `@computed` here and just make this a tracked property

pzuraq: Rather, `session.authToken` should be considered a tracked property, and this can just be a normal getter

guides/release/models/customizing-serializers.md
```diff
@@ -580,15 +595,17 @@ You would define your relationship like this:
 
 ```javascript {data-filename=app/serializers/post.js}
 import DS from 'ember-data';
+const { EmbeddedRecordsMixin, JSONSerializer } = DS;
```
pzuraq: The `EmbeddedRecordsMixin` will be tricky, ideally we can avoid mixins altogether. We'll have to work with the Ember Data team to figure out what the future is here.

guides/release/models/defining-models.md
pzuraq: Yup, I think for now we can just make this a standard ES6 getter, and not a computed property

guides/release/models/customizing-serializers.md
efx: indeed. I chatted briefly with @runspired in #dev-ember-data to get a quick port of the syntax. Same theme there.

runspired:
> In that discussion, @pzuraq noted we should use const { Model, attr } = DS; to then use @attr for model attributes. I think this syntax is clearer, but it is slightly inconsistent with example ember-data snippets. (e.g. extending DS.JSONSerializer). Should we use destructuring anywhere we import from DS?

Yes. See also the blueprints we  just merged that do destructure.

jenweber:
Thank you @efx for your work and sorting out some of these open questions!

efx:
certainly @jenweber ! I'll follow up with another PR that addresses #566 and the tutorial section.

PR #559, opened 2019-03-04
start fixing a few links in state management

efx:
Fixes a couple links in State Management Guide: https://github.com/ember-learn/guides-source/pull/536
I didn't get to all of them and can create an issue if we want to track.
@pzuraq I think  broccoli-static-site-json or whatever converts the guides markdown does not support reference style links.

### DIFF
--- a/guides/release/state-management/index.md
+++ b/guides/release/state-management/index.md
@@ -37,7 +37,7 @@ In Ember, we care mostly about _object_ state in the form of:
 - **Controller state**
 - **Service state**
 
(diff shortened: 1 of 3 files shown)


pzuraq:
Ah, I think I may have just forgotten to put the references in for those. I started using references because they're used in other sections of the guides and they clean things up quite nicely

efx:
@pzuraq nice, I'll just close this then. I do think they are much more readable.

efx:
per @locks, in a recent meeting the team has decided to use inline markdown links.

PR #563, opened 2019-03-05
Lint markdown in CI (and fix linting issues).

rwjblue:


rwjblue:
I wasn't sure if I should add the 3.7 guides to `.remarkignore` (a bunch of the other branches are) or actually fix the content so I went with fixing. Just let me know if we want to go the other direction...

PR #567, opened 2019-03-07
Fix missing curly braces and file path

jenweber:
Itty bitty PR to fix one thing we overlooked.


knownasilya:
You beat me to it, was just about to submit this fix 👍

PR #569, opened 2019-03-07
Changes to prevent default as per

sonic1981:
https://github.com/emberjs/website/issues/3850


acorncom:
@sonic1981 those improvements look great. Tests were failing here, so I've merged master in again, but will merge if things go green here

sonic1981:
Thanks @acorncom

Issue #564, opened 2019-03-05
[OCTANE] rewrite ember-data section recommending use of `volatile`


efx:
We recommend using [volatile](https://github.com/ember-learn/guides-source/blame/octane/guides/release/models/customizing-adapters.md#L248-L251) for custom adapters. This has been [deprecated](https://github.com/emberjs/rfcs/blob/master/text/0370-deprecate-computed-volatile.md#summary).
We should remove this as part of the octane effort. I will try to address it in #543.
Slightly related: #20.

PR #578, opened 2019-03-11
[OCTANE] rewrite text and example to use getters

efx:
Address #564


jenweber:
Thanks @efx for the changes and @rwjblue for reviewing!

Issue #579, opened 2019-03-11
unify import examples to use single quotes


efx:
A quick search of the [guides/release](https://github.com/ember-learn/guides-source/tree/octane/guides/release) directory shows a few import snippets that use double quotes instead of single quotes:
```
$ rg 'import \{.*"'
tutorial/simple-component.md
166:import { action } from "@ember/object";

testing/unit-testing-basics.md
139:import { observer } from "@ember/object";

testing/testing-models.md
36:import { run } from "@ember/runloop";
```
Most examples in the guides use single quotes, so we should update these ones to do so as well. 
This task should also include further search for such quoting inconsistencies.

efx:
In researching this, I think it is exactly the problem noted here: https://github.com/emberjs/guides/issues/613
Automating this as part of the linting checks would be the longer term fix.

PR #589, opened 2019-03-12
add script for extracting code blocks

mansona:
I don't know how useful this would be going forward but if we want to have it available we could think about merging this in some way 🤔 I think for now we should leave this PR open and decide what to do with it in a more general way later 👍 

If anyone wants to use this script in another branch you can just run the following: 

```
git checkout extract-code-script extractCode.js
npm i gfm-code-blocks
```

And then you can run: 

```
node extractCode.js
``` 

and it will export all of the code samples into a `code` directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@
 
 # Deployment credentials
 config/credentials.json
+
+/code
--- a/extractCode.js
+++ b/extractCode.js
@@ -1,0 +1,53 @@
+/* eslint-env node */
+/* eslint no-console:0 */
+
+const walkSync = require('walk-sync');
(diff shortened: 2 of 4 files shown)


efx:
This would help addressing https://github.com/ember-learn/guides-source/issues/579. I would use this to then pipe those files to `eslint` and PR in the interim before automating this entirely.

efx:
I used this to programmatically verify consistent quotes in the JavaScript blocks and think it could work well to automate checking the coding style. It might be feasible to pass the code block to babel to check for syntax errors. 

https://github.com/ember-learn/guides-source/compare/extract-code-script...efx:try-linting-blocks

PR #591, opened 2019-03-13
change survey config to be generlized as infoBanner for Octane

jenweber:
Sister PR to https://github.com/ember-learn/guidemaker-ember-template/pull/12 which should be merged first.

Before merging this, we need to make the guidemaker-ember-template release, reference it in package.json of this app, update the lock file, and then this can be merged. We should also specify the correct link once it is decided, either before merging this PR or we should open another issue.

Screen Shot 2019-03-13 at 12 46 23 AM

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -57,8 +57,8 @@ module.exports = function(environment) {
       },
     ],
     infoBanner: {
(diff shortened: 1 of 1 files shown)


jenweber:
Closed in favor of #599

PR #593, opened 2019-03-13
[OCTANE] add missing import for code example

efx:


jenweber:
Thank you @efx! Merged 🎉

Issue #588, opened 2019-03-12
QUEST - reading and editing the Octane Guides preview


jenweber:
Hello! Thanks in advance for your help! While Octane isn't officially released, and won't be for a while, we are providing a "preview" period where adventurous devs can try it out, and they need guides documentation! A lot of work has already been done, so at this stage, we are looking for people to read the draft of the Octane guides and help us fix errors and fill in the blanks. Can you help us polish things up? Word choice doesn't matter much, so if you see stylistic improvements, please open an issue instead of including them in a PR. However spelling, typo fixes, and and factual accuracy are very important and we would love your help with those.

All work should be done by branching from `octane` in this repository, and all PRs should be pointed at `octane`, not `master`. See https://github.com/ember-learn/guides-source/issues/583 for ideas of what to look for, and links to a script that will aggregate code blocks, if that is useful to you. Thank you!!!

If you can help, please comment on this issue to say which page within the guides you can review. In your comment, post a link. When you have a PR, in the description please link to this issue.

## List of pages that may need volunteers
(read comments below to see if they are taken)

- [x] "Getting Started", url: 'getting-started'
    - [x] "Quick Start", url: "quick-start", #590
    - [x] "Installing Ember", url: "index", checked during https://github.com/ember-learn/guides-source/pull/590
    - [x] "Core Concepts", url: "core-concepts" - needs an updated image
- [x] "Anatomy of an App", url: "anatomy-of-an-app"
    - [x] "Overview", url: "index"
- [x] "Templating", url: 'templates'
    - [x] "Templating Basics", url: "handlebars-basics" https://github.com/ember-learn/guides-source/pull/598
    - [x] "Conditionals", url: "conditionals" https://github.com/ember-learn/guides-source/pull/598
    - [x] "Displaying a List of Items", url: "displaying-a-list-of-items" https://github.com/ember-learn/guides-source/pull/598
    - [x] "Displaying the Keys in an Object", url: "displaying-the-keys-in-an-object" https://github.com/ember-learn/guides-source/pull/598
    - [x] "Binding Element Attributes", url: "binding-element-attributes" https://github.com/ember-learn/guides-source/pull/598
    - [x] "Actions", url: "actions" https://github.com/ember-learn/guides-source/pull/598
    - [x] "Built-in Helpers", url: "built-in-helpers" https://github.com/ember-learn/guides-source/pull/598
    - [x] "Writing Helpers", url: "writing-helpers" #598 and #597 
[...]

Issue #300, opened 2018-12-03
tutorial/service.md wording discussion


efx:
Sub issue of #245.
>  "An Ember Service is an object that can be made available for different parts of an application." does not differentiate from components. Should indicate that it's special because there is only one instance of it, and it can be used almost anywhere to store shared information or methods.

We define services here: https://github.com/ember-learn/guides-source/blob/master/guides/v3.5.0/applications/services.md.

The problematic, half-definition is given here: https://github.com/ember-learn/guides-source/blob/master/guides/v3.5.0/tutorial/service.md. 

# proposed solution / change
I think we should remove the line attempting to define a service in `tutorial/service.md` and simply link the prior line end `Ember services.` to applications/services.md. By linking we give the user a clear path to add services to their mental model and avoid the liability of competing definitions.

locks:
@efx can you send a PR pliz?

PR #594, opened 2019-03-13
Reading/Editing/Octaning “Testing” section

muziejus:
Working on #588

### DIFF
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -292,7 +292,6 @@ export default class LocationIndicatorComponent extends Component {
   @service location;
 
   // when the coordinates change, call the location service to get the current city and country
-  @computed('locationService.currentLocation')
   get city() {
     return this.locationService.getCurrentCity();
   }
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -227,7 +227,7 @@ export default class CommentFormComponent extends Component {
   submitComment() {
     this.args.submitComment({ comment: this.comment });;
   }
-});
+}
 ```
 
 ```handlebars {data-filename="app/templates/components/comment-form.hbs"}
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -14,7 +14,7 @@ export default class PrettyColorComponent extends Component {
   get style() {
     return `color: ${this.args.name}`;
   }
-};
+}
 ```
 
 ```handlebars {data-filename="app/templates/components/pretty-color.hbs"}
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -293,7 +293,7 @@ export default class LocationIndicatorComponent extends Component {
 
   // when the coordinates change, call the location service to get the current city and country
   get city() {
-    return this.locationService.getCurrentCity();
+    return this.location.getCurrentCity();
   }
 
   get country() {
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -226,7 +226,7 @@ export default class CommentFormComponent extends Component {
 
   @action
   submitComment() {
-    this.submitComment({ comment: this.comment });;
+    this.args.submitComment({ comment: this.comment });;
   };
 });
 ```
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -476,7 +476,7 @@ export default class DelayedTypeaheadComponent extends Component {
   handleTyping() {
     //the fetchResults function is passed into the component from its parent
     debounce(this, this.fetchResults, this.searchValue, 250);
-  };
+  }
 };
 ```
 
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -227,7 +227,7 @@ export default class CommentFormComponent extends Component {
   @action
   submitComment() {
     this.args.submitComment({ comment: this.comment });;
-  };
+  }
 });
 ```
 
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -8,20 +8,19 @@ The `style` attribute of the component is bound to its `style` property.
 > component pretty-color`.
 
 ```javascript {data-filename="app/components/pretty-color.js"}
-import Component from '@ember/component';
-import { computed } from '@ember/object';
+import Component from '@glimmer/component';
 
-export default Component.extend({
-  attributeBindings: ['style'],
-
-  style: computed('name', function() {
-    return `color: ${this.name}`;
-  })
-});
+export default class PrettyColorComponent extends Component {
+  get style() {
+    return `color: ${this.args.name}`;
+  }
+};
 ```
 
 ```handlebars {data-filename="app/templates/components/pretty-color.hbs"}
-Pretty Color: {{this.name}}
+<div style={{this.style}}>
+  Pretty Color: {{@name}}
+</div>
 ```
 
 The `module` from QUnit will scope your tests into groups of tests which can be configured and run independently.
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -172,7 +172,7 @@ export default class MagicTitleComponent extends Component {
 ```handlebars {data-filename="app/templates/components/magic-title.hbs"}
 <h2>{{this.title}}</h2>
 
-<button class="title-button" {{action "updateTitle"}}>
+<button class="title-button" onclick={{action this.updateTitle}}>
   Update Title
 </button>
 ```
@@ -232,7 +232,7 @@ export default class CommentFormComponent extends Component {
 ```
 
 ```handlebars {data-filename="app/templates/components/comment-form.hbs"}
-<form {{action "submitComment" on="submit"}}>
+<form onsubmit={{action this.submitComment}}>
   <label>Comment:</label>
   {{textarea value=this.comment}}
 
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -301,7 +301,7 @@ export default class LocationIndicatorComponent extends Component {
   @computed('locationService.currentLocation')
   get country() {
     return this.locationService.getCurrentCountry();
-  });
+  }
 };
 ```
 
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -232,7 +232,7 @@ export default class CommentFormComponent extends Component {
 ```
 
 ```handlebars {data-filename="app/templates/components/comment-form.hbs"}
-<form onsubmit={{action this.submitComment}}>
+<form {{action this.submitComment on="submit"}}>
   <label>Comment:</label>
   {{textarea value=this.comment}}
 
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -297,7 +297,7 @@ export default class LocationIndicatorComponent extends Component {
   }
 
   get country() {
-    return this.locationService.getCurrentCountry();
+    return this.location.getCurrentCountry();
   }
 };
 ```
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -165,7 +165,7 @@ export default class MagicTitleComponent extends Component {
   updateTitle() {
     this.set('title', 'This is Magic');
   }
-};
+}
 ```
 
 ```handlebars {data-filename="app/templates/components/magic-title.hbs"}
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -156,17 +156,17 @@ Imagine you have the following component that changes its title when a button is
 > component magic-title`.
 
 ```javascript {data-filename="app/components/magic-title.js"}
-import Component from '@ember/component';
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
 
-export default Component.extend({
-  title: 'Hello World',
+export default class MagicTitleComponent extends Component {
+  title = 'Hello World';
 
-  actions: {
-    updateTitle() {
-      this.set('title', 'This is Magic');
-    }
-  }
-});
+  @action
+  updateTitle() {
+    this.set('title', 'This is Magic');
+  };
+};
 ```
 
 ```handlebars {data-filename="app/templates/components/magic-title.hbs"}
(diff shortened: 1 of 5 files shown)


muziejus:
I don't know how to properly separate two PRs at once, so the first five commits here come from my other current PR. Nevertheless, I'm using this space to take some notes on the testing pages:

1. In the "Introduction," I think that `this.owner` is described too abruptly.

> In the example below, we import the setupTest function from ember-qunit and call it with the hooks object to setup the test context with access to the this.owner property. This provides us direct container access to interact with Ember's Dependency Injection system.

The link to the Dependency Injection guide is handy, but it's not quite clear why, if `this.owner` is a property of the test, that it "owns" an `ApplicationInstance`. It may be better to use text like:

> In the example below, we import the setupTest function from ember-qunit and call it with the hooks object to setup the test context. This uses Ember's Dependency Injection system to create an ApplicationInstance that serves as the "owner" for objects that are instantiated during testing. We can access the ApplicationInstance with the `this.owner` property to "look up" (i.e. instantiate and / or retrieve) objects.

muziejus:
Next, the testing sections present a few difficulties for octanization:

2. In "Testing Basics," there is a whole section to testing observers. As @pzuraq notes, "[observers and event listeners are not a recommended pattern anymore](https://www.pzuraq.com/coming-soon-in-ember-octane-part-1-native-classes/)," so maybe this section should be cut entirely or refactored to use `@tracked`.

3. The first test in "Testing Components" relies on `attributeBindings`. These are, if I understand correctly, not to be used in Octane, as something like a component's `style` attribute can be set programmatically with something like `<div style={{this.style}}>…</div>`. I would propose a simpler test that uses Glimmer component basics.

4. I don't know what the Octane orthodoxy is vis-à-vis controllers, but given that the Controllers section of the guides is not on the sidebar, that it has not been updated to use native class syntax, and that the controllers section of the "Routing" guide is gone, I think perhaps it's time to retire the controllers testing section, as well?

guides/release/testing/testing-components.md
```diff
@@ -156,23 +156,23 @@ Imagine you have the following component that changes its title when a button is
...
+    this.set('title', 'This is Magic');
+  };
+};
 ```
 
 ```handlebars {data-filename="app/templates/components/magic-title.hbs"}
 <h2>{{this.title}}</h2>
 
-<button class="title-button" {{action "updateTitle"}}>
+<button class="title-button" onclick={{action this.updateTitle}}>
```
rwjblue: We are avoiding `onclick=` usage (see @mixonic's awesome explanation over at https://github.com/ember-learn/guides-source/pull/598#discussion_r265390517)

```suggestion
<button class="title-button" {{action this.updateTitle}}>
```

guides/release/testing/testing-components.md
```diff
@@ -218,21 +218,21 @@ passing along the form's data:
...
+  comment = '';
 
-  actions: {
-    submitComment() {
-      this.submitComment({ comment: this.comment });;
-    }
-  }
+  @action
+  submitComment() {
+    this.submitComment({ comment: this.comment });;
```
rwjblue: On glimmer components, the passed in arguments are not on `this`, they are in `this.args`.

```suggestion
    this.args.submitComment({ comment: this.comment });;
```

guides/release/testing/testing-components.md
```diff
@@ -218,21 +218,21 @@ passing along the form's data:
...
+  };
```
rwjblue: Inside a class body there is no semicolon or comma separating methods.

```suggestion
  }
```

guides/release/testing/testing-components.md
```diff
@@ -218,21 +218,21 @@ passing along the form's data:
...
 });
 ```
 
 ```handlebars {data-filename="app/templates/components/comment-form.hbs"}
-<form {{action "submitComment" on="submit"}}>
+<form onsubmit={{action this.submitComment}}>
```
rwjblue:
```suggestion
<form {{action this.submitComment on="submit"}}>
```

guides/release/testing/testing-components.md
```diff
@@ -285,22 +285,24 @@ Imagine you have the following component that uses a location service to display
...
+export default class LocationIndicatorComponent extends Component {
+  @service locationService;
 
   // when the coordinates change, call the location service to get the current city and country
-  city: computed('locationService.currentLocation', function () {
+  @computed('locationService.currentLocation')
+  get city() {
     return this.locationService.getCurrentCity();
-  }),
+  });
```
rwjblue:
```suggestion
  }
```

guides/release/testing/testing-components.md
```diff
@@ -285,22 +285,24 @@ Imagine you have the following component that uses a location service to display
...
 
-  country: computed('locationService.currentLocation', function () {
+  @computed('locationService.currentLocation')
+  get country() {
     return this.locationService.getCurrentCountry();
-  })
-});
+  });
```
rwjblue:
```suggestion
  }
```

guides/release/testing/testing-components.md
```diff
@@ -465,17 +467,17 @@ Imagine you have a typeahead component that uses [`Ember.run.debounce`](https://
...
-      debounce(this, this.fetchResults, this.searchValue, 250);
-    }
-  }
-});
+export default class DelayedTypeaheadComponent extends Component {
+  @action 
+  handleTyping() {
+    //the fetchResults function is passed into the component from its parent
+    debounce(this, this.fetchResults, this.searchValue, 250);
+  };
```
rwjblue:
```suggestion
  }
```

guides/release/testing/unit-testing-basics.md
```diff
@@ -17,13 +17,14 @@ based on a `foo` property.
...
-  foo: 'bar',
+export default class SomeThingService extends Service {
+  foo = 'bar';
 
-  computedFoo: computed('foo', function() {
-    return `computed ${this.foo}`;
+  @computed('foo')
+  get computedFoo() {
+    return `${this.foo}`;
   })
```
rwjblue:
```suggestion
  }
```

guides/release/testing/unit-testing-basics.md
```diff
@@ -138,14 +139,15 @@ Suppose we have an object that has a property and a method observing that proper
 import Service from '@ember/service';
 import { observer } from "@ember/object";
 
-export default Service.extend({
-  foo: 'bar',
-  other: 'no',
+export default class SomeThingService extends Service {
```
rwjblue: Lets revert this change for now, I don't think we can actually talk about observers in the new world (because there is no decorator form of `observes` in core, and ultimately folks should _probably_ move away from them).

rwjblue:
> In "Testing Basics," there is a whole section to testing observers. As @pzuraq notes, "observers and event listeners are not a recommended pattern anymore," so maybe this section should be cut entirely or refactored to use @tracked.

Ya, agreed. Lets remove the observer test example section.

rwjblue:
> The first test in "Testing Components" relies on attributeBindings. These are, if I understand correctly, not to be used in Octane, as something like a component's style attribute can be set programmatically with something like <div style={{this.style}}>…</div>. I would propose a simpler test that uses Glimmer component basics.

I actually think the example there is still valid. In a `@glimmer/component` world `style` attribute does not have any special merging behaviors. If you wanted to have a component that took `color` as an argument, it would still _mostly_ need the same code.

I'd write the example component as:

```js

export default class PrettyColor extends Component {
  get style() {
    return `color: ${this.args.name}`;
  }
});
```

```hbs
<div style={{this.style}}>
  Pretty Color: {{@name}}
</div>
```

rwjblue:
> but given that the Controllers section of the guides is not on the sidebar, that it has not been updated to use native class syntax

I think it still needs to be updated, octane does not remove the need for controllers at all. However, since the controllers section in the sidebar is removed, perhaps we should just remove the controller testing section from the sidebar for now as well. That way we can bring both back together, when they are ready.

guides/release/testing/unit-testing-basics.md
```diff
@@ -65,13 +66,13 @@ the `foo` property).
...
-export default Service.extend({
-  foo: 'bar',
+export default class SomeThingService extends Service {
+  foo = 'bar';
 
   testMethod() {
     this.set('foo', 'baz');
   }
-});
+};
```
efx:
```suggestion
}
```

guides/release/services/services.md
```diff
@@ -25,8 +25,8 @@ Services must extend the [`Service`](https://www.emberjs.com/api/ember/release/m
 ```javascript {data-filename=app/services/shopping-cart.js}
 import Service from '@ember/service';
 
-export default Service.extend({
-});
+export default class ShoppingCartService extends Service {
+};
```
efx:
```suggestion
}
```

muziejus:
@rwjblue and @efx, any thoughts on adding a bit more explanation to `this.owner` in the introduction section? The controllers test also describes it better than the intro, imo:

> Next we use the owner API to gain access to the controller we'd like to test.
Using the `this.owner.lookup` method we get the instance of the `PostsController` and can check the action in our test.
The `this.owner.lookup` helper returns objects generated by the framework in your applications
and is also exposed in tests for your usage. Here it will return a singleton instance of the `PostsController`.

rwjblue:
OK, I think we need a rebase now (to drop out the commits that were also in #588). Does that make sense? Also, happy to help with that if you'd like...

muziejus:
@rwjblue I think I got it.

muziejus:
Thanks, all! 🎸

efx:
@muziejus I think discussing how we introduce the owner API is a good idea. I had a similar problem with how the current guides talked about services: https://github.com/ember-learn/guides-source/issues/300#issue-386970151
I prefer linking to the better definition and even the API documentation itself. Having 2 definitions of an API floating around in the guides could lead to confusion as you rightly note.

muziejus:
@efx Good point about not duplicating definitions, as I sort of do that in my proposal, copy and pasting from the dependency injection page's discussion of `ApplicationInstance`.

Could the glossary help here? 

Similarly, something like:

```markdown
In the example below, we import the `setupTest` function from `ember-qunit` and call it with the `hooks` object to setup the test context. This provides us with the `this.owner` property. As the [API docs note]:

> When Ember's [Dependency Injection] system creates an [ApplicationInstance], it exposes a `this.owner` property that serves as the "owner" for objects that are instantiated during testing. We can access the `ApplicationInstance` with the `this.owner.lookup` to "look up" (i.e. instantiate and / or retrieve) objects. Similarly, we can register...
```

Might be good, since it marks that we're relying on a different document for the quoted material. That may be friendlier than just a link.

guides/release/testing/unit-testing-basics.md
```diff
@@ -65,13 +66,13 @@ the `foo` property).
 ```javascript {data-filename=app/services/some-thing.js}
 import Service from '@ember/service';
```
rwjblue:
```suggestion
import Service from '@ember/service';
import { tracked } from '@glimmer/tracking';
```

guides/release/testing/unit-testing-basics.md
```diff
@@ -65,13 +66,13 @@ the `foo` property).
...
 
-export default Service.extend({
-  foo: 'bar',
+export default class SomeThingService extends Service {
+  foo = 'bar';
```
rwjblue: This should be a tracked property:

```suggestion
  @tracked foo = 'bar';
```

guides/release/testing/unit-testing-basics.md
```diff
@@ -65,13 +66,13 @@ the `foo` property).
...
 
   testMethod() {
     this.set('foo', 'baz');
```
rwjblue:
```suggestion
    this.foo = 'baz';
```

guides/release/testing/unit-testing-basics.md
```diff
@@ -102,14 +103,14 @@ that returns a value based on some internal state.
 ```javascript {data-filename=app/services/some-thing.js}
 import Service from '@ember/service';
 
-export default Service.extend({
-  count: 0,
+export default class SomeThingService extends Service {
+  count = 0;
```
rwjblue:
```suggestion
  @tracked count = 0;
```

guides/release/testing/unit-testing-basics.md
```diff
@@ -102,14 +103,14 @@ that returns a value based on some internal state.
 ```javascript {data-filename=app/services/some-thing.js}
 import Service from '@ember/service';
```
rwjblue:
```suggestion
import Service from '@ember/service';
import { tracked } from '@glimmer/tracking';
```

guides/release/testing/unit-testing-basics.md
```diff
@@ -102,14 +103,14 @@ that returns a value based on some internal state.
...
 
-export default Service.extend({
-  count: 0,
+export default class SomeThingService extends Service {
+  count = 0;
 
   calc() {
     this.incrementProperty('count');
```
rwjblue:
```suggestion
    this.count += 1;
```

efx:
@muziejus big 👍 from me to your revision. Honestly don't think we even need to quote the API documentation as it is linked. I also want to say we should continue any wording tweaks outside of this PR because the *grand work* you're doing here should be be merged once the technical review notes are done.

guides/release/testing/testing-components.md
```diff
@@ -8,20 +8,19 @@ The `style` attribute of the component is bound to its `style` property.
...
 
-  style: computed('name', function() {
-    return `color: ${this.name}`;
-  })
-});
+export default class PrettyColorComponent extends Component {
+  get style() {
+    return `color: ${this.args.name}`;
+  }
+};
```
efx:
```suggestion
}
```

guides/release/testing/testing-components.md
```diff
@@ -156,23 +155,23 @@ Imagine you have the following component that changes its title when a button is
...
-    updateTitle() {
-      this.set('title', 'This is Magic');
-    }
-  }
-});
+  @action
+  updateTitle() {
+    this.set('title', 'This is Magic');
+  };
+};
```
efx:
```suggestion
}
```

guides/release/testing/testing-components.md
```diff
@@ -156,23 +155,23 @@ Imagine you have the following component that changes its title when a button is
...
-  actions: {
-    updateTitle() {
-      this.set('title', 'This is Magic');
-    }
-  }
-});
+  @action
+  updateTitle() {
+    this.set('title', 'This is Magic');
+  };
```
efx:
```suggestion
  }
```
I think method definitions and the class exports do not require a semicolon so we should remove them. You can see the bulk of the examples in https://github.com/ember-learn/guides-source/blob/octane/guides/release/upgrading/editions.md#native-classes too for reference.

guides/release/testing/testing-components.md
```diff
@@ -218,21 +217,21 @@ passing along the form's data:
...
 
-  actions: {
-    submitComment() {
-      this.submitComment({ comment: this.comment });;
-    }
+  @action
+  submitComment() {
+    this.args.submitComment({ comment: this.comment });;
   }
 });
```
efx:
```suggestion
}
```

guides/release/testing/testing-components.md
muziejus: Right. I have been remiss in making sure I do this everywhere.

muziejus:
@rwjblue Per your rec, I deleted (4e8badb) the section on testing observers, but I do think it'd be useful to instead replace it with a section on testing tracked properties, so that may be a way to advance the points you bring up.

muziejus:
@efx Thanks. You're right… I think once the code is somewhat stable I can have a stab at the language in separate PRs to work on better comprehension. As a relative noob, I think I have a couple insights into sniffing out where things get too confusing too quickly!

PR #598, opened 2019-03-13
Review and Octanify the template guides

mixonic:
Re: https://github.com/ember-learn/guides-source/issues/588

/cc @pzuraq @jenweber

### DIFF
--- a/guides/release/templates/actions.md
+++ b/guides/release/templates/actions.md
@@ -8,64 +8,70 @@ the context of the template:
 
 ```javascript {data-filename=app/components/post/component.js}
 import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 
 export default class Post extends Component {
+  @tracked isShowingBody;
+
   @action
   toggleBody() {
-    this.toggleProperty('isShowingBody');
+    this.isShowingBody = !this.isShowingBody;
   }
 }
 ```
 
-You can then add this action directly to an [_event handler
-property_](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers)
-on an element, like `onclick` or `onmouseenter`:
+You can then add this action to an element using the
+[`{{action}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/action?anchor=action)
+helper:
 
 ```handlebars {data-filename=app/components/post/template.hbs}
 <h3>
-  <button onclick={{this.toggleBody}}>{{this.title}}</button>
+  <button {{action this.toggleBody}}>{{@title}}</button>
 </h3>
 
 {{#if this.isShowingBody}}
-  <p>{{this.body}}</p>
+  <p>{{@body}}</p>
 {{/if}}
 ```
 
-This assigns the action to the standard browser event handler for that function.
-It'll receive the event as its first parameter, and you can handle it like any
-standard JavaScript event:
-
-```javascript {data-filename=app/components/post/component.js}
-import Component from '@glimmer/component';
-import { action } from '@ember/object';
-
-export default class Post extends Component {
-  @action
-  toggleBody(event) {
-    event.preventDefault();
-    event.stopPropagation();
-    this.toggleProperty('isShowingBody');
-  }
-}
-```
-
+The `{{action}}` helper calls your action function when the element is
+clicked.
 You will learn about more advanced usages in the Component's [Actions and
 Events](../../components/actions-and-events/) guide, but you should familiarize
-yourself with the following basics first.
+yourself with the basics on this page first.
+
+Templates rendered for your application's routes are backed by controllers, so
+you may also see actions defined on a controller using the same `@action`
+decorator.
+
+<div class="cta">
+  <div class="cta-note">
+    <div class="cta-note-body">
+      <div class="cta-note-heading">Zoey says...</div>
+      <div class="cta-note-message">
+If this doesn't seem familiar you might be looking for documentation
+about Ember components and controller that use the older <code>.extend({</code> syntax.
+In those files you define actions in an object on the class then reference
+the action name with a string in the template.
+For more examples of that syntax see <a href="https://guides.emberjs.com/v3.6.0/templates/actions/">previous version of this Guide entry</a>.
+      </div>
+    </div>
+    <img src="/images/mascots/zoey.png" role="presentation" alt="Ember Mascot">
+  </div>
+</div>
 
 ## Action Parameters
 
-You can optionally pass arguments to the action with the
-[`{{action}}`](https://www.emberjs.com/api/ember/release/classes/Ember.Templates.helpers/methods/action?anchor=action)
+You can optionally pass arguments to the action
 helper. For example, if the `post` argument was passed:
 
 ```handlebars {data-filename=app/components/post/template.hbs}
 <p>
-  <button onclick={{action this.select this.post}}>
+  <button {{action this.select @post}}>
     ✓
   </button>
-  {{this.post.title}}
+  {{this.selectedPost.title}}
 </p>
 ```
 
@@ -87,8 +93,18 @@ export default class Post extends Component {
 }
 ```
 
-If you pass arguments like this, the event will be the _last_ argument that is
-passed to the handler:
+## Attaching Actions to Other Events
+
+Actions don't need to be triggered on click, in fact they can be attached
+to any event Ember is already listening to. For example this form will
+call the `updateText` action when submitted, but prevent the default
+form submission logic in the browser from running:
+
+```handlebars {data-filename=app/components/post/template.hbs}
+<form {{action this.createPost on="submit" preventDefault=true}}>
+  Post title: <Input @value={{this.title}} />
+</form>
+```
 
 ```javascript {data-filename=app/components/post/component.js}
 import Component from '@glimmer/component';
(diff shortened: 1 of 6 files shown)


guides/release/templates/actions.md
```diff
@@ -8,64 +8,70 @@ the context of the template:
...
+    this.isShowingBody = !this.isShowingBody;
   }
 }
 ```
 
-You can then add this action directly to an [_event handler
-property_](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers)
-on an element, like `onclick` or `onmouseenter`:
+You can then add this action to an element using the
+[`{{action}}`](https://www.emberjs.com/api/ember/release/classes/Ember.Templates.helpers/methods/action?anchor=action)
```
acorncom: Let's make this the newer server we're running
```suggestion
[`{{action}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/action?anchor=action)
```

guides/release/templates/binding-element-attributes.md
```diff
@@ -38,39 +38,41 @@ If `isAdministrator` is `false`, Handlebars will produce the following:
 <input type="checkbox">
 ```
 
-### Adding Other Attributes (Including Data Attributes)
+### Adding Attributes to Component Invocations
 
-By default, components only accept a limited number of HTML attributes.
-This means that some uncommon but perfectly valid attributes, such as `lang` or
-custom `data-*` attributes must be specifically enabled. For example, this template:
+Components in Ember can also accept attributes to be places on an element
```
acorncom:
```suggestion
Components in Ember can also accept attributes to be placed on an element
```

guides/release/templates/binding-element-attributes.md
```diff
@@ -38,39 +38,41 @@ If `isAdministrator` is `false`, Handlebars will produce the following:
...
+  Fotos
+</LinkTo>
 ```
 
-Will render the following HTML:
+Which will render render the following HTML:
 
 ```html
-<a id="ember239" class="ember-view" href="#/photos">Fotos</a>
+<a href="#/photos" data-toggle="dropdown" lang="es" id="ember239" class="ember-view">
```
acorncom: Seems worth simplifying down to what we'd like produce here?
```suggestion
<a href="/photos" data-toggle="dropdown" lang="es" id="ember239" class="ember-view">
```

guides/release/templates/binding-element-attributes.md
```diff
@@ -38,39 +38,41 @@ If `isAdministrator` is `false`, Handlebars will produce the following:
...
-});
-```
-
-Now the same template above renders the following HTML:
-
-```html
-<a id="ember239" class="ember-view" href="#/photos" data-toggle="dropdown" lang="es">Fotos</a>
-```
+Components can control which of their elements receive attributes.
+[Read more about how to customize a component's element](/components/customizing-a-components-element/).
```
acorncom: We'll want to use relative paths here so we don't hit snags when this file becomes part of `3.9`/`3.10` guides ...
```suggestion
[Read more about how to customize a component's element](../../components/customizing-a-components-element/).
```

guides/release/templates/built-in-helpers.md
```diff
@@ -168,10 +168,10 @@ when the `debugger;` breakpoint is hit, you can attempt to retrieve this value:
 {{/each}}
 ```
 
-You'll be able to get values from the current item:
+You'll be able to get value of the current item:
```
acorncom:
```suggestion
You'll be able to get the value of the current item:
```

guides/release/templates/writing-helpers.md
```diff
@@ -238,21 +238,23 @@ And two named arguments:
...
 and a hash), they perform an operation on those inputs, and return a
 single output. They have no side-effects and don't save any information
 that is used on subsequent runs of the function.
 
 In some situations, however, you may need to write a helper that interacts with
-the rest of your application. You can create class-based helpers that have
-access to services in your application, and can optionally save state as well,
-although this is usually unnecessary and error-prone.
+other parts of your application. You can create class-based helpers that have
+access to services. Because class based helper have an object they could
```
acorncom:
```suggestion
access to services. Because class based helpers have an object they could
```

guides/release/templates/writing-helpers.md
```diff
@@ -290,15 +292,18 @@ export default class IsAuthenticated extends Helper {
   compute() {
     let authentication = this.authentication;
 
-    if (authentication.get('isAuthenticated')) {
-      return 'Welcome back, ' + authentication.get('username');
+    if (authentication.isAuthenticated) {
+      return 'Welcome back, ' + authentication.username;
```
acorncom: Let's switch to a template literal here?
```suggestion
      return `Welcome back, ${authentication.username}`;
```

guides/release/templates/writing-helpers.md
```diff
@@ -290,15 +292,18 @@ export default class IsAuthenticated extends Helper {
...
     } else {
       return 'Not logged in';
     }
   }
 }
 ```
 
+Becuase this helper is not passed `isAuthenticated` or `username` as an
```
acorncom:
```suggestion
Because this helper is not passed `isAuthenticated` or `username` as an
```

guides/release/templates/actions.md
```diff
@@ -87,71 +93,37 @@ export default class Post extends Component {
...
-    event.preventDefault();
-    event.stopPropagation();
-    this.selectedPost = post;
+  createPost() {
+    // Do something with `this.title`
   }
 }
 ```
 
-## Modifying the action's first parameter
```
acorncom: Did you remove this section because it ended up seeming confusing?

guides/release/templates/binding-element-attributes.md
mixonic: Cool, confirmed the non-hash version is actually what we produce out of the box.

guides/release/templates/actions.md
mixonic: @acorncom excellent question! While editing these guide entries I was surprised to see `onclick={{action this.foo}}`-style actions being used widely. I brought it up vaguely in `#st-octane` and once my thoughts were more organized in `#core-framework` on Discord.

In Glimmer components there is no actions hash. Even with the `@action` decorator the following doesn't work:

```js

export default class extends Component {
  @action
  doThing() {
    // Do it!
  }
}
```

```hbs
<button {{action 'doThing'}}>Do it!</button>
```

You cannot use a string to reference the action on the actions hash because there is no actions hash. And on glimmer component we see no need to create one. Note that a similar issue exists for native-class-Ember-component's which use the `@action` decorator and for native-class-syntax-controllers which do the same. They don't have an actions hash to reference with the string.

Instead, we intend to suggest developers reach for the `@action`-decorated property/method itself. For example:

```js

export default class extends Component {
  @action
  doThing() {
    // Do it!
  }
}
```

```hbs
<button {{action this.doThing}}>Do it!</button>
```

### But wait, I can just use `onclick`!

Clever developers will note that the above is using the action helper where it isn't strictly going to be needed. They might reach for the following:

```hbs
{{!
  I know that @action binds the method to the component, so I'll just use
  use the bare method/action.
}}
<button onclick={{this.doThing}}>Do it!</button>
```

We would like to make this fail a linter check. Teaching that developers write an action this way has several downsides:

* `onclick=` teaches that you can always attach an event listener with `onfoo=` syntax. This is not actually true. Some DOM elements trigger events (such as `focusin`) without calling the related property. Additionally web components don't call these properties. Instead you must attach an event listener. If developers learn this API first they will start off on the wrong foot.
* `onclick=` works because Ember today has a "prop first, fallback to attribute" rule for how to set attributes on an element. Props can have subtle differences between attributes, even for the same values. More significantly properties have no way to be serialized into HTML, making them very difficult and in some cases impossible to pass to the browser when doing server-side rendering.
* `onclick=` provides the developer no opportunity to pass arguments to the event listener. For example `passive: true` or `capture: true`. We should provide developers an API where opting into options for their listener is straightforward.
* Finally we already have an API in `<button {{action this.doThing}}>` that does the work of `onclick=`. There are definitely ways to improve on the action modifier and helper, but `onclick=` doesn't offer much.

### But it does offer me one thing: The event!

Correct! Using `onclick=` done one thing that the modifier `{{action` does not: It gives you access to the event. For example this calls the action `alert` with the value of the input every time it changes:

```hbs
<input onchange={{action this.alert value="target.value"}}>
```

And this exact use of the API is what this subsection of the guide page is talking about. Although this behavior is not deprecated (and may be useful) we will not be suggesting `onfoo=` elsewhere for all the reasons outlined above, and so I do not think we should introduce this uncommon usage in the guides. It remains documented [in the API docs](https://api.emberjs.com/ember/3.8/classes/Ember.Templates.helpers/methods/action?anchor=action)

### But I really don't like `<button {{action this.foo}}>`

Right. I agree it isn't all we want. For example you still can't add a listener for a custom event emitted by a web component to the action modifier. You can pass some useful arguments but not `passive: true`. You cannot get access to the event object.

Additionally some of the more advanced developers might know that the action modifier and helper have a special property: They bind an action passed to them to the `context` (or the value of the `target=` option if you pass one). They might think: "Ah ha you don't need the `@action` decorator!". But we do. We need the `@action` decorator to bind the actions onto the component so that in the future we can move from the action modifier (and maybe helper) to something better. We will provide a check that fails if you try to use `{{action this.someProp}}` and `someProp` has not had the `@action` decorator applied.

There is strong support among framework core that something akin to https://github.com/rwjblue/ember-on-modifier could become a built-in for Ember. The `on` modifier would solve most of the pain points around `{{action`. We cannot be sure it would ship in Octane at this late date, but we should consider soon how we can migrate existing action usage over to that new and better tool.

### Final legacy note

Ember components and controllers with classic `.extend({` should continue to use the action modifier or helper with the string argument. That class definition syntax does not have the ability to use the action decorator, and we see no need to churn code using that API with anything new.

guides/release/templates/writing-helpers.md
```diff
@@ -238,21 +238,23 @@ And two named arguments:
 
 ### Class-based Helpers
 
-By default, helpers are _stateless_. They are passed inputs (parameters
+Generated helpers are simply functions. They are passed inputs (parameters
```
rwjblue:
```suggestion
Generated helpers are simply functions. They are passed arguments (positional
```

guides/release/templates/writing-helpers.md
```diff
@@ -238,21 +238,23 @@ And two named arguments:
...
 and a hash), they perform an operation on those inputs, and return a
```
rwjblue:
```suggestion
and named), they perform an operation on those inputs, and return a
```

Issue #554, opened 2019-03-01
what should we teach in introducing Ember Data Models? computeds vs. decorators


efx:
We teach computed properties [when introducing people to Ember data models](https://guides.emberjs.com/release/models/defining-models/#toc_defining-attributes).
In the Octane Edition we introduce [decorators](https://github.com/ember-learn/guides-source/blob/octane/guides/release/upgrading/editions.md#decorators), and later [tracked properties](https://github.com/ember-learn/guides-source/blob/octane/guides/release/upgrading/editions.md#tracked-properties).
This raises a broader question when refactoring the [canonical user model `fullName` example ](https://github.com/ember-learn/guides-source/blob/master/guides/release/models/defining-models.md#defining-attributes). I think this question needs answered before question 2.

1. What should we teach in this example? Computed's, tracked properties?
1. What should this introductory snippet be refactored to?

MelSumner:
@igorT @dgeb can you weigh in here?

pzuraq:
I believe we should use Tracked Properties here, and in general. Since tracked props interop with computeds, attrs and relationships should Just Work ™️:

```js
import DS from 'ember-data';
import { computed } from '@ember/object';

export default DS.Model.extend({
  firstName: DS.attr(),
  lastName: DS.attr(),

  get fullName() {
    return `${this.firstName} ${this.lastName}`;
  }
});
```

There are two things we need to resolve here:

1. Can we teach extending from `DS.Model` using native class syntax? I believe the answer is yes, but I'd like the Ember Data team to weigh in here. The biggest concerns would be around `init`/`constructor`, and how those two would interop, since we are trying in general to teach native classes as `constructor` only (this is part of why classic components are _not_ being taught with native classes).

2. What is the final syntax for defining a native getter on a classic class? We debated this the week before last, I need to get the above syntax benchmarked and if it looks alright, we could move ahead with it. Otherwise we need to introduce `descriptor`.

runspired:
There are a couple of direction we can go from here, I'm going to lay them out and then argue that a particular direction seems best to me:

* minimal upgrade of syntax
* leave alone until custom Record Classes lands
* drop teaching of computed on model
* full elimination of non-API properties

## 1. minimal upgrade of syntax

Keeping in mind that changes to the modeling layer are coming, we do the minimal syntactical upgrade and do no rework any documentation further at this time.

## 2. drop teaching of computed on model

[...]

PR #603, opened 2019-03-14
[OCTANE] migrate a few other ember data examples

efx:
Finish #543 
@acorncom this should wrap up the few strictly ember data sections I had recalled. #594 covers the testing section that had some ember data examples as well.


guides/release/tutorial/ember-data.md
```diff
@@ -25,29 +25,31 @@ When we open the model file, we can see a blank class extending [`DS.Model`](htt
...
-  category: DS.attr(),
-  image: DS.attr(),
-  bedrooms: DS.attr(),
-  description: DS.attr()
-});
+const { Model, attr } = DS;
+
+export default class RentalModel extends Model {
+  @attr title;
+  @attr owner;
```
mike-north: Ember-data decorators weren't listed in https://github.com/emberjs/rfcs/pull/408. I pinged @igort to confirm that this is (a) supported and (b) something we want to steer people towards.

guides/release/tutorial/ember-data.md
```diff
@@ -25,29 +25,31 @@ When we open the model file, we can see a blank class extending [`DS.Model`](htt
...
-export default DS.Model.extend({
+export default class RentalModel extends Model {
 
-});
+}
 ```
 
 Let's define the structure of a rental object using the same attributes for our rental that we [previously used](../model-hook/) in our hard-coded array of JavaScript objects -
 _title_, _owner_, _city_, _category_, _image_, _bedrooms_ and _description_.
 Define attributes by giving them the result of the function [`DS.attr()`](https://www.emberjs.com/api/ember-data/release/classes/DS/methods/attr?anchor=attr).
```
jenweber: Can you update the prose to check for other references to the method `DS.attr()`?

efx: @jenweber ah, so update it to `attr()`?

guides/release/tutorial/ember-data.md
efx: cc: @mike-north thanks for fact checking this. I had opened another issue related to this. pzuraq gave some insight in this comment: https://github.com/ember-learn/guides-source/issues/554#issuecomment-469369178

guides/release/tutorial/ember-data.md
efx: I updated each reference to `attr` as well as other references using `DS.something`

guides/release/tutorial/ember-data.md
igorT: I double checked with @runspired and it's supported and people should use it

pzuraq: Oh, one thing that does need to change here though, I'm pretty sure you'll still have to do `@attr()` here, since these are functions that _return_ decorators, and are not decorators themselves. Bit of a pain, but this will be the case until the ecosystem updates to support paren-less invocations.

PR #613, opened 2019-03-15
#588 components/yields

betocantu93:
#588

### DIFF
--- a/guides/release/components/yields.md
+++ b/guides/release/components/yields.md
@@ -18,7 +18,7 @@ compose components just like you would HTML elements:
 Users can pass a block to your component, but by default your component doesn't
 know where to put it. You have to decide this by using the `{{yield}}` helper:
 
-```handlebars {data-filename=app/components/modal-dialog/template.hbs}
+```handlebars {data-filename=app/templates/components/modal-dialog.hbs}
 <dialog>
   {{yield}}
 </dialog>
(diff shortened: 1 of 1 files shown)


guides/release/components/yields.md
```diff
@@ -63,9 +63,9 @@ then nothing will be done with the block.
 You can check whether or not a user passed a block to the component with the
 `hasBlock` helper:
 
-```handlebars {data-filename=app/components/modal-dialog/template.hbs}
+```handlebars {data-filename=app/templates/components/modal-dialog.hbs}
 <dialog>
-  {{#if hasBlock}}
+  {{#if (has-block)}}
```
mixonic: This surprised me as I haven't used `(has-block` the helper before. [I tested out a few variants](https://ember-twiddle.com/9d0fcebdab2a5d0f6e13aec759bf0147?openFiles=templates.components.my-component.hbs%2Ctemplates.components.my-component.hbs).

AFAICT there are no RFCs for any form :see_no_evil: no API docs for any form :hear_no_evil: and the new form of `(has-block` was introduced without a deprecation for the old form :speak_no_evil:.

I guess we teach `(has-block` as a helper, as you have done here. We should update the Ember API docs, as this is [not even listed as a helper](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers). And the `hasBlock` form should get a deprecation RFC.

So I think your change here is correct.

guides/release/components/yields.md
```diff
@@ -63,9 +63,9 @@ then nothing will be done with the block.
 You can check whether or not a user passed a block to the component with the
 `hasBlock` helper:
```
mixonic:
```suggestion
`(has-block)` helper:
```

guides/release/components/yields.md
mixonic: I've confirmed this with some other framework core people :+1:

guides/release/components/yields.md
betocantu93: Good catch!

Issue #611, opened 2019-03-15
[OCTANE] do some CSS hacks to hide the edit pencil and the search bar


jenweber:
Do you like hacky quick fixes? Well, we have a job for you. Add some css that adds {display: none} for the edit pencil icon on every page of the guides, and also hides the search bar.

Why? We are deploying the Octane Guides Preview to a special staging app and neither of these things will work. It's important for UX.

This work should branch from the existing `octane` and make a PR back into `octane`. Thanks for your help!

rwjblue:
Closed by #616

PR #616, opened 2019-03-15
[Octane] CSS hacks to hide the edit pencil and the search bar

thisduck:
Addressing #611 by hiding edit pencil and the search bar.


rwjblue:
Thank you!

PR #617, opened 2019-03-15
[OCTANE] remove double semi, other semi colon

efx:
I just learned the semi-colon's are optional after the class definition and methods so it doesn't *really* matter. Though, I've observed most examples in the octane branch do not include them so I think it is best to make the style consistent. @maciej-ka so don't take my nit picking over semi colons personally 😄 
We could continue @mansona's work (https://github.com/ember-learn/guides-source/pull/589#issue-260513096) and automate prettifying the code snippets.

### DIFF
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -225,7 +225,7 @@ export default class CommentFormComponent extends Component {
 
   @action
   submitComment() {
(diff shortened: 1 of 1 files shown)


acorncom:
Thanks @efx! If you have time, I think automating this style would be quite helpful as we work to encourage a style within Ember

PR #618, opened 2019-03-15
[OCTANE] Fixup Actions and Events page/links

pzuraq:
Update the link to the Actions and Events page so it actually works, and update the page to use the `{{action}}` modifier instead of the `onclick=` property


jenweber:
Thank you!

wycats:
In my DOM guide, I use the `{{on` addon, and I think we should discuss whether we should use it more broadly, or separately, whether we should bring it into core.

Issue #583, opened 2019-03-12
Octane guides review: ensure code examples are complete


mike-north:
The octane guides (#455) are coming along really well!

One thing that we could use some help with is reviewing all markdown "fenced code blocks"

````
```javascript

```
```handlebars

```
````

are...

1. Valid language syntax.

	A good way of checking this is to open up [https://babeljs.io/repl](https://babeljs.io/repl) and paste a code sample in the left editor pane.

2. Complete

	Some of them seem to be missing imports, particularly for decorators like `@tracked`, `@action`, etc... The good news is, other examples are complete, and will provide an indication of where to import from.

    If examples use something like "..." to provide an excerpt, please make sure it's in a comment, so that if readers copy/paste code into their editor, it's only _incomplete_ and not _broken_.

**Comment by mansona** [comment]
Here is a tool that we used to verify all of the code blocks during the AngleBrackets migration: https://github.com/ember-learn/guides-source/pull/589

I don't know how useful it would be to you but it's there if you need it 🎉

**Comment by efx** [comment]
Here are 2 aspects of this I found while working against the octane branch:
https://github.com/ember-learn/guides-source/issues/565
https://github.com/ember-learn/guides-source/issues/579
I think they also apply to master.

**Comment by efx** [comment]
I think there are a couple approaches with short and long term approaches.
1. integrate the extraction script and pass relevant blocks to eslint 
2. Prettier supports parsing markdown and formats embedded snippets. We could run it automatically if the opinions align. To see the changes, try running:
```bash
npx prettier --parser markdown --single-quote --write "guides/release/**/*.md"
```

I did try [integrating prettier](https://github.com/efx/guides-source/blob/try-prettier-blocks/extractCode.js#L45-L55) with the `extractCode` snippet and found it provides helpful errors for some invalid code snippets. Some of the failures were false positives though.
I think full prettier integration could make sense at some point, so perhaps for initial review we could focus on making the extract script provide useful output and possibly copies of the files with the corrected changes.

jenweber:
Closed by a dozen or so PRs ahead of EmberConf

jenweber:
Thanks everyone for your help!

PR #619, opened 2019-03-15
Fix syntax

leros:
Previous syntax gave error:  Invalid block parameters syntax


efx:
Thanks @leros!  this helps address point 1 in https://github.com/ember-learn/guides-source/issues/583#issue-420153850

locks:
Thank you!

PR #620, opened 2019-03-15
DOM documentation

wycats:


jenweber:
A couple high level things:
- the use of both `<Article />` and `<article>` could be confusing
- it’s good to avoid words like “simple” and “straightforward” if there are good alternatives

jenweber:
Thank you @wycats!

PR #627, opened 2019-03-16
Adjust robots.txt and Google Analytics for staging deployment

### LINKED ISSUE #610: Set up robots.txt and GoogleAnalytics for the octane staging site
We have some small adjustments to make to prepare for showing the Octane Guides in a separate staging app. This work should be done from the existing `octane` branch, and the PR should target `octane` and not master.

We need to adjust `public/robots.txt` to reflect the upcoming home of the Octane Guides staging, `octane-guides-preview.emberjs.com/release/`.

We also need to update Google Analytics hostname in the router: [Edit this line](https://github.com/ember-learn/guides-source/blob/master/app/router.js#L30) to be `octane-guides-preview.emberjs.com`

jenweber:
I think this resolves the blocker for standing up `octane-guides-preview.emberjs.com`

closes #610

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -27,7 +27,7 @@ const Router = EmberRouter.extend({
       const title = this.getWithDefault('currentRouteName', 'unknown');
 
       // this is constant for this app and is only used to identify page views in the GA dashboard
-      const hostname = 'guides.emberjs.com';
+      const hostname = 'octane-guides-preview.emberjs.com';
 
       this.metrics.trackPage({ page, title, hostname });
     });
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,3 +1,3 @@
 # http://www.robotstxt.org
 User-agent: *
-Disallow:
+Disallow: /


app/router.js
```diff
@@ -27,7 +27,7 @@ const Router = EmberRouter.extend({
       const title = this.getWithDefault('currentRouteName', 'unknown');
 
       // this is constant for this app and is only used to identify page views in the GA dashboard
-      const hostname = 'guides.emberjs.com';
+      const hostname = 'octane-guides-preview.emberjs.com';
```
tomdale: Is there a reason we hardcode the hostname at all, vs. getting it from `window.location.hostname`?

mansona: this is **purely** for google analytics. And hostname isn't what you think it is here 🙈 it's just a flag to allow us to group things in the GA admin

public/robots.txt
```diff
@@ -1,3 +1,3 @@
 # http://www.robotstxt.org
 User-agent: *
-Disallow:
+Disallow: /
```
tomdale: I'm worried we'll forget to remove this when we merge to master. Is there an alternative to robots.txt to block this domain (like an HTTP header we can configure for the Netlify app?), or should we do something to make really sure we remember?

mansona: I think this is best for now and don't worry I won't miss this when it comes to merging 😉

tomdale:
Thank you!!

public/robots.txt
jenweber: I am tracking our crimes in the main Octane PR so we don’t forget to undo them

PR #634, opened 2019-03-17
#588 - octane - guides update - model/relationships

dennismende:


dennismende:
removed computed and make use of getter to stick to decisions made in #543 and #553

locks:
Hm, I believe these should be tracked properties, otherwise it won't update? cc @pzuraq

mixonic:
I *think* @pzuraq was suggesting this very change in https://github.com/ember-learn/guides-source/pull/553#discussion_r263493091.

pzuraq:
Yup! So, `@attr` is a computed property, which means that when it is encountered by the autotracker it will be autotracked. So we can convert any CPs that rely on that property to standard getters instead 😄 any _non_ `@attr`s/computeds do still need to be converted to tracked properties, but that doesn't apply in this case

guides/release/models/relationships.md
```diff
@@ -182,29 +182,27 @@ export default class PaymentMethod extends Model {
...
 import PaymentMethod from './payment-method';
 const { attr } = DS;
 
 export default class PaymentMethodCc extends PaymentMethod {
   @attr() last4;
 
-  obfuscatedIdentifier: computed('last4', function () {
+  get obfuscatedIdentifier() {
     return `**** **** **** ${this.last4}`;
   })
```
jamescdavis:
```suggestion
   }
```

jenweber:
@dennismende Could you make the two punctuation fixes noted by James (`})` should be `)` in two places)? Then this will be good to merge!

It is helpful if you check "allow maintainers to make changes" when you open PRs, if you are ok with it. It's not required, but it means we can do things like update stale branches or fix small typos without bugging you about it. Thanks for the help!

jenweber:
I pulled these commits by @dennismende into a fresh PR, and then made the typo fixes. It's been merged! Thanks Dennis for your help!

Issue #584, opened 2019-03-12
Octane guides review: back out of MU project layout for file paths


mike-north:
In light of [Module Unification project layout](https://blog.emberjs.com/2019/03/11/update-on-module-unification-and-octane.html) being de-scoped from the octane edition, we need the updates to the guides (#455) to be converted back to "classic" project layout (`/app` and `/addon` folders).

These paths are mostly found in markdown "fenced code blocks"

````
```hbs {data-filename=src/ui/routes/application/template.hbs}

```
````

In this case, we'd convert the file path as follows

#### A top-level template
```diff
-src/ui/routes/application/template.hbs
+app/templates/application.hbs
```

Other examples

#### A component's JS module
```diff
-src/ui/components/hello/component.js
+app/components/hello.js
```

#### A component's HBS module
```diff
-src/ui/components/hello/template.hbs
+app/templates/components/hello.hbs
```

#### A controller

```diff
-src/ui/routes/application/controller.js
+app/controllers/application.js
```

efx:
@betocantu has started addressing this in https://github.com/ember-learn/guides-source/pull/612.

dennismende:
With #636 all occurrences of MU project structure should have been changed back to current structure.

PR #636, opened 2019-03-17
#588 - octane - guides update - non MU file paths

dennismende:


dennismende:
Should close #584 because I couldn't find any other usage of MU file structure in the guides anymore.

locks:
Thanks for the help!

Issue #395, opened 2019-02-07
Octane guides: Component section (tracking issue)


MelSumner:
PM TODO: create issues for each of these sub-sections, and add issue to this tracking issue

Components will need the most refactor work and new content for Octane, so here’s the subtopic breakdown:

- [ ] Creating a component (includes Component Types post-Octane) [issue #396]
- [x] Displaying data in a Template [issue #397]
- [x] Adding actions [issue #398 ]
- [ ] Nesting components (includes suggestions of when to break things into components) [issue #428]
- [x] Passing arguments [issue #431]
- [ ] Using computed properties (extremely basic example, link to the Computed Property section)
- [ ] Working with arrays of data
- [ ] Lifecycle hooks

chancancode:
I'm working on documentation for how to work with DOM in Octane (replacement for lifecycle hooks). It should be in an MVP state in the next day or so.

wycats:
I'm working on documentation for working with the DOM in Octane. It should be ready in an MVP state in the next day or so.

Frozenfire92:
was about to make an issue when I came across this one

![Screen Shot 2019-03-17 at 1 02 40 PM](https://user-images.githubusercontent.com/5893112/54494049-f7c97900-48b4-11e9-8dd6-c226c8db7ecd.png)

It seems there are two links which are covering DOM manipulation

`Interactiving with the DOM`: `/release/components/interacting-with-the-dom/`
`Working with the DOM`: `/release/components/glimmer-components-dom/`

The first of which is broken (non existent file)

I suspect the first heading/url is desired based on ` "Interacting with the DOM", url: "interacting-with-the-dom"` in #588

wycats:
Is this still being updated?

jenweber:
Closed, I think we have hit all these points. Thanks all!

PR #637, opened 2019-03-17
[OCTANE] Update component-basics

Frozenfire92:
This PR updates the route `/release/components/component-basics` for #588

### Summary
- [x] switch to using `{{action this.sayHello}}` instead of `onclick={{this.sayHello}}` as outlined in: https://github.com/ember-learn/guides-source/pull/598#discussion_r265390517
- [x] change code fences to ` ```handlebars` and ` ```javascript`
- [x] correct `action` import
- [x] add component usage filenames of application template
- [x] update example code with `CustomButton` to be simpler and more consistent with other sections of the guides (`Icon` component instead, consistent argument usage)
- [x] fix yield helper reference
- [x] add mention of `Interacting with the DOM` to the further guides footnote. Assuming this is the correct section name as mentioned in: https://github.com/ember-learn/guides-source/issues/395#issuecomment-473680389

### Remaining work
- [ ] comment: `[TODO: Screenshot of the Component tree in the Ember Inspector]` depends on emberjs/ember-inspector#931

### DIFF
--- a/guides/release/components/component-basics.md
+++ b/guides/release/components/component-basics.md
@@ -10,14 +10,14 @@ Like we mentioned in the section on [Templates](../../templates/handlebars-basic
 can have both a template and a class definition, like so:
 
 ```handlebars {data-filename=app/templates/components/hello-button.hbs}
-<button onclick={{this.sayHello}}>
+<button {{action this.sayHello}}>
   {{@buttonText}}
 </button>
 ```
 
-```js {data-filename=app/components/hello-button.js}
+```javascript {data-filename=app/components/hello-button.js}
 import Component from '@glimmer/component';
-import action from '@ember/object';
+import { action } from '@ember/object';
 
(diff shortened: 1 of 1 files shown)


Frozenfire92:
Question:

I noticed when generating components within an octane blueprint app (`ember new my-app -b @ember/octane-app-blueprint`)

they add `Component` to the end of the class name, like so:

```js
export default class HelloButtonComponent extends Component {
```

is this going to be like the case when octane releases? if so we should likely update this example to have `Component` on the end of the class name as well

mike-north:
>  if so we should likely update this example to have Component on the end of the class name as well

I'd go beyond this and ask that code snippets in the guides match exactly what readers will see/experience as they follow along.

jenweber:
Looks good to go! I would like to merge this after #638, since I think it's MVP.

PR #642, opened 2019-03-17
Finish up philosophy

wycats:
Added a few more sections to the DOM guide with more details on the philosophy.


guides/release/components/glimmer-components-dom.md
```diff
@@ -320,6 +359,7 @@ import action from "@ember/object";
 export default class AudioPlayer extends Component {
   @action
   play() {
+    // TODO
```
mike-north: is this intentional?

wycats: Yes -- it gets updated with something else later in the documentation. It's a TODO from the perspective of the docs :P

PR #656, opened 2019-03-18
Octane: Fix improper/missing @action imports

mike-north:


efx:
Looks good to me.
nitpick: the first code snippet uses single quotes and the rest use double but this is something captured here: https://github.com/ember-learn/guides-source/issues/579

guides/release/components/glimmer-components-dom.md
```diff
@@ -433,7 +433,7 @@ We could then use the `modal` component this way:
 
 ```js {data-filename="app/components/sidebar.js"}
 import Component from "@glimmer/component";
-import action from "@ember/object";
+import { action } from "@ember/object";
 
 export default class Sidebar extends Component {
   @tracked showingHelp = false;
```
mixonic: Hm also missing `@tracked` here

guides/release/components/glimmer-components-dom.md
```diff
@@ -173,7 +173,7 @@ If you want to add an event handler to an HTML element, you can use the `{{on` e
 
 ```js {data-filename="app/components/counter.js"}
 import Component from "@glimmer/component";
-import action from "@ember/object";
+import { action } from "@ember/object";
 
 export default class Counter extends Component {
   @tracked count = 0;
```
mixonic: And here :grimacing:

PR #657, opened 2019-03-18
Octane: Quickstart a11y improvements

### LINKED ISSUE #655: Fix "Quickstart" example to use interactive HTML elements
In our [Quickstart example in the Guides](https://guides.emberjs.com/release/getting-started/quick-start/#toc_click-events), we are teaching people to add a click event to a list item, `<li>`, and it should be a `<button>` instead. However `<li>` are not an interactive element, which means that some browsers may not fire the event, it may not receive focus for keyboard nav, and it does not work well with assistive technology like screen readers.

Here's the section to fix. Refactor this to not use `<ul>` and `<li>`. Use an interactive element like `<button>` for the action.

```
<ul>
  {{#each this.people as |person|}}
    <li>{{person}}</li>
    <li {{action "showPerson" person}}>{{person}}</li>
  {{/each}}
</ul>
```

Please also read the writing that goes with it carefully, to see if it refers to the items as lists.

mike-north:
Fixes #655, hopefully in a way that doesn't require too much re-working of associated content

This template change avoids encouraging readers to use click event listeners w/ non-focusable and non-interactive elements

```diff
<h2>{{this.title}}</h2>

<ul>
  {{#each this.people as |person|}}
-    <li>{{person}}</li>
+    <li>
+      <button {{action "showPerson" person}}>{{person}}</button>
+    </li>
  {{/each}}
</ul>
```

### DIFF
--- a/guides/release/getting-started/quick-start.md
+++ b/guides/release/getting-started/quick-start.md
@@ -192,8 +192,6 @@ Note that we've changed the title from a hard-coded string ("List of Scientists"
 We've also renamed `scientist` to the more-generic `person`,
 decreasing the coupling of our component to where it's used.
 
-_Note: While the button element will ensure that your code is accessible, you may require an extra style or two if you wish to have it look like regular text. You might be tempted to use a regular link here, but that will cause your accessibility tests to fail._
-
 Save this template and switch back to the `scientists` template.
 Replace all our old code with our new componentized version.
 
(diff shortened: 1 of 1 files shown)


mixonic:
@mike-north for the purposes of UX (not having the list of users suddenly become a list of buttons) should a) the `button` be next to the name with the text "Show" or b) wrapped in an `<a href="#" {{action 'show' person preventDefault=true}}>` instead so it remains text?

MelSumner:
@mixonic @jamescdavis both are incorrect from the a11y standpoint (acceptable in a pinch, but shouldn't be taught in a guide).

PR #665, opened 2019-03-18
Review "configuring" (part one)

mixonic:
* Configuring Your App - I did some serious re-wording here
* Configuring Ember CLI - Left as is
* Handling Deprecations - Just cleaned up a header. This should probably be reconsidered / balanced with the existence of the "upgrading/deprecations" guide entry.


acorncom:
Thanks!

PR #674, opened 2019-03-19
Fix blueprint name

locks:


acorncom:
The error we're getting here about spelling mistakes is not in this PR

PR #675, opened 2019-03-19
Fix typo in Angle Bracket Syntax section

kgautreaux:


locks:
I happened to see https://github.com/ember-learn/guides-source/pull/676 first, but thank you for taking the time to contribute!

PR #691, opened 2019-03-27
Wrap arrays to avoid prototype extensions

### LINKED ISSUE #532: Guides should not use prototype extensions
The guides should not use prototype extensions. They cause a lot of problems in FastBoot environments, where objects cross realms.

## TODO
- [ ] Remove instances of using prototype extensions in the guides
- [ ] find appropriate place to use a note/callout (Zoey/Tomster says...) - "Prototype extensions can cause a lot of problems in FastBoot (link) environment, where objects cross realms."

rwwagner90:
Fixes #532

### DIFF
--- a/guides/release/object-model/enumerables.md
+++ b/guides/release/object-model/enumerables.md
@@ -4,6 +4,11 @@ objects, and which allows you to work with those children using the
 enumerable in the majority of apps is the native JavaScript array, which
 Ember.js extends to conform to the enumerable interface.
 
+*Note:* Ember used to enabled prototype extensions by default. This can get messy,
+especially in FastBoot, where objects cross realms. We recommend 
+[disabling prototype extensions](https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/) 
+and always manually wrapping your arrays and objects, when you need these methods.
+
 By providing a standardized interface for dealing with enumerables,
 Ember.js allows you to completely change the way your underlying data is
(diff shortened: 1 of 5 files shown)


guides/release/object-model/enumerables.md
```diff
@@ -4,6 +4,11 @@ objects, and which allows you to work with those children using the
 enumerable in the majority of apps is the native JavaScript array, which
 Ember.js extends to conform to the enumerable interface.
 
+*Note:* Ember used to enabled prototype extensions by default. This can get messy,
```
jenweber:
```
If you are using [Ember Fastboot](https://ember-fastboot.com/), we recommend 
[disabling prototype extensions](https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/) 
and always manually wrapping your arrays and objects when you need these methods.
```

I think this should give enough info for an advanced Ember dev to know what to do, while avoiding confusing a newer Ember dev.

jenweber: Attempt # 2:

```
It is best practice to [disable prototype extensions](https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/) in your app. For example, this means that Ember's convenience methods like `firstObject` will only be available on `EmberArray`s and will not work on Native JavaScript arrays. This is especially important for apps that use [Ember Fastboot](https://ember-fastboot.com/).
```

PR #694, opened 2019-03-28
Refactor links to inline

### LINKED ISSUE #330: Markdown link inconsistencies
Since working on the linting and spellchecking improvements, I've noticed that the guides are inconsistent on how markdown links are created.

In the cli-guides, there are all inline `[link name](https://somelink...)` and the guides-source has a mix between inline and keyed links.  In some cases within the same file.  Not sure about other docs.

Key links are in the markdown and the links defined at the bottom of the file

```
[My Link][]

[My Link]: http://www.myspeciallink.com
```

This came about from having different contributors working on the guides.  I raise the question should the format be standardized and the guides refactored to match?

While I normally use inline links in my markdown, the use of keys in large document sets has advantages.

- They improve markdown readability, which makes reviewing PRs easier because the text isn't interrupted by URLs
- If you use the same link multiple times in a document, you only have to define it in one location
- Improves maintainability. If a link needs to be updated, there is only one place to look.  Each document needs to be to visited, but you don't have to search through the document

There are probably more benefits  that I'm missing. The Remark linters will flag keys without a definition, so editing errors will be caught. Dead links are still an issue but I believe @mansona is working a solution for that.

maxwondercorn:
Refactored markdown links to inline for consistency and match learning team requirements.  Links from 3.0.0 on have been converted to inline.

As I started this several weeks ago, I will probably do another PR to clean up links added after I started the refactoring

resolves #330

### DIFF
--- a/.local.dic
+++ b/.local.dic
@@ -77,8 +77,8 @@ mixins
 modularity
 naïve
 nav
-*NPM
 NVDA
+*NPM
 Orca
 Param
 param
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,26 +126,17 @@ When linking to API pages:
 
 ## Spellchecking/linting
 
-The guides are spellchecked and linted for markdown consistency. You can test your contributions by running `npm run lint:md`. Linting and spellchecking must pass or they will fail in Travis-CI. 
+The guides are spellchecked and linted for markdown consistency. You can test your contributions by running `npm run lint:md`. Linting and spellchecking must pass or testing in Travis-CI will fail. 
 
-Markdown issues that will generate errors include:
+In addition to spelling errors, the following markdown issues that will generate errors:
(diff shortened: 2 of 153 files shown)


locks:
Can you rebase? 🙈

maxwondercorn:
working on it

maxwondercorn:
@locks It looks like I remerged the branch again when trying to rebase.  I've only rebased a few times but this one was different then what I've seen in the past.  I don't know if this can all be squashed into one commit or I need to do something on my end?  We also could close the PR and I can generate a new one based on updated master.

jessica-jordan:
@maxwondercorn Do you mind rebasing your branch onto master or send another PR with your commits based on the current master to the repo? Happy to review as well!

maxwondercorn:
I'll just do another PR easier all the way around. Hopefully, I can do it later today

Issue #692, opened 2019-03-28
Switch to <AngleBracket /> built-in components on Octane branch (input, link-to, etc)


chancancode:
It's time to switch the Octane guides to use <AngleBracket /> built-in components!

* `{{link-to}}` -> `<LinkTo />`
* `{{input}}` -> `<Input />`
* `{{textarea}}` -> `<Textarea />`

This should mostly just be a syntactic change, but we could also take the opportunity to adjust the flow, remove legacy concepts, etc. For example, since angle bracket invocation takes HTML attributes, we can update the examples to pass them directly (i.e. `class=` instead of `@class=`).

To help out, please leave a comment to "claim" the item (🔒 below) so others know not to duplicate the effort! It's fine to claim the item before you have a PR ready (but please release it if it turns out you didn't have the time to do it). Branch from the `octane` branch of the Guides and open your PR to be merged back into `octane`. Thank you!

Some resources:

* See PR #681 for an example
* [RFC #459](https://emberjs.github.io/rfcs/0459-angle-bracket-built-in-components.html) for details about the feature
* [API docs PR](https://github.com/emberjs/ember.js/pull/17810)

Guides that need to be updated:

* [x] `components/actions-and-events.md` (:lock: @alexBaizeau ~~#697~~)
* [x] `models/index.md` (🔒 @chancancode ~~#681~~)
* [x] `routing/defining-your-routes.md` (🔒 @chancancode ~~#681~~ ~~#693~~)
* [x] `routing/preventing-and-retrying-transitions.md` (🔒 @chancancode ~~#681~~)
* [x] `routing/query-params.md` (:lock: @alexBaizeau ~~#698~~)
* [x] `routing/specifying-a-routes-model.md` (:lock: @Charizard ~~#699~~)
* [x] `templates/binding-element-attributes.md` (🔒 @chancancode ~~#681~~)
* [x] `templates/input-helper.md` (:lock: @ddoria921 #696)
* [x] `templates/links.md` (🔒 @chancancode ~~#681~~)
* [x] `testing/testing-components.md` (:lock: @Charizard ~~#700~~)
* [x] `tutorial/auto-complete-component.md` (:lock: @zachgarwood ~~#702~~)
* [x] `tutorial/model-hook.md` (:lock: @mdbiscan ~~#704~~)
* [x] `tutorial/routes-and-templates.md` (:lock: @ClaytonTurner ~~#705~~)
[...]

PR #698, opened 2019-03-28
Update docs for the built-in components in `routing/query-params.md`

alexBaizeau:
As per #692

### DIFF
--- a/guides/release/routing/query-params.md
+++ b/guides/release/routing/query-params.md
@@ -83,17 +83,18 @@ With this code, we have established the following behaviors:
    full router transition (i.e. it won't call `model` hooks and
    `setupController`, etc.); it will only update the URL.
 
-### link-to Helper
+### LinkTo component
 
-The `link-to` helper supports specifying query params using the
-`query-params` subexpression helper.
+The `<LinkTo />` component supports specifying query params using the `@query`
+argument, along with the `{{hash}}` helper
 
 ```handlebars
 // Explicitly set target query params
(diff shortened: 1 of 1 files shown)


alexBaizeau:
I have a hard time figuring out why the build failed, any pointer?

guides/release/routing/query-params.md
```diff
@@ -83,25 +83,26 @@ With this code, we have established the following behaviors:
    full router transition (i.e. it won't call `model` hooks and
    `setupController`, etc.); it will only update the URL.
 
-### link-to Helper
+### LinkTo component
```
chancancode: I think we probably want `<LinkTo />` here to be consistent, but I'm not 100% sure 🤔

guides/release/routing/query-params.md
```diff
@@ -83,25 +83,26 @@ With this code, we have established the following behaviors:
...
+
 ```
 
 In the above examples, `direction` is presumably a query param property
 on the `posts` controller, but it could also refer to a `direction` property
 on any of the controllers associated with the `posts` route hierarchy,
 matching the leaf-most controller with the supplied property name.
 
-The `link-to` helper takes into account query parameters when determining
+The `<LinkTo>` component takes into account query parameters when determining
```
chancancode: `<LinkTo />`, to be consistent (I'm actually not a 100% sure if we want to consistently do `<LinkTo>` or consistently do `<LinkTo />`; but so far we have been doing the latter and I think it's better to keep it consistent and fix everything in one go later if we want to)

PR #707, opened 2019-03-29
Update subroute guides to use angle bracket built-ins

dknutsen:
Update the subroute page of the guides to use angle bracket invocation for link-to helper.
(Part of #692)


pzuraq:
Thanks a bunch 😄

PR #708, opened 2019-03-29
Refactor "Specifying a route's model" to be agnostic of data library choice

jenweber:
See https://github.com/ember-learn/guides-source/issues/20 for the motivation. In short, in order to make Ember Data guides a standalone resource and learning concept, we need to refactor how we teach data fetching in other parts of the app. I've made this article longer, and I don't take that lightly, but based on my experience teaching others, we can't skimp on this. Understanding models is one of the greatest new learner pain points, so more examples and simpler language are critical.

If there are controversial refactors, I can break this into multiple PRs, but I think it's best evaluated as a whole.

This PR is a broad refactor of https://guides.emberjs.com/release/routing/specifying-a-routes-model/

## Here's how I recommend reviewing:
- Read this entire PR comment first
- Read my version and see if it checked all the boxes that the old version did (the diff is going to be hard to read)
- Scrutinize the code samples for mistakes
- See if the raw `new Promise` example makes any sense. I am skeptical.
- Look for typos

## Changes

Here's a summary of the things I changed and the things I didn't.

- I did not destroy or edit any heading anchors, aside from making one header one level deeper
- I moved "multiple models" higher up the page
- I added two new heading anchors within the existing content, "Linking to a dynamic segment" and "Creating a dynamic segment" since this article is so long
- I did not remove any learning concept scope
- I added more examples of fetching data in the model
- I created a debugging section
- I wrote a proper intro. The article introduced models like someone already knows what they are and what they do.
- incorporated generating a dynamic segment using the Ember CLI
- strengthen the "why" for using dynamic segments

## Rationale for the biggest changes

The biggest changes are:

- Adding more (non-Ember-Data) examples for what to return from the model hook.
    - This is motivated by #20, the desire to make Ember Data learnable as a stand-alone resource. It's still worth mentioning, I think.
    - The `ember-ajax` example is provided based on discussion in #20.
    - The "static data" example is based on trying to help a new JS dev who was stuck for a week on models because they couldn't tell whether they were doing async wrong or doing something wrong in the template.
    - The raw Promise example is to drive home that you can do whatever you want in the model hook when it comes to async.
- Adding a debugging section.
    - This is based on personal experience teaching other devs, who needed help every time they were writing a model hook from scratch. I would have loved to send them a link to this.
- Adding a summary statement of the powers of model hooks.
    - This is providing some rationale for loading data in the route instead of hand-rolling async stuff in components.
- I changed the dynamic segment example to use `:id` instead of `:photos_id`
    - The underscore will make a modern Ember app's linter yell at you. Dashes are pain.
- Clarified that dynamic segments can have any key name that you feel like in `params`. It's not something magical.

## Learning topics that are covered

In the old version, the following learning topics were covered. This refactor fulfills the scope of the old version as follows:

- [ ] One job of a route is to display data in the model
- [ ] Show example `router.js`
- [ ] Show a model hook in use, which returns and Ember Data promise
- [ ] describe that a model hook can return any Promise
- [ ] say that the model is received by a Controller and then provided to a template
- [ ] show what a dynamic segment looks like and why it should be used
- [ ] Explain gotchas for link-tos and dynamic segments
- [ ] handling multiple models (totally unchanged except moving it up the page)
- [ ] Reusing route context (totally unchanged)

### DIFF
--- a/guides/release/routing/specifying-a-routes-model.md
+++ b/guides/release/routing/specifying-a-routes-model.md
@@ -38,7 +38,7 @@ or a JavaScript data fetching library of your choice
 5. When you load data in the `model` hook, you can take advantage of other niceties that Ember provides, like 
 [automatic route transitions](/preventing-and-retrying-transitions)
 after the data is returned,
-[error handling](/loading-and-error-substates),
+[loading screens, error handling](/loading-and-error-substates),
 and more
 6. An optional feature of routes is the ability to show a [`loading` template](/loading-and-error-substates) while the app is waiting for the `model` hook to resolve
 7. The `model` hook can be automatically re-run under certain conditions, described later in this topic.
--- a/guides/release/routing/specifying-a-routes-model.md
+++ b/guides/release/routing/specifying-a-routes-model.md
@@ -1,4 +1,4 @@
-A route's JavaScript file is one of the best places in an app to make requests to a back-end server.
+A route's JavaScript file is one of the best places in an app to make requests to an API.
 In this section of the guides, you'll learn how to use the
 [`model`](http://api.emberjs.com/ember/3.8/classes/Route/methods/model?anchor=model)
 function to fetch data and render it in a route's `hbs` template, or pass it down to a component.
@@ -35,7 +35,7 @@ export default Route.extend({
 [Ember Data](../../models/),
 [`ember-ajax`](https://github.com/ember-cli/ember-ajax),
 or a JavaScript data fetching library of your choice
-5. When you load data in a `model`, you can take advantage of other hooks, like 
+5. When you load data in the `model` hook, you can take advantage of other niceties that Ember provides, like 
 [automatic route transitions](/preventing-and-retrying-transitions)
 after the data is returned,
 [error handling](/loading-and-error-substates),
@@ -45,7 +45,7 @@ and more
 
 ## Using the `model` hook
 
-To start, here's an example of returning static sample data from a `model`. This is helpful while you are doing the initial development of a new route:
+To start, here's an example of returning static sample data from the `model` hook. This is helpful while you are doing the initial development of a new route:
 
 ```javascript {data-filename=app/routes/favorite-posts.js}
 import Route from '@ember/routing/route';
@@ -187,15 +187,15 @@ This section will focus on dynamic segments, sometimes also referred to as "dyna
 ### What does a dynamic segment look like?
 
 Here's what it would look like in the URL if a route has a dynamic segment: `http://localhost:4200/posts/42`. The number `42` is the "dynamic" part.
-A user could follow a link taking them to `http://localhost:4200/posts/3` and see post number 3 instead.
+A user can follow a link taking them to `http://localhost:4200/posts/3` and see post number 3 instead.
 What's important here is that the data that should be displayed is tied to the url.
-When you are doing routing well, a user could refresh the page, or visit a bookmarked page, and they would see the right data.
+When you are doing routing well, a user can refresh the page, or visit a bookmarked page, and they will see the right data.
 
-URLs are one of the main superpowers of browser-based apps, and your app should take advantage of them.
+URLs are one of the main superpowers of browser-based apps, and Ember makes it easy to take advantage of them.
 
 ### Creating a dynamic segment
 
-A dynamic segment can be created using the Ember CLI:
+A dynamic segment can be created using Ember CLI:
 
 ```bash
 ember generate route photo --path "photos/:id"
@@ -216,7 +216,7 @@ import Route from '@ember/routing/route';
 
 export default Route.extend({
   model(params) {
-    console.log('This came from the URL: ', params.id)
+    console.log('This came from the URL: ' + params.id)
   }
 });
 ```
--- a/guides/release/routing/specifying-a-routes-model.md
+++ b/guides/release/routing/specifying-a-routes-model.md
@@ -1,5 +1,7 @@
-Often, you'll want a template to display data from a model. Loading the
-appropriate model is one job of a route.
+A route's JavaScript file is one of the best places in an app to make requests to a back-end server.
+In this section of the guides, you'll learn how to use the
+[`model`](http://api.emberjs.com/ember/3.8/classes/Route/methods/model?anchor=model)
+function to fetch data and render it in a route's `hbs` template, or pass it down to a component.
 
 For example, take this router:
 
@@ -9,117 +11,123 @@ Router.map(function() {
 });
 ```
 
-To load a model for the `favorite-posts` route, you would use the [`model()`](https://www.emberjs.com/api/ember/release/classes/Route/methods/model?anchor=model)
-hook in the `favorite-posts` route handler:
+When a user first visits the `/favorite-posts` route, the `model` hook in `app/routes/favorite-posts.js` will automatically run.
+In Ember, functions that automatically run during rendering or setup are commonly referred to as "hooks". 
+Here's an example of a model hook in use within a route:
 
 ```javascript {data-filename=app/routes/favorite-posts.js}
 import Route from '@ember/routing/route';
 
 export default Route.extend({
   model() {
-    return this.store.query('post', { favorite: true });
+    console.log('The model hook just ran!')
+    // write code to return a Promise or some static data
   }
 });
 ```
 
-Typically, the `model` [hook](../../getting-started/core-concepts/#toc_hooks) should return an [Ember Data](../../models/) record,
-but it can also return any [promise](https://www.promisejs.org/) object (Ember Data records are promises/),
-or a plain JavaScript object or array.
-Ember will wait until the data finishes loading (until the promise is resolved/) before rendering the template.
+`model` hooks have some special powers:
 
-The route will then set the return value from the `model` hook as the `model` property of the controller.
-You will then be able to access the controller's `model` property in your template:
+1. When you return data from this model, it becomes automatically available in the route's `.hbs` file as `this.model`
+2. A `model` hook can return either a JavaScript [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises), or static sample data, like a string, object, or array
+3. If you return a Promise from the model hook, your route will wait for the Promise to resolve before it renders the template
+4. `model` hooks work great with
+[Ember Data](../../models/),
+[`ember-ajax`](https://github.com/ember-cli/ember-ajax),
+or a JavaScript data fetching library of your choice
+5. When you load data in a `model`, you can take advantage of other hooks, like redirecting after the data is returned, error handling, and more
+6. An optional feature of routes is the ability to show a `loading` template while the app is waiting for the `model` hook to resolve
+7. The `model` hook can be automatically re-run under certain conditions, described later in this topic.
 
-```handlebars {data-filename=app/templates/favorite-posts.hbs}
-<h1>Favorite Posts</h1>
-{{#each this.model as |post|}}
-  <p>{{post.body}}</p>
-{{/each}}
-```
+## Using the `model` hook
 
-## Dynamic Models
+To start, here's an example of returning static sample data from a `model`. This is helpful while you are doing the initial development of a new route:
 
-Some routes always display the same model. For example, the `/photos`
-route will always display the same list of photos available in the
-application. If your user leaves this route and comes back later, the
-model does not change.
-
-However, you will often have a route whose model will change depending
-on user interaction. For example, imagine a photo viewer app. The
-`/photos` route will render the `photos` template with the list of
-photos as the model, which never changes. But when the user clicks on a
-particular photo, we want to display that model with the `photo`
-template. If the user goes back and clicks on a different photo, we want
-to display the `photo` template again, this time with a different model.
-
-In cases like this, it's important that we include some information in
-the URL about not only which template to display, but also which model.
-
-In Ember, this is accomplished by defining routes with [dynamic
-segments](../defining-your-routes/#toc_dynamic-segments).
-
-Once you have defined a route with a dynamic segment,
-Ember will extract the value of the dynamic segment from the URL for
-you and pass them as a hash to the `model` hook as the first argument:
-
-```javascript {data-filename=app/router.js}
-Router.map(function() {
-  this.route('photo', { path: '/photos/:photo_id' });
-});
-```
-
-```javascript {data-filename=app/routes/photo.js}
+```javascript {data-filename=app/routes/favorite-posts.js}
 import Route from '@ember/routing/route';
 
 export default Route.extend({
-  model(params) {
-    return this.store.findRecord('photo', params.photo_id);
+  model() {
+    return [
+      { title: 'Ember Roadmap' },
+      { title: 'Accessibility in Ember' },
+      { title: 'EmberConf Recap' }
+    ]
   }
 });
 ```
 
-In the `model` hook for routes with dynamic segments, it's your job to
-turn the ID (something like `47` or `post-slug`) into a model that can
-be rendered by the route's template. In the above example, we use the
-photo's ID (`params.photo_id`) as an argument to Ember Data's `findRecord`
-method.
+Now that data can be used in the `favorite-posts`  template:
 
-Note: A route with a dynamic segment will always have its `model` hook called when it is entered via the URL.
-If the route is entered through a transition (e.g. when using the [link-to](../../templates/links/) Handlebars helper),
-and a model context is provided (second argument to `link-to`), then the hook is not executed.
-If an identifier (such as an id or slug) is provided instead then the model hook will be executed.
-
-For example, transitioning to the `photo` route this way won't cause the `model` hook to be executed (because `link-to`
-was passed a model/):
-
-```handlebars {data-filename=app/templates/photos.hbs}
-<h1>Photos</h1>
-{{#each this.model as |photo|}}
-  <p>
-    {{#link-to "photo" photo}}
-      <img src="{{photo.thumbnailUrl}}" alt="{{photo.title}}" />
-    {{/link-to}}
-  </p>
+```handlebars {data-filename=app/templates/favorite-posts.hbs}
+{{#each this.model as |post|}}
+  <div>
+    {{post.title}}
+  </div>
 {{/each}}
 ```
 
-while transitioning this way will cause the `model` hook to be executed (because `link-to` was passed `photo.id`, an
-identifier, instead):
+Let's compare with examples that have outgoing, asynchronous requests to the back end.
 
-```handlebars {data-filename=app/templates/photos.hbs}
-<h1>Photos</h1>
-{{#each this.model as |photo|}}
-  <p>
-    {{#link-to "photo" photo.id}}
-      <img src="{{photo.thumbnailUrl}}" alt="{{photo.title}}" />
-    {{/link-to}}
-  </p>
+This next example uses Ember Data's `findAll` method, which returns a Promise, and resolves with an array of [Ember Data records](../../models/). Ember Data is a powerful (but optional) library included by default in new Ember apps. Note that Ember Data also has a feature called a `Model`, but it's a separate concept from a route's `model` hook.
+
+```javascript {data-filename=app/routes/favorite-posts.js}
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  model() {
+    return this.store.findAll('posts');
+  }
+});
+```
+
+Now let's look at [`ember-ajax`](https://github.com/ember-cli/ember-ajax), an different data-fetching library that you could install in your app.
+It's a convenient wrapper around `jQuery.ajax`, a popular general JavaScript library. Like Ember Data, it returns data from within a Promise:
+
+```javascript {data-filename=app/routes/favorite-posts.js}
+import Route from '@ember/routing/route';
+import { inject as service } from `@ember/service`
+
+export default Route.extend({
+  ajax: Ember.inject.service(),
+  model(params) {
+    return this.get('ajax').request('https://some-api-endpoint.com/posts', {
+      method: 'GET'
+    });
+  }
+});
+```
+
+Lastly, you could hand-write a promise that you return from the model hook:
+
+```javascript {data-filename=app/routes/favorite-posts.js}
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  model() {
+    return new Promise((resolve, reject) => {
+      // get some data
+      resolve( /* data goes here */);
+    })
+  }
+});
+```
+
+Just like in the static data example, the results of an asynchronous request in the `model` hook can be used in the template:
+
+```handlebars {data-filename=app/templates/favorite-posts.hbs}
+{{#each this.model as |post|}}
+  <div>
+    {{post.title}}
+  </div>
 {{/each}}
 ```
 
-Routes without dynamic segments will always execute the model hook.
+Behind the scenes, what is happening is that the route's controller receives the results of the model hook, and makes those results available to the template. Your app may not have a controller file for the route, but the behavior is the same regardless.
 
 ## Multiple Models
+
+What should you do if you need the `model` to return the results of multiple API requests?
 
 Multiple models can be returned through an
 [RSVP.hash](https://www.emberjs.com/api/ember/release/classes/rsvp/methods/hash?anchor=hash).
@@ -163,11 +171,118 @@ each record in the song model and album model:
 
 If you use [Ember Data](../../models/) and you are building an `RSVP.hash` with the model's relationship, consider instead properly setting up your [relationships](../../models/relationships/) and letting Ember Data take care of loading them.
 
+## Dynamic Models
+
+In the examples above, we showed a route that will always return the same data, a collection of favorite posts. Even when the user leaves and re-enters the `/favorite-posts` route, they will see the same thing.
+But what if you need to request different data after user interaction?
+What if a specific post should load based on the URL that the user visited, or a selection that was made?
+In Ember, this is accomplished by defining routes with [dynamic
+segments](../defining-your-routes/#toc_dynamic-segments), or by using [query parameters](/query-params).
+This section will focus on dynamic segments, sometimes also referred to as "dynamic models or routes."
+
+### What does a dynamic segment look like?
+
+Here's what it would look like in the URL if a route has a dynamic segment: `http://localhost:4200/posts/42`. The number `42` is the "dynamic" part.
+A user could follow a link taking them to `http://localhost:4200/posts/3` and see post number 3 instead.
+What's important here is that the data that should be displayed is tied to the url.
+When you are doing routing well, a user could refresh the page, or visit a bookmarked page, and they would see the right data.
+
+URLs are one of the main superpowers of browser-based apps, and your app should take advantage of them.
+
+### Creating a dynamic segment
+
+A dynamic segment can be created using the Ember CLI:
+
+```bash
+ember generate route photo --path "photos/:id"
+```
+
+This command will create a `.js` file, an `.hbs` file, and add a route in `router.js`:
+
+```javascript {data-filename=app/router.js}
+Router.map(function() {
+  this.route('photo', { path: '/photos/:id' });
+});
+```
+
+Whatever shows up in the URL at the `:id`, the dynamic segment, will be available in the params for the route's `model` hook.
+
+```javascript {data-filename=app/routes/photo.js}
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  model(params) {
+    console.log('This came from the URL: ', params.id)
+  }
+});
+```
+
+The `:id` name here isn't something special.
+Whatever name you give the dynamic segment, that's the key it will show up as in the `model` parameters.
+You could call it `:puppies` in `router.js` if you wanted to, and access it in the model as `params.puppies`.
+
+In the `model` hook for routes with dynamic segments, it's your job to
+use the ID (something like `47` or `post-slug`) when you make a data request.
+
+For example, if your app uses Ember Data to make API requests, you could include the ID in a `findRecord`:
+
+```javascript {data-filename=app/routes/photo.js}
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  model(params) {
+    this.store.findRecord('photo', params.id)
+  }
+});
+```
+
+### Linking to a dynamic segment
+
+There are two ways to link to a dynamic segment from an `.hbs` template using the
+[link-to](../../templates/links/)
+helper.
+Depending on which approach you use, it will affect whether that route's model hook is run.
+To learn how to link to a dynamic segment from within the JavaScript file, see the API documentation on
+[`transitionTo`](https://api.emberjs.com/ember/3.8/classes/RouterService/methods/transitionTo?anchor=transitionTo)
+instead.
+
+When you provide a string or number to the `link-to`, the dynamic segment's model hook will run when the app transitions to the new route.
+In this example, `photo.id` might have a value that's an ID like `4`:
+
+```handlebars
+{{#each model as |photo|}}
+  {{#link-to "photo" photo.id}}
+    link text to display
+  {{/link-to}}
+{{/each}}
+```
+
+However, if you provide the entire model context, the model hook for that URL segment will _not_ be run. In this example, we are passing the entire `photo` record:
+
+```handlebars
+{{#each model as |photo|}}
+  {{#link-to "photo" photo}}
+    link text to display
+  {{/link-to}}
+{{/each}}
+```
+
+If a route you are trying to link to has multiple dynamic segments, be sure to specify all the necessary information for each segment. For example, a route like `/photos/4/comments/18` would need:
+
+```handlebars
+{{#link-to "photos.photo.comments.comment" 4 18}}
+  link text to display
+{{/link-to}}
+```
+
+Routes without dynamic segments will always execute the model hook.
+
 ## Reusing Route Context
 
 Sometimes you need to fetch a model, but your route doesn't have the parameters, because it's
 a child route and the route directly above or a few levels above has the parameters that your route
 needs.
+You might run into this if you have a URL like `/photos/4/comments/18`, and when you're in the comments route, you need a photo ID.
 
 In this scenario, you can use the `paramsFor` method to get the parameters of a parent route.
 
@@ -224,3 +339,14 @@ export default Route.extend({
 ```
 
 And calling `modelFor` returned the result of the `model` hook.
+
+## Debugging models
+
+If you are having trouble getting a model's data to show up in the template, here are some tips:
+
+- return static, sample data as a test to see if the problem is really in the model hook, or elsewhere down the line
+- study JavaScript Promises in general, to make sure you are returning data from the Promise correctly
+- make sure your `model` hook has a `return`
+- check to see whether the data returned from a `model` hook is an object, array, or JavaScript Primitive. For example, if the results of `model` are an array, using `{{this.model}}` in the template won't work. You will need to iterate over the results with an `{{#each}}` helper. If the results of `model` are a single object, you need to access the individual attributes like `{{this.model.title}}` to see anything rendered
+- use your browser's development tools to examine the outgoing and incoming API responses and see if they match what your code expects
+- if you are using Ember Data, use the [Ember Inspector](../../ember-inspector/) browser plugin to see if the records made it into the store
--- a/guides/release/routing/specifying-a-routes-model.md
+++ b/guides/release/routing/specifying-a-routes-model.md
@@ -35,8 +35,12 @@ export default Route.extend({
 [Ember Data](../../models/),
 [`ember-ajax`](https://github.com/ember-cli/ember-ajax),
 or a JavaScript data fetching library of your choice
-5. When you load data in a `model`, you can take advantage of other hooks, like redirecting after the data is returned, error handling, and more
-6. An optional feature of routes is the ability to show a `loading` template while the app is waiting for the `model` hook to resolve
+5. When you load data in a `model`, you can take advantage of other hooks, like 
+[automatic route transitions](/preventing-and-retrying-transitions)
+after the data is returned,
+[error handling](/loading-and-error-substates),
+and more
+6. An optional feature of routes is the ability to show a [`loading` template](/loading-and-error-substates) while the app is waiting for the `model` hook to resolve
 7. The `model` hook can be automatically re-run under certain conditions, described later in this topic.
 
 ## Using the `model` hook
--- a/guides/release/routing/specifying-a-routes-model.md
+++ b/guides/release/routing/specifying-a-routes-model.md
@@ -1,7 +1,7 @@
 A route's JavaScript file is one of the best places in an app to make requests to an API.
 In this section of the guides, you'll learn how to use the
-[`model`](http://api.emberjs.com/ember/3.8/classes/Route/methods/model?anchor=model)
-function to fetch data and render it in a route's `hbs` template, or pass it down to a component.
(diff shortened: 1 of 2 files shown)


guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  model(params) {
+    this.store.findRecord('photo', params.id)
+  }
+});
+```
+
+### Linking to a dynamic segment
```
acorncom: Thank you **SO MUCH** for documenting this, it's been a snag that keeps hitting folks ...

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -224,3 +343,14 @@ export default Route.extend({
...
+## Debugging models
+
+If you are having trouble getting a model's data to show up in the template, here are some tips:
+
+- return static, sample data as a test to see if the problem is really in the model hook, or elsewhere down the line
+- study JavaScript Promises in general, to make sure you are returning data from the Promise correctly
+- make sure your `model` hook has a `return`
+- check to see whether the data returned from a `model` hook is an object, array, or JavaScript Primitive. For example, if the results of `model` are an array, using `{{this.model}}` in the template won't work. You will need to iterate over the results with an `{{#each}}` helper. If the results of `model` are a single object, you need to access the individual attributes like `{{this.model.title}}` to see anything rendered
+- use your browser's development tools to examine the outgoing and incoming API responses and see if they match what your code expects
+- if you are using Ember Data, use the [Ember Inspector](../../ember-inspector/) browser plugin to see if the records made it into the store
```
acorncom: Do we want to mention the `{{debugger}}` and ``{{log this.model}}` helpers here too?

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
 
-The route will then set the return value from the `model` hook as the `model` property of the controller.
-You will then be able to access the controller's `model` property in your template:
+1. When you return data from this model, it becomes automatically available in the route's `.hbs` file as `this.model`
+2. A `model` hook can return either a JavaScript [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises), or static sample data, like a string, object, or array
+3. If you return a Promise from the model hook, your route will wait for the Promise to resolve before it renders the template
+4. `model` hooks work great with
+[Ember Data](../../models/),
+[`ember-ajax`](https://github.com/ember-cli/ember-ajax),
+or a JavaScript data fetching library of your choice
```
mike-north: In the spirit of keeping things as simple as possible, I think we may be able to eliminate this point as well.

If there's a feeling that we should point that this works well with data fetching tools, we may want to consider reaching for the simplest and most universal (to those unfamiliar with Ember) examples that will do the job 
(i.e., [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), or potentially AJAX)

Talking about core browser APIs instead of these libraries is a big help at reducing the feeling that we're introducing too much complexity too early

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
+5. When you load data in the `model` hook, you can take advantage of other niceties that Ember provides, like 
+[automatic route transitions](/preventing-and-retrying-transitions)
+after the data is returned,
+[error handling](/loading-and-error-substates),
```
mike-north:
```suggestion
[loading screens, error handling](/loading-and-error-substates),
```

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
+and more
+6. An optional feature of routes is the ability to show a [`loading` template](/loading-and-error-substates) while the app is waiting for the `model` hook to resolve
```
mike-north: I think you can combine this with the point above

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
+7. The `model` hook can be automatically re-run under certain conditions, described later in this topic.
```
mike-north:
```suggestion
7. The `model` hook may automatically re-run in certain situations, as you'll read about below.
```

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
+
+## Using the `model` hook
+
+To start, here's an example of returning static sample data from the `model` hook. This is helpful while you are doing the initial development of a new route:
```
mike-north:
```suggestion
To start, here's an example of returning a simple array from the `model` hook. Even if we eventually plan to fetch this data over a network, starting with something simple makes initial development of a new route quick and easy.
```

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
 {{#each this.model as |post|}}
-  <p>{{post.body}}</p>
+  <div>
+    {{post.title}}
+  </div>
 {{/each}}
 ```
 
-## Dynamic Models
+Let's compare with examples that have outgoing, asynchronous requests to the back end.
```
mike-north:
```suggestion
Let's compare with examples that make asynchronous HTTP requests to a server somewhere.
```

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
+export default Route.extend({
+  model() {
+    return this.store.findAll('posts');
+  }
 });
 ```
 
-```javascript {data-filename=app/routes/photo.js}
+Now let's look at [`ember-ajax`](https://github.com/ember-cli/ember-ajax), an different data-fetching library that you could install in your app.
+It's a convenient wrapper around `jQuery.ajax`, a popular general JavaScript library. Like Ember Data, it returns data from within a Promise:
```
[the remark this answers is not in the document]
mike-north: +1000 for eliminating `ember-ajax` from the guides (and blueprints), and either using `ember-fetch` or `fetch`. Let's show readers how simple Ember can be.

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
-  </p>
-{{/each}}
+Lastly, you could hand-write a promise that you return from the model hook:
+
+```javascript {data-filename=app/routes/favorite-posts.js}
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  model() {
+    return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
mike-north: Aren't we teaching readers about promises in general at this point, and not about Ember? Can we get away with eliminating this example?

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
+```handlebars {data-filename=app/templates/favorite-posts.hbs}
+{{#each this.model as |post|}}
+  <div>
+    {{post.title}}
+  </div>
 {{/each}}
 ```
 
-Routes without dynamic segments will always execute the model hook.
+Behind the scenes, what is happening is that the route's controller receives the results of the model hook, and makes those results available to the template. Your app may not have a controller file for the route, but the behavior is the same regardless.
```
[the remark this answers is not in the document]
mike-north: or possibly http://api.emberjs.com/ember/release/classes/Route/methods/model?anchor=setupController

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+```
+
+### Linking to a dynamic segment
+
+There are two ways to link to a dynamic segment from an `.hbs` template using the
+[link-to](../../templates/links/)
+helper.
+Depending on which approach you use, it will affect whether that route's model hook is run.
+To learn how to link to a dynamic segment from within the JavaScript file, see the API documentation on
+[`transitionTo`](https://api.emberjs.com/ember/3.8/classes/RouterService/methods/transitionTo?anchor=transitionTo)
```
mike-north:
```suggestion
[`transitionTo`](https://api.emberjs.com/ember/release/classes/RouterService/methods/transitionTo?anchor=transitionTo)
```

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+
+```handlebars
+{{#each model as |photo|}}
+  {{#link-to "photo" photo}}
+    link text to display
+  {{/link-to}}
+{{/each}}
+```
+
+If a route you are trying to link to has multiple dynamic segments, be sure to specify all the necessary information for each segment. For example, a route like `/photos/4/comments/18` would need:
```
mike-north: This _might_ be a good place to point out that multiple dynamic segments with the same name are all filled in with the same value. 

i.e., `/photos/:id/comments/:id` could never be `/photos/4/comments/18` since there is exactly one segment in the URL pattern, which is repeated twice

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+
+```handlebars
+{{#each model as |photo|}}
+  {{#link-to "photo" photo.id}}
+    link text to display
+  {{/link-to}}
+{{/each}}
+```
+
+However, if you provide the entire model context, the model hook for that URL segment will _not_ be run. In this example, we are passing the entire `photo` record:
```
mike-north: I encourage folks to typically pass ids instead of models so that they get consistent behavior regardless of whether they enter the application through a particular route, or transition to it from within the app.

I feel like if I were to teach that it's ok to do either of these things, I'd also have to encourage that each route be involved in acceptance tests _both_ by being visited directly, _and_ by being transitioned to while skipping the model hook.

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+If a route you are trying to link to has multiple dynamic segments, be sure to specify all the necessary information for each segment. For example, a route like `/photos/4/comments/18` would need:
+
+```handlebars
+{{#link-to "photos.photo.comments.comment" 4 18}}
+  link text to display
+{{/link-to}}
+```
+
+Routes without dynamic segments will always execute the model hook.
+
```
mike-north: An additional common thing that trolls people, which I didn't see is mention of, the default route behavior for dynamic segments like formatted like `:book_id`.

For a route with no model hook implemented (even a route with no `.js` file at all), the default implementation of `model()` will be equivalent to

```js
model(params) {
 return this.store.find('book', params.book_id);
}
```

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -224,3 +343,14 @@ export default Route.extend({
...
 
 And calling `modelFor` returned the result of the `model` hook.
+
+## Debugging models
+
+If you are having trouble getting a model's data to show up in the template, here are some tips:
+
+- return static, sample data as a test to see if the problem is really in the model hook, or elsewhere down the line
+- study JavaScript Promises in general, to make sure you are returning data from the Promise correctly
+- make sure your `model` hook has a `return`
```
mike-north:
```suggestion
- make sure your `model` hook has a `return` statement
```

guides/release/routing/specifying-a-routes-model.md
mike-north: `{{debugger}}` is the very first thing I try in this situation.

mike-north: If you're going to mention the ember inspector, the "model" column in the "View tree" tab would be the "go-to" place for this. Only if I saw that the model was null/undefined/empty would I want to check out loaded e-d records

 
Screen Shot 2019-03-30 at 11 48 17 PM

mike-north:
Extracting ember-data out into an independent topic would be a huge positive change for the beginning of the learning experience. I hope we can continue moving in this direction, and ultimately land at a place where the first time readers experiment with Ember, it's in the simplest context possible, involving only the most important entities and concepts.

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
-Routes without dynamic segments will always execute the model hook.
+Behind the scenes, what is happening is that the route's controller receives the results of the model hook, and makes those results available to the template. Your app may not have a controller file for the route, but the behavior is the same regardless.
 
 ## Multiple Models
 
+What should you do if you need the `model` to return the results of multiple API requests?
+
 Multiple models can be returned through an
 [RSVP.hash](https://www.emberjs.com/api/ember/release/classes/rsvp/methods/hash?anchor=hash).
 The `RSVP.hash` method takes an object with promises or values as properties as an argument, and returns a single promise.
```
lisaychuang: I think this might be a typo: " as properties as an argument"? I don't quite understand what this sentence means... 

Based on the [API docs](https://api.emberjs.com/ember/release/classes/rsvp/methods/hash?anchor=hash), it sounds like the `hash` method takes an object (which can contain multiple promises), and returns a promise. The returned promise will resolve when all the promises listed IN the object are resolved, or rejected if ANY of them are rejected.

With the code example below, it may be helpful to show in comments what the hash will resolve to? Does `songs` resolves to an array of song records?

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
 
 If you use [Ember Data](../../models/) and you are building an `RSVP.hash` with the model's relationship, consider instead properly setting up your [relationships](../../models/relationships/) and letting Ember Data take care of loading them.
```
lisaychuang: If the goal is to focus on the `model` hook, maybe we can remove this sentence?  As a reader, I feel the need to understand how to use `RSVP.hash` with Ember Data after reading this. It also introduce questions about how Ember Data works, and what you mean by building a hash with "the model's relationship" -- even though there have been no mention of model relationship so far in this guide...  I assume that refers to things like `belongs to`  and `have many` ?

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+For example, if your app uses Ember Data to make API requests, you could include the ID in a `findRecord`:
+
+```javascript {data-filename=app/routes/photo.js}
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  model(params) {
+    this.store.findRecord('photo', params.id)
+  }
+});
```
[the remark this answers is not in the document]
lisaychuang: Referring to previous comments on line 112. I had suggested this order of examples (from broad to most specific):

- General example 
- Web API
- Ember-specific examples (e.g. ember-data / ember-ajax etc)

Perhaps use a general example here if possible?

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+```
+
+### Linking to a dynamic segment
+
+There are two ways to link to a dynamic segment from an `.hbs` template using the
+[link-to](../../templates/links/)
+helper.
+Depending on which approach you use, it will affect whether that route's model hook is run.
```
lisaychuang: Depending on which approach you use, it will affect whether that route's `model` hook is run.

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+To learn how to link to a dynamic segment from within the JavaScript file, see the API documentation on
+[`transitionTo`](https://api.emberjs.com/ember/3.8/classes/RouterService/methods/transitionTo?anchor=transitionTo)
+instead.
+
+When you provide a string or number to the `link-to`, the dynamic segment's model hook will run when the app transitions to the new route.
```
lisaychuang: the dynamic segment's `model` hook will run when the app

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+In this example, `photo.id` might have a value that's an ID like `4`:
```
lisaychuang: In this example, `photo.id` might have a value of `4`:

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+```
+
+Routes without dynamic segments will always execute the model hook.
+
 ## Reusing Route Context
 
 Sometimes you need to fetch a model, but your route doesn't have the parameters, because it's
 a child route and the route directly above or a few levels above has the parameters that your route
 needs.
+You might run into this if you have a URL like `/photos/4/comments/18`, and when you're in the comments route, you need a photo ID.
```
lisaychuang: when you're in the comments route, you need a photo `:id`

In this example, you are using the convention of `mode_id` instead of just plain `:id` ( i.e. `album_id`), this might be confusing for readers as the previous examples above are using `:id`?

Since the code example uses `song` and `album`, change the URL example to match that instead (e.g. `/albums/4/asongs/18`)

----
The section below was not edited so I can't add comments inline:

Firstly, talking to @mike-north he said that trying to get query params from the parents' route is an advanced concept, and an uncommon thing to do.  If we want to include the most important 20% of content that people need 80% of the time, this section does not pass the 80/20 rule. This looks like an an edge case and have "code smell".

Specific comments on why this section is confusing:

- [ ] Line 305 - 306:
> This is guaranteed to work because the parent route is loaded. But if you tried to do `paramsFor` on a sibling route, you wouldn't have the results you expected.

Q1:  Why do we have a guarantee with the parent route? Simply stating a guarantee is in one place but not the other doesn't help me form a mental model.
Q2: Why is using this method with a sibling route undesirable?

- [ ] Line 308
> This is a great way to use the parent context to load something that you want.
Q: What does the generic term "something that you want" mean?
In the example of `song` loading something from `album`, are we referring to data on the `album` e.g. album name? 

- [ ] Line 309 - 311
> Using `paramsFor` will also give you the query params defined on that route's controller.
This method could also be used to look up the current route's parameters from an action
or another method on the route, and in that case we have a shortcut: `this.paramsFor(this.routeName)`.

😩I have no idea what this paragraph means. 

- [ ] Line 313
> In our case, the parent route had already loaded its songs, so we would be writing unnecessary fetching logic.

There has been no explanation on the parent route and how it loads data so far, so how did we arrive at this conclusion that `songs` are already loaded?

- [ ] Line 329
> In the case above, the parent route looked something like this
Q1: Again, I'm still confused about what this example is
Q2: Why do we need to use `RSVP.hash` in the parent route?


I think either this entire section needs to be eliminated, or it need to be reviewed and updated to a sensible order:
- Start explaining from the "child route" (e.g. `song`) and move upwards (e.g. `album)
- Start explaining from the parent route and move downwards to child route

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -224,3 +343,14 @@ export default Route.extend({
 ```
 
 And calling `modelFor` returned the result of the `model` hook.
+
+## Debugging models
+
+If you are having trouble getting a model's data to show up in the template, here are some tips:
+
+- return static, sample data as a test to see if the problem is really in the model hook, or elsewhere down the line
```
lisaychuang: Referring to previous comments, can we change the phrase  _"static, sample data"_ to hard-coded data or something that's equally common phrased?

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -224,3 +343,14 @@ export default Route.extend({
...
+- study JavaScript Promises in general, to make sure you are returning data from the Promise correctly
+- make sure your `model` hook has a `return`
+- check to see whether the data returned from a `model` hook is an object, array, or JavaScript Primitive. For example, if the results of `model` are an array, using `{{this.model}}` in the template won't work. You will need to iterate over the results with an `{{#each}}` helper. If the results of `model` are a single object, you need to access the individual attributes like `{{this.model.title}}` to see anything rendered
```
lisaychuang: Mix of singular and plural in this sentence. The `model` hook returns one result, and it can be an object / array / string etc.  Suggested changes:

For example, if the **result** of `model` is an array, using `{{this.model}}` in the template won't work. You will need to iterate over the **array** with an `{{#each}}` helper. If the result **is an object**, you need to access the individual **attribute** like `{{this.model.title}}` to **render it in the template.**

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
-Typically, the `model` [hook](../../getting-started/core-concepts/#toc_hooks) should return an [Ember Data](../../models/) record,
-but it can also return any [promise](https://www.promisejs.org/) object (Ember Data records are promises/),
-or a plain JavaScript object or array.
-Ember will wait until the data finishes loading (until the promise is resolved/) before rendering the template.
+`model` hooks have some special powers:
 
-The route will then set the return value from the `model` hook as the `model` property of the controller.
-You will then be able to access the controller's `model` property in your template:
+1. When you return data from this model, it becomes automatically available in the route's `.hbs` file as `this.model`
+2. A `model` hook can return either a JavaScript [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises), or static sample data, like a string, object, or array
```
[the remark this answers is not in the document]
zachgarwood: I agree with @mike-north . Perhaps this could be restated as:
"A model hook can return just about any type of data, like a string, object, or array, but most the most common pattern is to return a Javascript Promise."

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
-```javascript {data-filename=app/routes/photo.js}
+Now let's look at [`ember-ajax`](https://github.com/ember-cli/ember-ajax), an different data-fetching library that you could install in your app.
+It's a convenient wrapper around `jQuery.ajax`, a popular general JavaScript library. Like Ember Data, it returns data from within a Promise:
+
+```javascript {data-filename=app/routes/favorite-posts.js}
 import Route from '@ember/routing/route';
+import { inject as service } from `@ember/service`
 
 export default Route.extend({
+  ajax: Ember.inject.service(),
```
zachgarwood:
```javascript
ajax: service(),
```

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,118 +11,128 @@ Router.map(function() {
...
-or a plain JavaScript object or array.
-Ember will wait until the data finishes loading (until the promise is resolved/) before rendering the template.
+`model` hooks have some special powers:
 
-The route will then set the return value from the `model` hook as the `model` property of the controller.
-You will then be able to access the controller's `model` property in your template:
+1. When you return data from this model, it becomes automatically available in the route's `.hbs` file as `this.model`
+2. A `model` hook can return either a JavaScript [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises), or static sample data, like a string, object, or array
+3. If you return a Promise from the model hook, your route will wait for the Promise to resolve before it renders the template
+4. `model` hooks work great with
```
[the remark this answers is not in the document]
jenweber: This is a good point, together with @mike-north's below. Revised to

"4. Since the `model` hook is Promise-aware, it great for making API requests (using tools like [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)) and returning the results. "

guides/release/routing/specifying-a-routes-model.md
jenweber: Agreed, revised

guides/release/routing/specifying-a-routes-model.md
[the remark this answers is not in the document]
jenweber: Replaced with ember-fetch.

guides/release/routing/specifying-a-routes-model.md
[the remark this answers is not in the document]
jenweber: Mike raised a good point. I'm removing this example altogether. I want to avoid introducing RSVP.

guides/release/routing/specifying-a-routes-model.md
jenweber: agreed, removed

guides/release/routing/specifying-a-routes-model.md
jenweber: I reworded this, based on the API docs phrasing.

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
 
 If you use [Ember Data](../../models/) and you are building an `RSVP.hash` with the model's relationship, consider instead properly setting up your [relationships](../../models/relationships/) and letting Ember Data take care of loading them.
 
+## Dynamic Models
```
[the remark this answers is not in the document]
jenweber: I removed my new example and instead referenced the example from the previous page. I think you are right about colliding ids.

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+But what if you need to request different data after user interaction?
+What if a specific post should load based on the URL that the user visited, or a selection that was made?
+In Ember, this is accomplished by defining routes with [dynamic
+segments](../defining-your-routes/#toc_dynamic-segments), or by using [query parameters](/query-params).
+This section will focus on dynamic segments, sometimes also referred to as "dynamic models or routes."
+
+### What does a dynamic segment look like?
+
+Here's what it would look like in the URL if a route has a dynamic segment: `http://localhost:4200/posts/42`. The number `42` is the "dynamic" part.
+A user can follow a link taking them to `http://localhost:4200/posts/3` and see post number 3 instead.
```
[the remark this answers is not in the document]
jenweber: I removed this section because it was redundant with the previous page

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+What's important here is that the data that should be displayed is tied to the url.
```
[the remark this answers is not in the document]
jenweber: I removed this section because it was redundant with the previous page

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+
+```bash
+ember generate route photo --path "photos/:id"
+```
+
+This command will create a `.js` file, an `.hbs` file, and add a route in `router.js`:
+
+```javascript {data-filename=app/router.js}
+Router.map(function() {
+  this.route('photo', { path: '/photos/:id' });
```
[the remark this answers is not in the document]
jenweber: I removed this section because it is redundant with the previous page

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -163,11 +175,118 @@ each record in the song model and album model:
...
+export default Route.extend({
+  model(params) {
+    console.log('This came from the URL: ', params.id)
+  }
+});
+```
+
+The `:id` name here isn't something special.
+Whatever name you give the dynamic segment, that's the key it will show up as in the `model` parameters.
+You could call it `:puppies` in `router.js` if you wanted to, and access it in the model as `params.puppies`.
```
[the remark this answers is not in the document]
jenweber: I'm not sure about the link-to magic, but I can remove it anyway for conciseness.

guides/release/routing/specifying-a-routes-model.md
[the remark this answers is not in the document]
jenweber: Removed this example altogether

guides/release/routing/specifying-a-routes-model.md
jenweber: This is covered in the previous page of the guides.

guides/release/routing/specifying-a-routes-model.md
jenweber: Wait, what? It automatically provides a model hook that retrieves data? 😱

guides/release/routing/specifying-a-routes-model.md
jenweber: I agree on a lot of these points. I didn't touch this section because it was super messy :P I would like to break this out into a new Issue/PR

guides/release/routing/specifying-a-routes-model.md
jenweber: This is interesting, I always ignored that column as gibberish.

jenweber: I added debugger and log as the first suggestion

jenweber:
@mike-north @lisaychuang @zachgarwood and @acorncom, thanks so much for the reviews! This is such an important part of understanding Ember, and I'm thankful we all invested the time to make it better :)

@acorncom will you do one more read through? I think I have addressed all the comments, except about how reusing a route's context is super confusing. I agree and would like to address it in a separate PR.

jenweber:
Also please squash & merge this when it's ready, or our git history will be ruined forever, haha

jenweber:
Wow something bad happened with the git diff... locally `git diff --name-only master` shows only two files changed. @acorncom any ideas?

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,122 +11,113 @@ Router.map(function() {
...
-Ember will wait until the data finishes loading (until the promise is resolved/) before rendering the template.
+`model` hooks have some special powers:
 
-The route will then set the return value from the `model` hook as the `model` property of the controller.
-You will then be able to access the controller's `model` property in your template:
+1. When you return data from this model, it becomes automatically available in the route's `.hbs` file as `this.model`
+2. A `model` hook can return just about any type of data, like a string, object, or array, but the most common pattern is to return a JavaScript [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises)
+3. If you return a Promise from the model hook, your route will wait for the Promise to resolve before it renders the template
+4. Since the `model` hook is Promise-aware, it great for making API requests (using tools like [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)) and returning the results. 
+5. When use the `model` hook to load data, you can take advantage of other niceties that Ember provides, like
```
mansona:
```suggestion
5. When using the `model` hook to load data, you can take advantage of other niceties that Ember provides, like 
```

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,122 +11,113 @@ Router.map(function() {
...
+`model` hooks have some special powers:
 
-The route will then set the return value from the `model` hook as the `model` property of the controller.
-You will then be able to access the controller's `model` property in your template:
+1. When you return data from this model, it becomes automatically available in the route's `.hbs` file as `this.model`
+2. A `model` hook can return just about any type of data, like a string, object, or array, but the most common pattern is to return a JavaScript [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises)
+3. If you return a Promise from the model hook, your route will wait for the Promise to resolve before it renders the template
+4. Since the `model` hook is Promise-aware, it great for making API requests (using tools like [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)) and returning the results. 
+5. When use the `model` hook to load data, you can take advantage of other niceties that Ember provides, like 
+[automatic route transitions](/preventing-and-retrying-transitions)
```
mansona:
```suggestion
[automatic route transitions](./preventing-and-retrying-transitions/)
```

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,122 +11,113 @@ Router.map(function() {
...
+after the data is returned,
+[loading screens, error handling](/loading-and-error-substates),
```
mansona:
```suggestion
[loading screens, error handling](./loading-and-error-substates/),
```

guides/release/routing/specifying-a-routes-model.md
mansona: I'm surprised this passed the link checker 🤔

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -9,122 +11,113 @@ Router.map(function() {
...
-      <img src="{{photo.thumbnailUrl}}" alt="{{photo.title}}" />
-    {{/link-to}}
-  </p>
-{{/each}}
-```
+Ember Data is a powerful (but optional) library included by default in new Ember apps. 
+In the next example, we will use Ember Data's [`findAll`](https://api.emberjs.com/ember-data/release/classes/DS.Store/methods/findAll?anchor=findAll) method, which returns a Promise, and resolves with an array of [Ember Data records](../../models/). 
 
-Routes without dynamic segments will always execute the model hook.
+_Note that Ember Data also has a feature called a [`Model`](https://api.emberjs.com/ember-data/release/classes/DS.Model), but it's a separate concept from a route's [`model`](https://api.emberjs.com/ember/release/classes/Route/methods/model?anchor=model) hook._
```
mansona:
```suggestion
_Note that Ember Data also has a feature called a [`Model`](https://api.emberjs.com/ember-data/release/classes/DS.Model), but it's a separate concept from a route's [`model()`](https://api.emberjs.com/ember/release/classes/Route/methods/model?anchor=model) hook._
```
I find it's easier to explain that it's a hook when you show that it's a function 👍 (this is a super optional suggestion)

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -161,13 +154,88 @@ each record in the song model and album model:
...
 ```
 
-If you use [Ember Data](../../models/) and you are building an `RSVP.hash` with the model's relationship, consider instead properly setting up your [relationships](../../models/relationships/) and letting Ember Data take care of loading them.
+## Dynamic Models
+
+In the examples above, we showed a route that will always return the same data, a collection of favorite posts. Even when the user leaves and re-enters the `/posts` route, they will see the same thing.
+But what if you need to request different data after user interaction?
+What if a specific post should load based on the URL that the user visited, like `posts/42`?
+In Ember, this can be accomplished by defining routes with [dynamic
+segments](../defining-your-routes/#toc_dynamic-segments), or by using [query parameters](/query-params), and then using the dynamic data to make requests.
```
mansona:
```suggestion
segments](../defining-your-routes/#toc_dynamic-segments), or by using [query parameters](./query-params), and then using the dynamic data to make requests.
```

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -161,13 +154,88 @@ each record in the song model and album model:
...
+
+```handlebars
+{{#each model as |photo|}}
+  {{#link-to "photo" photo}}
+    link text to display
+  {{/link-to}}
+{{/each}}
+```
+
+If you decide to pass the entire model, be sure to cover this behavior in your acceptance tests.
```
mansona: should we link to the testing section of the guides here?

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -161,13 +154,88 @@ each record in the song model and album model:
...
+
+If a route you are trying to link to has multiple dynamic segments, like `/photos/4/comments/18`, be sure to specify all the necessary information for each segment:
```
mansona: so this isn't 100% true 🤔 and to be honest I don't know how you would teach the alternative because it's quite a weird behaviour.

So first of all this is true if you're anywhere in the app **not** under the `photos` route. But if you're on `/photos/4` you can link to 

```handlebars
{{#link-to "photos.photo.comments.comment" 18}}
  link text to display
{{/link-to}}
```

and it will keep you within the first dynamic segment and use 18 for the second 🙈 kinda strange but c'est la vie 🤷‍♂

guides/release/routing/specifying-a-routes-model.md
mansona: yea this is something that I've known about for a long time but it's never been documented (that I know of) so it is completely alien to me 🙈 I have never relied on it once since forever

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -161,13 +154,88 @@ each record in the song model and album model:
...
+Routes without dynamic segments will always execute the model hook.
 
 ## Reusing Route Context
 
 Sometimes you need to fetch a model, but your route doesn't have the parameters, because it's
 a child route and the route directly above or a few levels above has the parameters that your route
 needs.
+You might run into this if you have a URL like `/photos/4/comments/18`, and when you're in the comments route, you need a photo ID.
 
 In this scenario, you can use the `paramsFor` method to get the parameters of a parent route.
```
mansona: I feel like we should use album_id **and** comment_id in this example to make it more realistic 🤔 also aren't you saying that we're in the comments route? shouldn't the file name for this be `app/routes/photo/comment.js` or something?

mansona: and to that point... if the code examples are talking about albums and songs shouldn't our in-prose example also be using the same model types?

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -224,3 +292,15 @@ export default Route.extend({
...
+
+## Debugging models
+
+If you are having trouble getting a model's data to show up in the template, here are some tips:
+
+- Use the [`{{debugger}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/debugger?anchor=debugger) or [`{{log}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/debugger?anchor=log) helper to inspect the `{{model}}` from the template
+- return hard-coded sample data as a test to see if the problem is really in the model hook, or elsewhere down the line
+- study JavaScript Promises in general, to make sure you are returning data from the Promise correctly
+- make sure your `model` hook has a `return` statement
+- check to see whether the data returned from a `model` hook is an object, array, or JavaScript Primitive. For example, if the result of `model` is an array, using `{{this.model}}` in the template won't work. You will need to iterate over the array with an `{{#each}}` helper. If the result is an object, you need to access the individual attribute like `{{this.model.title}}` to render it in the template.
```
mansona: should we link `{{#each}}` to the actual each helper?

guides/release/routing/specifying-a-routes-model.md
jenweber: yes, added

guides/release/routing/specifying-a-routes-model.md
jenweber: In the interest of teaching the happy path in the Guides, I think we should keep it as-is.

guides/release/routing/specifying-a-routes-model.md
jenweber: I agree, however I would like to address this in a future PR. This discrepancy is present in the current guides: https://guides.emberjs.com/release/routing/specifying-a-routes-model/

guides/release/routing/specifying-a-routes-model.md
jenweber: Added a note on this. Thanks for the info!

guides/release/routing/specifying-a-routes-model.md
jenweber: Yeah, I think we have plenty of model hook examples on this page, but it would make a good blog article or FAQ

guides/release/routing/specifying-a-routes-model.md
jenweber: yes, done

mansona:
🎉 congratulations @jenweber

jenweber:
Thanks everyone for the reviews! Merged.

PR #709, opened 2019-03-30
Updating upgrading editions.md

mdbiscan:
Updated the built-in components to use angle bracket syntax. See https://github.com/ember-learn/guides-source/issues/692

### DIFF
--- a/guides/release/upgrading/editions.md
+++ b/guides/release/upgrading/editions.md
@@ -1854,13 +1854,13 @@ bindings will not work:
   This does not work, since @value is
   an argument and is immutable
 -->
-<Input value=@value />
+<Input @value={{@value}} />
 
 <!--
   Instead, we should update the value
   by passing an _action_ to the component
 -->
-<Input value=@value key-up=@updateValue />
+<Input @value={{@value}} @key-up={{@updateValue}} />
 ```
 
 Additionally, the `mut` helper generally can't be used for the same reason:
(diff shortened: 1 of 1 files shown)


guides/release/upgrading/editions.md
```diff
@@ -1854,13 +1854,13 @@ bindings will not work:
   This does not work, since @value is
   an argument and is immutable
 -->
-{{input value=@value}}
+<Input value=@value />
```
mdbiscan: This doesn't seem right...

locks: Should be `<Input @value={{@value}} />`

locks:
```suggestion
<Input @value={{@value}} />
```

guides/release/upgrading/editions.md
```diff
@@ -1854,13 +1854,13 @@ bindings will not work:
...
 
 <!--
   Instead, we should update the value
   by passing an _action_ to the component
 -->
-{{input value=@value key-up=@updateValue}}
+<Input value=@value key-up=@updateValue />
```
locks:
```suggestion
<Input @value={{@value}} @key-up={{@updateValue}} />
```

mdbiscan:
> To make straight conversions, you need to prefix arguments with @, and then you need curlies on the right hand side, so something like <Input @value={{this.componentProperty}}.

I get that. I was taken aback on `@value={{@value}}`. Thanks for the protips!

locks:
Thanks for the help!

Issue #664, opened 2019-03-18
add CRUD to the glossary or remove the line


efx:
We mention `CRUD` in the [customizing serializers](https://github.com/ember-learn/guides-source/blame/master/guides/release/models/customizing-serializers.md#L766) page but it does not appear in the glossary:
> This method is called by `store.normalize(type, payload)` and is often
used for normalizing requests made outside of Ember Data because they
do not fall into the normal CRUD flow that the adapter provides

Given this is a guide, I think it makes sense to include a definition of that acronym or remove the problematic line

jenweber:
This is a good issue for new contributors! We should define in this section what CRUD stands for - Create, Read, Update, and Destroy. These are the changes that could be made to data on a server. Ember data adapters define what those requests to the server look like.

ClaytonTurner:
I'm interested in tackling this. 
Should these changes be made to all releases in guides/* or just to the release folder?

efx:
@ClaytonTurner great, thanks! For most new work I think we just change files in the [`guides/release`](https://github.com/ember-learn/guides-source/tree/master/guides/release) folder.

ClaytonTurner:
I believe this is ready for review.

PR #714, opened 2019-04-02
Add CRUD to Glossary (#664)

ClaytonTurner:
Fixes https://github.com/ember-learn/guides-source/issues/664.


ClaytonTurner:
I'm unsure of how my changes can be extending deploy lengths here. If anyone has insight into that, then that'd be helpful.

PR #717, opened 2019-04-03
refactor links inline

### LINKED ISSUE #330: Markdown link inconsistencies

maxwondercorn:
* Refactor links to inline, v3.x to release
* Updated CONTRIBUTING.md

PR refactors against current master

Resolves #330

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,28 +124,38 @@ When linking to API pages:
 (api link)
 * include parenthesis when linking to an API method, i.e., <code>&#96;store.push()&#96;</code>
 
+All linking should use inline-style links as shown below, not reference-style links
+
+```markdown
+[query params](../routing/query-params)
+```
+
+The following reference-style patterns should not be used:
+
+```markdown
+[query params][]
+[query params]: ../routing/query-params
+
+[query params][arbitrary text]
+[arbitrary text]: ../routing/query-params
+
+[query params][1]
+[1]: ../routing/query-params
+```
+
 ## Spellchecking/linting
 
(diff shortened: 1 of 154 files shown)


locks:
Master is the source of truth :)

MelSumner:
Thank you for doing this work! It looks like some updates will need to happen before we can merge, to resolve some conflicts. Once that is finished, I'm happy to merge this PR! :+1:

maxwondercorn:
@MelSumner conflicts fixed

locks:
Double checked!

maxwondercorn:
I see there are some non-inline links in 3.9 and 3.10. I will do another PR to fix those

locks:
Thanks for the help!

Issue #570, opened 2019-03-07
proposal: refactor $.getJSON examples to use fetch


efx:
Converting code examples for Octane has refreshed a discussion about using jQuery in trivial code examples. I think we should refactor code examples that feel dated due to using `$.getJSON`. See #33 for my prior hang ups. Tangential to https://github.com/ember-learn/guides-source/issues/499.
# current usage in release guides
* https://github.com/ember-learn/guides-source/blame/master/guides/release/models/index.md#L114
* https://github.com/ember-learn/guides-source/blame/master/guides/release/models/index.md#L143
* https://github.com/ember-learn/guides-source/blame/master/guides/release/models/finding-records.md#L93
* https://github.com/ember-learn/guides-source/blame/master/guides/release/routing/asynchronous-routing.md#L38

# why use window.fetch?
I propose we use [window.fetch](https://fetch.spec.whatwg.org/) for at least 3 reasons:
1. it is [widely supported](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#Browser_compatibility) by modern browsers.
2. In a land of jQuery free Ember applications, why should we promote depending on the entire jQuery library for a single function?
3. Easier teaching and mental model. By aligning with a web standard, we leverage existing teaching and developer experience. For new developers it gives them a helpful tool that could be considered a best practice in web development. Fetch's Promise based design flows nicely with concepts taught throughout the Ember guides and modern asynchronous web programming.

acorncom:
@efx I like the idea here. Given that the browser matrix for Ember still includes IE 11, it seems like it would be best to be suggesting using https://github.com/ember-cli/ember-fetch so we don't deliberately steer folks toward something that could cause them headaches in production. Thoughts?

efx:
@acorncom good point. I agree we should highlight that option with the reasoning for it so users can make informed decisions for their project.

PR #742, opened 2019-04-13
Replace ajax reference to fetch in release guide

dcyriller:
This relates to https://github.com/ember-cli/ember-cli/pull/8394


acorncom:
As we've been discussing in a few other places (https://github.com/ember-learn/guides-source/issues/570 and https://github.com/ember-learn/guides-source/pull/708#discussion_r270601273, switching to `fetch` usage without a polyfill such as `ember-fetch` means that we're encouraging usage that will break IE 11 folks. While that's something that is increasingly ok for apps depending on their users, I don't think we can yet encourage that as a global pattern while IE 11 is still part of the Ember framework support matrix.

@dcyriller any thoughts on how you'd update things here given that constraint? 😉

dcyriller:
There are roughly two options for the guides: either focus on completeness or on fundamentals.

We can go deep in the explanations around Ember (including how Ember polyfills for IE11)... Or we can show them how simple ember can be. :wink:

It's your call and I'll update the PR if you still want me to. But I wanted to let you know what I think 🙂. Let me know if my word makes you change your mind.

jenweber:
@dcyriller thanks for your work here! The first time ever in the guides that we are advising fetch is now merged, however it uses `ember-fetch`. If you update the code samples to show using `ember-fetch` imported, I think we can merge this PR too.

So many new devs copy & paste our code as it is written, that we want to play it safe. I had no idea myself that fetch was a problem until fairly recently, so I assume that other people may have the same issue. You can link to https://guides.emberjs.com/release/routing/specifying-a-routes-model/#toc_fetch-example if you want to point people towards an explanation of why we are using ember-fetch.

Does that sound ok?

jenweber:
P.S. I really like that you have made the prose more generalized. I think it really helps people who are just learning to figure out what is mentioned as examples vs what is a learning requirement.

MelSumner:
@jenweber @acorncom I wonder if this would be okay to merge, since ember-data now uses ember-fetch (which was a necessary part of removing the need for jQuery from Ember). What do you think?

dcyriller:
I've added fetch import in javascript files 👍

jenweber:
@MelSumner good point, that makes sense.

Thanks @dcyriller for your patience and keeping up with comments/edits! Let's merge 🎉

PR #743, opened 2019-04-15
Add note about the existing defined variable.

aditya-kapoor:
While following the listed example on the js-primer page in the guides, I felt that the variable `name` might conflict the existing `window.name` variable which is defined globally and the example might not work as expected because most of the people reading the guides would try it out in the Chrome or Firefox inspect (correct me if I am wrong on this.)


jenweber:
This is a good catch. I think it would be best if we just changed the variable name though, so we don't have to explain window context. What if we call it `username` instead in the example?

jenweber:
P.S. Thanks for stopping by to help fix this! So many people copy/paste the code examples and this will help them.

jenweber:
I added a commit to change `name` to `firstname` instead. Thanks for bringing this problem to light!

PR #752, opened 2019-04-18
Reference to ember data

dougbanville:
This example uses Ember data in an action without any reference to it, maybe an example using fetch also? There was also a syntax error

### DIFF
--- a/guides/release/routing/defining-your-routes.md
+++ b/guides/release/routing/defining-your-routes.md
@@ -323,7 +323,7 @@ Router.map(function() {
 In the above example we have successfully used a wildcard route to handle all routes not managed by our application
 so that when a user navigates to `/a/non-existent/path` they will be shown a message that says the page they're looking for wasn't found.
 
-Note that if you want to manually transition to this wildcard route, you need to pass an arbitrary (not empty) argument. For example:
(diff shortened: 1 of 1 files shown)


mansona:
Thanks for the contribution @dougbanville 🎉 

This issue is in relation to the first item on https://github.com/ember-learn/guides-source/issues/20

@jenweber do you think we should do a full-on "call to action" style warning about this code snippet like below: 

```
<div class="cta">
  <div class="cta-note">
    <div class="cta-note-body">
This example uses ember-data to fetch data but you can use other methods to access data in your route such as ember-fetch. For more information on ember-data you can read the <a href="/release/models/">ember-data section of the guides</a>.
      </div>
    </div>
  </div>
</div>
```

I'm not sure what the right next step with this is 🤔

jenweber:
I like the lightweight approach that @dougbanville has in this PR, instead of a callout. The only change I would suggest is a link to the Ember Data page of the guides, but this is mergeable as it is.

jenweber:
By the way, the rationale for a lighter-weight approach is that we could do this for every usage in the guides and it would be unobtrusive, but informative. Ember data is everywhere and if we did a ton of callouts, I think it would make things more confusing, not less.

PR #766, opened 2019-04-25
Remove extra semicolon

mikias10:
Cleanup.


jenweber:
Good catch, thank you!

PR #770, opened 2019-04-26
Fixes run method invocation in the example code

RichardOtvos:
Since we import run at the top (as we should), it should just say `run()` instead of `Ember.run()`


pzuraq:
I actually don't think this is necessary anymore on the latest Ember's, cc @rwjblue. We now always autorun, so it _should_ work I think

locks:
Or maybe use the click from the test helper package?

rwjblue:
@pzuraq 

>I actually don't think this is necessary anymore on the latest Ember's, cc @rwjblue. We now always autorun, so it should work I think

This guide is for Ember 2.18, which still has the autorun issue to contend with.

@locks 

> Or maybe use the click from the test helper package?

We didn't switch to the "new" testing style until 3.x, this guide is for 2.18 and doesn't use `@ember/test-helpers`.

RichardOtvos:
^ What @rwjblue said.

This is the guide for v2.18.x, and IMHO we don't wan to change the guide or how we teach testing, especially not retroactively. 

This PR is only about fixing a typo / syntax error, so if someone copies the example code and tries to run it, they can without getting an error

locks:
Thanks for the fix!

PR #771, opened 2019-04-27
Update SuperRentals tutorial - Service section

greyhwndz:
When I was going thru the tutorial, I was using windows. 
I would like to propose to include the windows-specific terminal command to set an environment variable for the leaflet maps api key.


toddjordan:
Thanks for the PR!  Updates to make things easier for windows users are great and always welcome.

Issue #669, opened 2019-03-18
Quest - Changes we need to make before the official release of Octane Guides


jenweber:
The Learning team needs your help getting Octane ready to ship on emberjs.com! To help out, please do your work in the `guides/release` folder of the `octane` branch, and open PRs back into `octane`.

This issue described is MVP to merge the guides for 3.14. You can see the current Octane preview guides at https://octane-guides-preview.emberjs.com/release/

### URL persistence
- [x] writing redirects from old pages to new, #719  PR #1210 

### Version upgrade guide
- [x] Rework the conversion guide to be readable/useful in both directions (to an from Octane)
- [x] Revise the Upgrade Guide intro to be short & sweet - has PR by JW https://github.com/ember-learn/guides-source/pull/796
- [x] Link to examples in the upgrade guides, instead of using new examples (for example, show "old version" of syntax and link to the "new version"

### Syntax updates
- [x] document `{{on}}` (see https://github.com/emberjs/rfcs/blob/master/text/0471-on-modifier.md) - @sivakumar-kailasam is doing this
- [x] document `{{fn}}`(see https://github.com/emberjs/rfcs/blob/master/text/0470-fn-helper.md) @sivakumar-kailasam is doing this
- [x] show angle brackets versions of LinkTo and Input (refer to them as "built in components" - done by https://github.com/ember-learn/guides-source/pull/696 and JW #797 
- [x] search for computed properties and replace them with tracked. (There should only be 1 section where we talk about interop between tracked and CPs.) JW partially addressed in #805 
    - [x] Make sure that we have a section on interop, and write it if needed
- [ ] Make sure that Ember Data code samples actually work
- [ ] make sure tests work

### Octane tutorial

This is involved, probably 20 hours of work that will need to be divided up.

- [x] Bring the Tutorial prose up to speed to use Glimmer Components, Tracked, etc. Goes with the next item below.
- [x] Make a version of Super Rentals that uses Octane - WIP by VJ - see https://github.com/ember-learn/super-rentals/issues/118)

### Quick fixes
- [x] Search for uses of the word "Glimmer" and remove them unless they are referencing the package name. In one or two places, explain that glimmer is the rendering engine and you don't need to learn the library itself directly - has PR by JW  https://github.com/ember-learn/guides-source/pull/796
- [x] In the section about the Router, add a note that it's an Ember Object and not a class (note by JW - we don't even show the full router boilerplate and people only edit the map, so this is ok as-written in the Guides)
[...]

PR #805, opened 2019-05-21
Octane - Replace some references to Computed Properties with Tracked

jenweber:
This is incremental work towards #669

### DIFF
--- a/guides/release/applications/dependency-injection.md
+++ b/guides/release/applications/dependency-injection.md
@@ -234,16 +234,18 @@ For example, this component plays songs with different audio services based
 on a song's `audioType`.
 
 ```javascript {data-filename=app/components/play-audio.js}
-import Component from '@ember/component';
-import { computed } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
 import { getOwner } from '@ember/application';
 
 // Usage:
 //
-// {{play-audio song=song}}
+// <PlayAudio @song=this.song />
 //
-export default Component.extend({
-  audioService: computed('song.audioType', function() {
+export default class PlayAudioComponent extends Component {
+  @tracked song;
+
+  get audioService() {
     if (!this.song) {
       return null;
     }
@@ -252,11 +254,11 @@ export default Component.extend({
     let audioType = this.song.audioType;
 
     return applicationInstance.lookup(`service:audio-${audioType}`);
-  }),
+  };
 
   click() {
     let player = this.audioService;
     player.play(this.song.file);
-  }
-});
+  };
+};
 ```
--- a/guides/release/applications/dependency-injection.md
+++ b/guides/release/applications/dependency-injection.md
@@ -235,7 +235,6 @@ on a song's `audioType`.
 
 ```javascript {data-filename=app/components/play-audio.js}
 import Component from '@glimmer/component';
-import { tracked } from '@glimmer/tracking';
 import { getOwner } from '@ember/application';
 
 // Usage:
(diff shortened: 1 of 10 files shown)


guides/release/applications/dependency-injection.md
```diff
@@ -234,16 +234,18 @@ For example, this component plays songs with different audio services based
...
 
 // Usage:
 //
-// {{play-audio song=song}}
+// <PlayAudio @song=this.song />
 //
-export default Component.extend({
-  audioService: computed('song.audioType', function() {
+export default class PlayAudioComponent extends Component {
+  @tracked song;
```
rwjblue: I wouldn't use `@tracked` at all here (once migrated to `@glimmer/component`), I'd probably do:

```js
export default class PlayAudioComponent extends Component {
  get audioService() {
    if (!this.args.song) {
      return null;
    }

    let applicationInstance = getOwner(this);
    let { audioType } = this.args.song;

    return applicationInstance.lookup(`service:audio-${audioType}`);
  }

  click() {
    let player = this.audioService;
    player.play(this.args.song.file);
  }
}
```

guides/release/applications/dependency-injection.md
```diff
@@ -252,11 +254,11 @@ export default Component.extend({
     let audioType = this.song.audioType;
 
     return applicationInstance.lookup(`service:audio-${audioType}`);
-  }),
+  };
```
rwjblue: I find it fairly odd to have semicolons here (I had never seen this before in class bodies until recently), I would normally not put anything here (no semicolon, no comma, just a new line).

I don't think this is a giant issue, but we do need to ensure that the guides consistently follow one pattern or the other (either semicolons or not).

guides/release/controllers/index.md
```diff
@@ -103,5 +103,5 @@ Yes! Controllers are still an integral part of an Ember application architecture
 #### When should we create a Controller?
 
 * We want to pass down actions or variables to share with a Route’s child components
-* We have a computed property that depends on the results of the model hook
+* We have a getter that needs the results of the model hook
```
rwjblue: I think that just saying "getter" here is a tad odd, we **are** computing a value (just not a `Ember.computed`), what do yo uthink about:

```suggestion
* We need to compute a value based on the results of the model hook
```

guides/release/models/relationships.md
```diff
@@ -267,7 +267,7 @@ updated and saved this often results in the creation of a great deal
 of code for very little benefit. An alternate approach is to define
 these relationships using an attribute with no transform
 (`attr()`). This makes it easy to access readonly values in
-computed properties and templates without the overhead of defining
+getters and templates without the overhead of defining
```
rwjblue:
```suggestion
other objects and templates without the overhead of defining
```

guides/release/routing/query-params.md
```diff
@@ -41,15 +41,14 @@ and the `category` property on `controller:articles`. In other words,
 once the `articles` route has been entered, any changes to the
 `category` query param in the URL will update the `category` property
 on `controller:articles`, and vice versa.
-Note that you can't bind `queryParams` to computed properties, they
+Note that you can't make `queryParams` be a getter; they
```
rwjblue:
```suggestion
Note that you can't make `queryParams` be a dynamically generated property (neither computed property, nor property getter); they
```

guides/release/applications/dependency-injection.md
jenweber: We're using a mix everywhere now :( I'll make a note to catch it in an editing pass.

PR #812, opened 2019-05-24
Update ember-data facilitating understanding.md

maurodibert:
I'm really new to Ember but think this kind of small details could help to retain concepts.

I'm not really sure if D.S. means that (LOL), but searched in the guides and I didn't find it, and since every time the concept appears, it's the other way around, as "structure of data".

This time could be a silly detail but if it helps to understand and also reduce the time of understanding, could be useful.

### DIFF
--- a/guides/release/tutorial/ember-data.md
+++ b/guides/release/tutorial/ember-data.md
@@ -3,7 +3,7 @@ As our application grows, we will want to persist our rental data on a server, a
 
 Ember comes with a data management library called [Ember Data](https://github.com/emberjs/data) to help deal with persistent application data.
 
-Ember Data requires you to define the structure of the data you wish to provide to your application by extending [`DS.Model`](https://www.emberjs.com/api/ember-data/release/classes/DS.Model).
+Ember Data requires you to define the structure of the data you wish to provide to your application by extending [`DS.Model`](https://www.emberjs.com/api/ember-data/release/classes/DS.Model).("D" for Data, "S" for structure).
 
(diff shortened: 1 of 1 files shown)


guides/release/tutorial/ember-data.md
```diff
@@ -3,7 +3,7 @@ As our application grows, we will want to persist our rental data on a server, a
 
 Ember comes with a data management library called [Ember Data](https://github.com/emberjs/data) to help deal with persistent application data.
 
-Ember Data requires you to define the structure of the data you wish to provide to your application by extending [`DS.Model`](https://www.emberjs.com/api/ember-data/release/classes/DS.Model).
+Ember Data requires you to define the structure of the data you wish to provide to your application by extending [`DS.Model`](https://www.emberjs.com/api/ember-data/release/classes/DS.Model).("D" for Data, "S" for structure).
```
locks: I believe it comes from "Data Store", but really it does not have any actual meaning nowadays, it is the namespace for Ember Data classes.
This is [going to change soon](https://github.com/ember-learn/guides-source/pull/814/files) and the old-school namespace won't be needed anymore! :)

maurodibert: @locks Ricardo! Cool then! And really thanks for the explanation! Should I close the branch right?

locks: Yeah, if you don't mind!

maurodibert: At all! ;-)

PR #814, opened 2019-05-25
Update ember data guides to use @ember-data packages

bmac:


jenweber:
Thank you so much @bmac!

Notes to maintainers - the gameplan here is that for 3.11, we cut a release of the Guides, and then merge this in immediately so it shows up for 3.11 users but not 3.10. I added a label and an issue to help remind us https://github.com/ember-learn/guides-source/issues/825

locks:
@jenweber we can do both in this PR because we need a commit creating the 3.10 folder, right?

locks:
Rebased to resolve conflicts.

jenweber:
We can update, re-review, and merge, when the generators match the style of this PR.  The generators currently produce:

```
import DS from 'ember-data';
const { Model } = DS;

export default Model.extend({

});
```

I'm marking this PR as WIP. Please let us know when we should take another look!

jamescdavis:
The blueprints were updated to use package imports in https://github.com/emberjs/data/pull/6358, which was released in ember-data@3.14.0

PR #815, opened 2019-05-25
Add example to "handling-events.md"

konstankinollc:
To be crystal clear, I have added an example.

### DIFF
--- a/guides/release/components/handling-events.md
+++ b/guides/release/components/handling-events.md
@@ -122,8 +122,19 @@ To utilize an `event` object as a function parameter:
 - Define the event handler in the component (which is designed to receive the
   browser event object).
 - Or, assign an action to an inline event handler in the template (which
-  creates a closure action and does receive the event object as an argument).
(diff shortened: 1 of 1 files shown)


locks:
Thanks for the contribution @konstankinollc, but isn't this code basically the same as the example right before it?

konstankinollc:
Yes, the code is the same. The point was to show an example of what the paragraph above is saying. For beginners that might be not clear.

konstankinollc:
I realize your concern in the code duplication. I could use #anchors instead to point user onto the code instead. Let me know what do you think.

jenweber:
A link back to the other example would be super!

jenweber:
Hey @konstankinollc, just following up on this PR. Is it something you still want to work on? Let us know if you have any questions! Thanks!

jenweber:
I am closing this PR due to inactivity, however if you would like to get this into the Guides, please feel free to open a new PR. Thanks!

PR #816, opened 2019-05-25
Update triggering-changes-with-actions.md

konstankinollc:


locks:
Thanks for the suggestion! 

I have a couple remarks about your additions.
The "Data Down Actions Up" pattern should be thought as a single unit, and breaking it apart like you suggest does not adequately communicate how DDAU helps you design the data flow in your application.
Related to that point, mentioning the parts of the pattern without presenting DDAU is to me a disservice to the reader, and they are left to find out for themselves what it is about.

I will close this PR for the reasons above. If you are interested in writing an explainer on DDAU, contact the Learning team on Discord and we might figure out the best path :)

PR #840, opened 2019-06-04
Fix output diff for "Manipulating Attributes" example

kevinansfield:
The output diff code had failed to show that the attribute is changed, instead only showing the text changing as in a previous example. Updated the code to show both attribute and text changing with appropriate diff markup.


jenweber:
@kevinansfield thanks for this fix! If you rebase on `octane`, you should see the test failures go away. They are unrelated to your changes. Then we can merge!

kevinansfield:
@jenweber rebased and tests are 💚

jenweber:
Merged, thank you!

PR #841, opened 2019-06-04
octane/glimmer-components-dom: Fix broken `Modal` template

Turbo87:
The `{{#if this.showingHelp}}` condition is used in the following template and is not needed inside of the `Modal` component itself.


jenweber:
@Turbo87 if you rebase on `octane`, the test failures should go away. Thanks for the correction!

Turbo87:
@jenweber this PR is rebased but apparently Percy is not happy about something

jenweber:
Merged, thanks!

PR #874, opened 2019-06-18
Two small typos.

muziejus:
One is of a semicolon where a colon should be used ({ inverse: true }), and
the other imports from @ember/component despite the fact that all of the
other component section pages import from @glimmer/component.


locks:
Appreciate you taking the time to send these correction in, keep'em coming :)

PR #875, opened 2019-06-18
Update text for acceptance-test

newyork-anthonyng:
The original text implies that there can only be one assert in the test.


locks:
Thanks!

PR #880, opened 2019-06-21
Update angle bracket syntax for nested components

### LINKED ISSUE #879: Angle bracket syntax guide incorrectly says it does not support nested components
https://guides.emberjs.com/release/reference/syntax-conversion-guide/#toc_when-to-use-classic-invocation-syntax

In Ember 3.10 support for nested angle bracket components was added, but the guides still say that it is not supported and does not show an example of how to use them.

kmiyashiro:
Support for nested angle brackets was added in 3.10

fixes #879


kmiyashiro:
@GavinJoyce Looks like the percy check needs approval as well, do you have a chance to take a look?

PR #890, opened 2019-06-24
[REFACTOR] Remove depreaction deprecate-router-events

billybonks:
was getting this deprecation while looking into something broccoli related. thought i would offer a fix. here is a reference that you lovely people maintain in some way :) https://deprecations.emberjs.com/v3.x/#toc_deprecate-router-events

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -12,9 +12,11 @@ const Router = EmberRouter.extend({
   metrics: service(),
   fastboot: service(),
 
-  didTransition() {
+  init() {
     this._super(...arguments);
-    this._trackPage();
+    this.on('routeDidChange', () => {
+      this._trackPage();
+    })
   },
 
   _trackPage() {


mansona:
@locks you tried this before right and it caused an infinite loop in the Ember code right? Was that ever fixed?

locks:
@mansona injecting the `router` service is what causes the loop, this sound be fine!

mansona:
oh... why aren't we suggesting this in the deprecation guide then? 🤔

locks:
@mansona OPs link is suggesting exactly this? I'm conffussee

billybonks:
I guess the question is under what conditions does injecting the router cause an infinite loop. Because from what I know that shout be okay.

locks:
@billybonks if you inject the router service into `EmberRouter`, because they depend on each other :x

PR #891, opened 2019-06-24
Octane - new content to help avoid jQuery confusion

jenweber:
## Purpose

This PR helps address https://github.com/emberjs/rfc-tracking/issues/3

The goal is to reflect that Octane will not have jQuery by default, and to give specific pointers to people about how to make API requests.

## Added

- A new top-level topic and Guide that lays out options for loading data in an Ember app :fire:. This is a continuation of the effort to untangle how we teach Ember Data. This section will have more articles as time goes on and we get to the point where Ember Data can be a stand-alone guide.

## Changed

- I modified the "optional features" guide to reflect my understanding of the way jQuery will work in Octane apps
- I removed a lingering jQuery reference

## Departures from [RFC 431](https://github.com/emberjs/rfcs/blob/master/text/0431-guides-restructure.md)

I renamed "data management" to be "Making back end requests" because the "state management" topic clashes with the original name. I reviewed state management and I think it's too confusing to try and add this data fetching stuff to that guide. I expect it to keep growing.

I am open to renaming this section.

## Next steps:
- another PR that finds the places where we should link to these pages
- In places where we mention a fetch polyfill, instead move that info to this page and just link to this resource


guides/release/back-end/index.md
```diff
@@ -0,0 +1,74 @@
...
+## How to make API requests
+
+Some common tools for making [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) (create, read, update, delete) requests in Ember include:
+
+- [Ember Data](../models/) is the official data persistence library for Ember. It has a powerful set of tools
+for formatting requests, normalizing responses, and efficiently
+managing a local cache of data. Install `@ember/data` to use it.
+It is included by default in new Ember apps.
+- Native JavaScript methods like [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). Install [`ember-fetch`](https://github.com/ember-cli/ember-fetch) in order to provide support for older browsers, and `import fetch from 'fetch'` to use it.
+- [jQuery Ajax](https://api.jquery.com/jquery.ajax/) requests. Install `@ember/jquery` and set `"jquery-integration": false` in your app's [`optional-features.json`](../configuring-ember/optional-features/) in order to be able to `import jQuery from 'jquery'` in your app.
```
jelhan: Shouldn't it be: _set `"jquery-integration": true` in your app's `optional-features.json`_?

In optional features guide it tells to use Ember CLI to configure a feature. Could be confusing if it recommends one time editing a file directly and another time to use a command.

pzuraq: It seems like this is what's recommended later on as well, so it may make sense to update these recommendations to be in sync

jenweber: Agreed, updated to just link to the guide on it

jenweber:
Thanks for the feedback @jelhan and @pzuraq! I updated this PR to include it.

PR #898, opened 2019-06-28
Update ember-data.md

Prady220:


Prady220:
Please add information about the new line of code. Not intuitive

jenweber:
Thank you for this update! I confirmed that this matches our example code https://github.com/ember-learn/super-rentals/blob/master/app/models/rental.js

PR #910, opened 2019-07-02
fix templating built-in-helpers examples

tzellman:
component template usages were missing the @ symbol, and missing a reference to the argument name


jenweber:
Good catch, thank you @tzellman! Merged.

PR #912, opened 2019-07-04
fn & on

NullVoxPopuli:
related info: https://www.pzuraq.com/ember-octane-update-action/


mixonic:
@NullVoxPopuli `@bound` is merely a proposal afaik, I'm not even sure if it is even a concrete proposal to TC39. I think the guides should stay focused on how to get thing done, and avoid contributing to FUD about what the future will bring.

NullVoxPopuli:
Gotchya.

yeah, after reading the propoal for decorators closer, it looks like bound would be something ember may provide, given this implementation example from the decorators proposal:
```ts
export decorator @bound {
  @initialize((instance, name) => instance[name] = instance[name].bind(instance))
}
```
These are the only decorators being proposed as part of the proposal:
image

guides/release/components/actions-and-events.md
```diff
@@ -297,7 +297,7 @@ import { action } from '@ember/object';
 
 export default class SendMessage extends Component {
   @action
-  sendMessage(messageType) {
+  async sendMessage(messageType) {
```
mixonic: I know some work supporting `async`/`await` landed into backburner and Ember in 3.x, but I'm not sure if FW core is encouraging this in app code yet. I've queried a few people for clarification.

Regardless, I'm not sure why this action is `async`... Simply to provide an example that it is possible?

guides/release/components/actions-and-events.md
```diff
@@ -392,11 +391,24 @@ block within the `"confirmDialog"` element:
 ```handlebars {data-filename=app/templates/components/button-with-confirmation.hbs}
 <button type="button" {{action this.launchConfirmDialog}}>{{this.text}}</button>
```
mixonic: This should also have been `on`-ified?

guides/release/components/actions-and-events.md
```diff
@@ -568,19 +588,21 @@ Note that `deleteCurrentUser` is now prepended with `@` as opposed to `this.`
...
+## String Action Notation and the `{{action}}` helper
 
 Historically the `{{action}}` helper (`<MyComponent @onClick=(action 'clickedButton')>`)
 and element modifier (`<button {{action 'clickedButton'}}>`) have accepted a
 string as the first argument. This form is no longer recommended, but might be
 seen in the wild when working on older Ember apps or addons.
 
 When you encounter a string based action it should be refactored to use the
-`@action` decorator (refactor away from the `actions` hash).
+`@action` decorator (refactor away from the `actions` hash) along with the `fn` helper for any partial application needed..
```
mixonic:
```suggestion
`@action` decorator (refactor away from the `actions` hash) along with the `fn` helper for any partial application needed.
```

guides/release/templates/actions.md
```diff
@@ -114,15 +124,23 @@ export default class Post extends Component {
...
+`createPost`. The form submission event needs to have `preventDefault` 
+called on it in order to opt out of the browser-default behavior -- which, 
+for forms, is to POST to the current URL.
+
 ## Attaching Actions to Non-Clickable Elements
 
-Note that while Ember currently permits you to add an action to any DOM element, not all DOM elements are eligible to receive focus, according to HTML standards.
+Note that while Ember currently permits you to add an action to any DOM 
+element, not all DOM elements are eligible to receive focus, according to 
+HTML standards.
```
mixonic: This `Note` is pretty confusing. As a reader I'm not sure what this comment (something something focus) has to do with actions on non-clickable elements.

This is likely trying to nudge the user to think about accessibility? Can it be structured so that a first time using coming to this section can get the information they expected before that consideration distracts them?

guides/release/components/actions-and-events.md
NullVoxPopuli: I've been using async/await _everywhere_ in emberclear, and haven't ran into issues. :man_shrugging: 

I think the the name implies that there is going to be some interaction with a backend, so I would expect it to be async :)

guides/release/components/actions-and-events.md
NullVoxPopuli: yes!

guides/release/templates/actions.md
NullVoxPopuli: > This Note is pretty confusing.

yeah, I struggled with it a bit, too. 

> This is likely trying to nudge the user to think about accessibility?

I think so. I've primarily only used Chrome and FireFox, so I haven't run in to any issues with browsers not *letting* things be clickable.  I wonder if this text is counting a screen reader or other a11y tool as a browser.

CvX:
Could/should the docs provide a migration guidance for the `value="target.value"` pattern?

For example, given this code:

```hbs
<input onclick={{action "update" value="target.checked"}} type="checkbox">
<input oninput={{action "update" value="target.value"}}>
```

```js
actions: {
  update(value) {
    console.log(value);
  }
}
```

Is this the recommended replacement?

```hbs
<input {{on "click" this.update}} type="checkbox">
<input {{on "input" this.update}}>
```

```js
@action
update({ target: { checked, value } }) {
  console.log(checked || value);
}
```

And is there a path for `(mut)` and `value` combo? 🙂 For example:

```hbs
<input oninput={{action (mut this.query) value="target.value"}}>
```

You'd have to create an action in the component that updates `this.query`, right?

AFAICT, the [{{on}} RFC](https://github.com/emberjs/rfcs/pull/471) didn't touch on that? @pzuraq

pzuraq:
@CvX I believe the first few examples should be the preferred form. Currently, we don't have a path to a direct replacement for `mut`/`value`, I agree this is a bit regrettable but we did explore this significantly during the design portion of the `on` and `fn` RFCs. The issue is that the current API is very counterintuitive:

* `mut` does nothing directly on its own, it returns a _function that sets a value_. You still have to _apply_ that function. This means that you have to know how the details of the surrounding environment work, which takes us to `action` and `value`
* `value` is a really weird API, basically allowing you to reach into an event and grab a nested path to pass to the next thing. You need to know that the path will be followed on the _event_ object itself, which is something that people trip up over frequently. It also does nothing valuable by itself - why wouldn't you just want the event in an event handler? It only makes sense in the context of the `mut` helper.

Basically, these APIs are convenient in that they are _terse_, but they are not easy to teach, and do not have much utility unless they are used together. We realized through the design process that what we really want here is some form of _closures_:

```hbs
<input {{on "input" (|event| (set this.query event.target.value))}} />
``` 

Something along these lines. However, this is a _huge_ lift, and adds a large amount of complexity to both Glimmer from an implementation perspective and templates from a language design persective, so I'm not sure if this is a direction we want to go yet.

I think in the meantime, it would be best to make a custom helper that handles common use cases like this, where it would be inconvenient to create an action handler. call it `pickAndSet`:

```js
<input {{on "input" (pickAndSet this "query" "target.value")}} />
```

This would pick the value off of whatever is passed to it and set it. Then at least the two actions are related to each other lexically, and can tell the users that they are related just by reading it. You could use a template transform to mimic the `mut` helper if that would be clearer:

```
<input {{on "input" (pickAndSet this.query "target.value")}} />
```

guides/release/templates/actions.md
```diff
@@ -22,21 +22,27 @@ export default class Post extends Component {
 ```
 
 You can then add this action to an element using the
-[`{{action}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/action?anchor=action)
+[`{{on}} (link tbd)`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/action?anchor=action)
```
muziejus: This may not be specific to this PR, but this seems like the best place to bring it up…

There's a conceptual leap that happens here, imo, that has been historically obscured by the duplication of the word `action`. Switching to `{{on}}`, thankfully, brings it to light. Namely, why does `toggleBody` need to be decorated with `@action`? What does that decoration actually _do_? For example:

```javascript
export default class MyButtonComponent extends Component {
  fire() {
    alert("fire!");
  }
}
```
and

```handlebars
<button {{on "click" this.fire}}>
  Click me!
</button>
```
works as expected—the `{{on}}` helper is enough to create interactivity in this example, without having to use `@action`.

NullVoxPopuli: ah, yeah. `@action` is needed when you want `this` to be bound -- I'll be sure that is mentioned somewhere

muziejus: yes. I think I understand this now!

muziejus: (but because of your and @locks interventions! which means I think I can try to help clarify the guide)

muziejus: tl;dr: binding is very confusing to me.

NullVoxPopuli: I'll try to find an article that can explain it nicely. It's pretty fundamental to javascript, so It's pretty important we find a good and clear explanation

NullVoxPopuli: @muziejus, thoughts on this article: https://yehudakatz.com/2011/08/11/understanding-javascript-function-invocation-and-this/

muziejus: Per our chat on Discord, this is the line that trips me up:

> Actions are methods that have been decorated with the @action decorator in the context of the template:

I would propose something like this (my terminology might be inexact):

> Actions are methods that have been decorated with the @action decorator in the context of the template. The decorator binds `this` to the `Post` component:

Then, perhaps, after the code, where the PR has:

> You can then add this action to an element using the [`{{on}} (link tbd)`]

Instead it can be:

> In binding the context, `@action` lets `toggleBody()` access the component's other properties when invoked from the template with `this`. Without the decorator, `this.isShowingBody` in `toggleBody()` is undefined, as in that case `this` refers not to the `Post` component, but to the `window` object. 

> You can trigger the method `toggleBody()` as a callback on a DOM event using the [`{{on}} (link tbd)`]...



This is rather verbose, but, to me at least, actually explains *why*, within an *Ember* context, you need to use the decorator.

NullVoxPopuli: I've updated the text for this, based on your suggestions, feel free to comment on the new stuff :)

guides/release/templates/actions.md
```diff
@@ -20,9 +20,10 @@ export default class Post extends Component {
   }
 }
 ```
+`@action` lets `toggleBody()` access the component's other properties when invoked from the template with `this`. Without the decorator, `this.isShowingBody` in `toggleBody()` is undefined. For more information on `this` [Understanding Javascript Function Invocation and this](https://yehudakatz.com/2011/08/11/understanding-javascript-function-invocation-and-this/)
```
muziejus: Thanks! My one suggestion is that the first sentence suggests that we use `this` to invoke the method. Perhaps this is syntactically clearer:

>`@action` lets `toggleBody()` access the component's other properties with `this` when it is invoked from the template.

locks: I think we should be clear that `@action` binds `this` to the instance of the class. And in turn, you can access component properties because of that.

muziejus: I like the use of the technical description (the keyword "binds") along with the lay description of what the result is.

NullVoxPopuli: Adjusted! :)

guides/release/templates/actions.md
NullVoxPopuli: Not sure what to do about these notes -- kinda out of scope for this Pr, but if someone else wants to tackle it / open an issue for it, be my guest!

guides/release/templates/actions.md
```diff
@@ -21,22 +21,30 @@ export default class Post extends Component {
 }
 ```
 
-You can then add this action to an element using the
-[`{{action}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/action?anchor=action)
+`@action` binds the `this` of `toggleBody()` to the instance of the class, allowing access the component's other properties when invoked. Without the decorator, `this.isShowingBody` in `toggleBody()` is undefined. For more information on `this` [Understanding Javascript Function Invocation and this](https://yehudakatz.com/2011/08/11/understanding-javascript-function-invocation-and-this/)
```
muziejus: Thanks @NullVoxPopuli and @locks for talking this through with me. I think this hits the key terms (`binds`, `class`) while also, with the second sentence, explaining in a more quotidian way why `@action` is needed!

guides/release/state-management/tracked-properties.md
```diff
@@ -255,7 +255,7 @@ export default class ApplicationController extends Controller {
 ```js {data-filename=app/templates/application.hbs}
 {{this.model.fullName}}
 
-<button {{action this.updateName 'Krati' 'Ahuja'}}>
+<button {{on (fn this.updateName 'Krati' 'Ahuja')}}>
```
lupestro: `<button {{on "click" (fn this.updateName 'Krati' 'Ahuja')}}>`

guides/release/tutorial/service.md
```diff
@@ -169,7 +169,7 @@ Finally open the template file for our `rental-listing` component and add the ne
 <article class="listing">
   <a
     class="image {{if this.isWide "wide"}}"
-    onclick={{action "toggleImageSize"}}
+    {{on "click" toggleImageSize}}
```
lupestro: {{on "click" this.toggleImageSize}}

guides/release/state-management/tracked-properties.md
NullVoxPopuli: gasp! thanks!

guides/release/tutorial/service.md
NullVoxPopuli: fixed!

guides/release/state-management/tracked-properties.md
NullVoxPopuli: fixed!

guides/release/components/index.md
```diff
@@ -142,4 +142,5 @@ In the following guides we'll talk about:
   which Ember fully supports
 
 - **Contextual Components**, which can be used dynamically to pass components
-  around as values, and allow them to be invoked in different locations.
\ No newline at end of file
+  around as values, and allow them to be invoked in different locations.
```
muziejus: that comma before "and" is not needed and causes confusion.

guides/release/components/index.md
```diff
@@ -142,4 +142,5 @@ In the following guides we'll talk about:
   which Ember fully supports
```
muziejus: While at id, this should have a period at the end to follow the style of the rest of this list.

jenweber:
Thank you @NullVoxPopuli and reviewers! If anyone wants to extend upon this work or make additional fixes, I'm happy to help get more PRs merged in!

PR #920, opened 2019-07-09
LinkTo documentation incorrectly had an attribute value as a string rather than interpolated variable

krainboltgreene:
The model property was supposed to reference the photo variable.

### DIFF
--- a/guides/release/templates/links.md
+++ b/guides/release/templates/links.md
@@ -15,7 +15,7 @@ Router.map(function() {
 ```handlebars {data-filename=app/templates/photos.hbs}
 <ul>
   {{#each this.photos as |photo|}}
-    <li><LinkTo @route="photos.edit" @model={{photo}}>{{photo.title}}</LinkTo></li>
+    <li><LinkTo @route="photos.edit" @model={{this.photo}}>{{this.photo.title}}</LinkTo></li>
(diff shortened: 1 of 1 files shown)


guides/release/templates/links.md
```diff
@@ -15,7 +15,7 @@ Router.map(function() {
 ```handlebars {data-filename=app/templates/photos.hbs}
 <ul>
   {{#each this.photos as |photo|}}
-    <li><LinkTo @route="photos.edit" @model="photo">{{photo.title}}</LinkTo></li>
+    <li><LinkTo @route="photos.edit" @model={{photo}}>{{photo.title}}</LinkTo></li>
```
rwjblue: We should also prefer to use `this.model` to avoid implicit `this` fallback, something like:

```suggestion
    <li><LinkTo @route="photos.edit" @model={{this.photo}}>{{this.photo.title}}</LinkTo></li>
```

What do you think?

krainboltgreene: Great point.

jenweber:
Thanks @krainboltgreene for fixing this! And thanks Robert for the review.

kategengler:
@superlou spotted today that this should be the local variable `photo` and not `this.photo` since it is in the #each

PR #947, opened 2019-07-19
Remove deprecated events

simonihmig:
As per https://github.com/emberjs/rfcs/blob/master/text/0486-deprecate-mouseenter.md

### DIFF
--- a/guides/release/components/handling-events.md
+++ b/guides/release/components/handling-events.md
@@ -151,11 +151,8 @@ Mouse events
 * `contextMenu`
 * `click`
 * `doubleClick`
-* `mouseMove`
 * `focusIn`
 * `focusOut`
-* `mouseEnter`
-* `mouseLeave`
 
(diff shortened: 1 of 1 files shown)


rwjblue:
This should be good to land (removing these from 3.11/3.12 is totally fine). I'm not merging directly because I can't approve the percy diff (and don't want to start failing CI for everything else).

mansona:
In case any future explorer sees @rwjblue's comment: that's not quite how percy works 😄 

Percy will always "auto-approve" the master branch, so PRs will always be just a diff between current master and the requested changes. We haven't set Percy as a required CI check because we can only have a team of 10 approvers so it should be used as a helpful tool and not a blocker for merging.

Just to reiterate: if you merge a PR that is failing on Percy it won't fail CI for everyone in the future. It works slightly different to what we expect with Travis 👍

Issue #929, opened 2019-07-12
Change "classic invocation" to "curly invocation" in angle bracket syntax conversion guide


rwjblue:
Currently the [angle bracket syntax conversion guide](https://guides.emberjs.com/release/reference/syntax-conversion-guide/#toc_angle-bracket-syntax) uses the term "classic invocation" to refer to curly invocation, unfortunately this conflicts pretty strongly with what we decided was "classic components".

Specifically, "classic components" means components extending from `Ember.Component`, it is pretty confusing to have "classic invocation" to use a different definition of "classic". For example, you can absolutely invoke `@glimmer/component` based components with curlies but they are _for sure_ not "classic components".

We should change usages of "classic" in the page I linked above to "curly" or "curlies" (based on usage).

MelSumner:
Looks like this was resolved with #954

PR #949, opened 2019-07-23
Update syntax-conversion-guide.md

MelSumner:
If merged, this PR will remove a sentence that has been reported to be confusing to users (and somewhat contradictory from a technical perspective).


locks:
Shouldn't we switch the whole thing to use "curly invocation" then?

MelSumner:
@locks @rwjblue Since it's still necessary to use "curly invocation" for some things, we chose the phrase classic invocation (and after many hours of discussion that I genuinely hope we aren't trying to re-litigate).

rwjblue:
> we chose the phrase "classic invocation" to delineate the nuance

Its just not correct, as I mentioned in https://github.com/ember-learn/guides-source/issues/929.

MelSumner:
I'm going to close this PR since the edit now has to happen in a different file.

PR #952, opened 2019-07-25
Update & fix api doc urls

sivakumar-kailasam:
### Background
Until the end of 2.x series, we were using the versioned API doc URLs in the guides. This ensured that the API URLs always existed. But in the 3.x series, we started using the `/release` paths in the guides. Since there have been multiple reorgs of APIs in the 3.x series on both ember and ember-data, most of those URLs would lead nowhere. 

Also in 3.11 ember-data restructured its packages which isn't completely represented in the API docs. Until it is sorted out, the release docs will point to the 3.10 ember-data docs instead of the 3.11 docs.

### what's here
- Rewrote old URL patterns to API docs to the latest pattern (eg., http://www.emberjs.com/api, http://emberjs.com/api).
- Rewrote `/release` API URLs in the 3.x folders to use their corresponding versioned URLs instead.
- Rewrote `/ember/release` API URLs in the 3.x folder to use `/ember/3.11` API URLs & `/ember-data/release` API URLs to use `/ember-data/3.10` instead.


jenweber:
#yolo
ariel the mermaid signing a contract with ursula

Issue #925, opened 2019-07-11
Missing optional features


rwjblue:
The following optional features are missing from [/configuring-ember/optional-features/](https://guides.emberjs.com/release/configuring-ember/optional-features/). The prose under each section is directly from the prompt in `@ember/optional-features` when you run `ember feature:enable <feature name>`

### application-template-wrapper

This will remove the `<div class="ember-view">` wrapper for the top-level application template (`app/templates/application.hbs`).
      While this may be a desirable change, it might break your styling in subtle ways:
        - Perhaps you were targeting the `.ember-view` class in your CSS.
        - Perhaps you were targeting the `<div>` element in your CSS (e.g. `body > div > .some-child`).
        - Depending on your choice of `rootElement`, your app might not be wrapped inside a block-level element anymore.

For more information, see https://github.com/emberjs/rfcs/pull/280.

### template-only-glimmer-components

This will change the semantics for template-only components (components without a `.js` file).
      Some notable differences include...
        - They will not have a component instance, so statements like `{{this}}`, `{{this.foo}}` and `{{foo}}` will be `null` or `undefined`.
        - They will not have a wrapper element: what you have in the template will be what is rendered on the screen.
        - Passing classes in the invocation (i.e. `{{my-component class="..."}}`) will not work, since there is no wrapper element to apply the classes to.

While these changes may be desirable for _new components_, they may unexpectedly break the styling or runtime behavior of your _existing components_.

To be conservative, it is recommended that you add a `.js` file for existing template-only components. (You can always delete them later if you aren't relying on the differences.)

For more information, see https://github.com/emberjs/rfcs/pull/278.

rwjblue:
IMHO, a good start for documenting these would be copying that prose above but I do suspect we'd want _some_ tweaks...

PR #955, opened 2019-07-26
Doc optional features "application wrapper" and "template only components"

mixonic:
Closes https://github.com/ember-learn/guides-source/issues/925

Document the optional features:

* application-template-wrapper
* template-only-glimmer-components


rwjblue:
Thank you!

PR #966, opened 2019-07-31
WIP: Testing of external links

### LINKED ISSUE #953: Testing - Write a link checker that looks for 404s in API docs
There are lots of links in the Guides! We need to know when the links to external resources like other website or the API docs are broken. Write a test that checks all API docs links for the release folder. We only want to check the release folder so we don't DDOS ourselves.

There are lots of open source libraries that can check links in markdown files. Pick one and see how it goes!

llunn:
Fixes #953 

Adds test to check if external links are still live (i.e. return status code 200).  By default, links at api.emberjs.com are not checked, but can be manually tested by:

```
cd to/install/path
npm run test:node-followapiurls
```

History of issues:

1. ~~Currently facing an issue that running a local instance of API docs and re-writing the url results in 404's, but accessing the live docs works (I need help resolving this);~~ Resolved by using random fetch to external urls and, by default, skipping api docs urls.
1. Codepen seems to respond with a 403 when accessing via fetch.  ~~Perhaps a custom header is needed.  Tested this with random codepen url's, all yield the same result as Mel's codepen link;~~ Appears to be an incompatibility with the fetch api / no support for CORS on codepen's part.
1. Might want to define a more robust regex on what an external link is defined as.  Currently links starting with `http` are considered external links; it's simple, but works.;
1. There is an issue markdown-link-extractor pulling a trailing apostrophe from links in code snippets.  I currently deal with this as an edge case and strip it off, and assume the world is happy; and
1. There is a nuance with the actual content of the guides that raises a need to introduce an exception list of external url's to not check.  It is likely the case that any url pointing to localhost defines this list.
1. Test timeout has been an issue causing false positive CI failures.  The recent increase to 30 seconds is still causing CI tests to intermittently fail.


jenweber:
Hi! Just an update, this is on my to-do list to review, but it's been a busy week. Thanks so much for this work! I'll get back to you soon.

llunn:
@jenweber I'm think I might come up with a better solution before you commit time to reviewing this.  This seems like the same issue as in the cli guides, which is still open in favor of an automated solution as mentioned [here] (https://github.com/ember-learn/cli-guides/issues/77#issuecomment-445346438).

I'm not sure how Travis ci would deal with fetch requests written into an acceptance test, or if that is some that would be desirable since it would significantly increase the time to run a full suite of tests.

llunn:
@jenweber @sivakumar-kailasam 

Starting on the 17th I will be indisposed until the 25th, but when I return to the keyboard, perhaps we could have a chat on discord on how to proceed with this test.

jenweber:
Ok, thanks for the work so far! Feel free to ping me on Discord when you want to discuss. Perhaps @sivakumar-kailasam or @toddjordan have some ideas.

sivakumar-kailasam:
@llunn we probably need to run the tests against api docs live server in a way that it doesn’t behave like a DOS attack. Before we can figure out a good way to do that, could you add the ability to skip the assertions against api URLs unless some flag/env variable is set?

llunn:
@sivakumar-kailasam yes, i can do that.  Would a suitable solution be to add a randomized or constant delay between requests?  Say 50 to 100 ms? The obvious consequence is an increase in time to complete a round of tests.

sivakumar-kailasam:
An increase in test duration is fine since this won't be a part of the standard test suite anyway.

llunn:
I'm pulling this back into WIP since even with a 30 second test timeout the CI failed this morning.  I'm beginning to believe that these external link checks might not be a good fit for CI.  I can bump the timeout up to to 1 minute, but what happens when that times out?

Should this be a manually executed test set only?

@jenweber @sivakumar-kailasam suggestions?

llunn:
I somehow got my fork in a complete mess.  Closing this PR for now, so I can resolve it.  I will open a new PR when I get this fixed.

PR #976, opened 2019-08-02
Update syntax-conversion-guide.md

MelSumner:
If merged, this PR updates the syntax-conversion guide for more clarity.


guides/release/reference/syntax-conversion-guide.md
```diff
@@ -102,11 +101,11 @@ export default Component.extend({
...
 ```
 
-### When to use classic invocation syntax?
+### Are curly braces gone completely?
 
-Although Angle Bracket syntax is considered to be the best approach, classic invocation syntax is fine to keep using.
 In some cases, classic invocation is still required. 
-When you need direct support for positional arguments or if your components are nested within the file tree, you should still reach for those curly braces:
+
+For example, when you need direct support for positional arguments, or if your components are nested within the file tree, curly braces are still appropriate:
```
locks: "or if your components are nested within the file tree" is no longer true, we have released `<Nested::Support />`!

MelSumner: I can't imagine that is accurate, since we don't have consensus on that syntax.

PR #978, opened 2019-08-03
Update patterns-for-state.md

optikalefx:
I believe style guide says to use `const` here instead of let. As well as not using `.set` in octane. I'm not sure if something else says that ember data models need to use `.set` but for consistency I thought it should be directly set as other objects are.

### DIFF
--- a/guides/release/state-management/patterns-for-state.md
+++ b/guides/release/state-management/patterns-for-state.md
@@ -156,12 +156,12 @@ export default class ProfileComponent extends Component {
 
   @action
   updateName(name) {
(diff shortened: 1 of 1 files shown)


locks:
Thanks for the contribution, but to me `let` seems to be correct here. Ember Data models are proxy objects extending EmberObject, so they still need `.set`. When this changes we'll be sure to update :)

optikalefx:
@locks makes sense for the proxy. If I may ask, why would let be preferred here? That implies that the user variable would be mutated in the future, which I don't think that would be expected behavior.

PR #1000, opened 2019-08-14
Add a note about ED singular usage naming

acorncom:
I just was working with someone on the help channel on Discord who had hit snags based around setting up an `ideas` model (with adapter/serializer) instead of an `idea` model. Thought it might be worth clarifying in the docs that the standard approach here is to go with a singular name


jenweber:
Thank you David! The CTA and the header look a little squished but that seems to be a CSS issue and not your placement. Merged.

PR #1001, opened 2019-08-14
change typo 'leaf-most' to 'left-most'

KellyJohnBraun:


KellyJohnBraun:
Ah, yes that makes sense, I thought it was referring the the opposite direction!

PR #1010, opened 2019-08-22
Automatically ingest Octane tutorial content

chancancode:
The [CI job](https://travis-ci.org/ember-learn/super-rentals-tutorial) from the [super-rentals-tutorial repo](https://github.com/ember-learn/super-rentals-tutorial) will [automatically](https://github.com/ember-learn/super-rentals-tutorial/pull/6) build and push changes to the [`super-rentals-tutorial`](https://github.com/ember-learn/guides-source/commits/super-rentals-tutorial) branch.

This workflow will automatically open a PR (like #1009) whenever there are changes to that branch for review.

This process ensures that we review the final markdown content for any accidental hiccups, before merging them into the [octane branch](https://github.com/ember-learn/guides-source/commits/octane).

PR #1015, opened 2019-08-23
Revert "Merge pull request #1010 from ember-learn/tutorial-ingestion"

chancancode:
This reverts commit dd34bff6c6effe4102f952c8ab633e9a9457a1c7, reversing changes made to c705244c087d4e9758e28a9001793d432b6eefd8.

As it turns out, we don't need this on master at all. Since the workflow if scoped to the `super-rentals-tutorial` branch, the version on the master branch apparently has no effect. Instead, we are supposed to add this workflow to the branch directly as that is the version that is actually used by GitHub.

```
¯\_(ツ)_/¯
```


jenweber:
```
¯\_(ツ)_/¯  merged!
```

PR #1017, opened 2019-08-23
Bring back the workflow to automatically open PRs

chancancode:
This reverts the revert in 84e2f32be6a20e81010d6ee851a1729790d1b27b, I guess. Also pulled in the latest changes to the workflow.

As it turns out, while the workflow code on `master` will never be run, if you don't have any workflows on `master` at all, it breaks the GitHub Actions feature entirely, and all workflows are disabled.

Maybe we just need a dummy workflow on master. Maybe we just need the folder with a `.gitkeep` even. Dunno. But at this point I basically give up!

I know for sure having it both on master and the branch works, because that's what I tested on a different repo last night. I also know for sure that the workflow code on the branch is what actually executes. I just don't care enough to golf it down further!

(based on commit dd34bff6c6effe4102f952c8ab633e9a9457a1c7)

(cherry picked from commit c2b2020edb0484098217e3227624e20e8b0f21e3)

### DIFF
--- a/.github/workflows/tutorial-ingestion.yml
+++ b/.github/workflows/tutorial-ingestion.yml
@@ -1,0 +1,53 @@
+name: Tutorial Ingestion
+
+on:
+  push:
+    branches:
+      - super-rentals-tutorial
+
+jobs:
+  pull-request:
+    name: Open Pull Request
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


chancancode:
Okay, I've tested it more... an empty folder on master doesn't cut it, but an unrelated dummy workflow does seem to work.

PR #1020, opened 2019-08-23
Add a dummy workflow to the master branch

chancancode:
This is needed, because if we didn't have any workflow on the master branch, GitHub thinks that we are not using the feature and will not run any workflows on this repo (that may exists on other branches).

Based on my testing, this seems to be sufficient. This workflow will probably never run in practice, but that seems to be sufficient.

Replaces #1017

### DIFF
--- a/.github/workflows/dummy.yml
+++ b/.github/workflows/dummy.yml
@@ -1,0 +1,12 @@
+name: Dummy Workflow
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


chancancode:
See https://github.com/ember-learn/test-gh-actions/commit/a1c39a117e6fdacc2122af53cb277f51934b9fa4

PR #1027, opened 2019-08-29
fix: @tracked category declared twice

Alonski:


rwjblue:
Agreed, we should fix emberjs/ember.js#18282 but this PR fixes a straight up bug in what we were already describing (and the changeset itself is unrelated to emberjs/ember.js#18282).

PR #1037, opened 2019-09-04
Create .nvmrc

mansona:
Using the latest node version to see if it will improve build times on netlify. This should **not** be merged if it does nothing 😄


sivakumar-kailasam:
This didn't seem to help🤕. Thanks for trying though 🙂

sivakumar-kailasam:
https://github.com/ef4/prember/issues/9 is probably worth revisiting to see if it helps.

mansona:
I was trying to look into something else before my break, the issue with rendering multiple pages at once is that it won't do much for a speed up on a single CPU machine (or VM) 🤔 Essentially I'm assuming that we're CPU bound with this process

Do we have an issue that we can use to track performance?

sivakumar-kailasam:
Travis VMs seem to have two cores. We closed the original issue since since it's resolved.

Issue #1040, opened 2019-09-05
Replace references to jsmodules.io with an MDN link


Phaeilo:
The getting started guide provides an explanation on JavaScript modules. This includes a link to `jsmodules.io`. This website contains some advertising text and no relevant information to JavaScript modules. I think this link should be changed to a more relevant resource, for example MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules

sivakumar-kailasam:
Thanks for bringing this to our attention. We didn't realise that the content on that site changed. 

We'd be happy to accept a PR from anyone with this change.

ClaytonTurner:
I could tackle this

sivakumar-kailasam:
Go for it!

rwjblue:
Closed by https://github.com/ember-learn/guides-source/pull/1041

PR #1041, opened 2019-09-05
Update jsmodules.io reference with an MDN Link (#1040)

ClaytonTurner:


sivakumar-kailasam:
Thanks for the quick PR @ClaytonTurner

PR #1046, opened 2019-09-08
Fix code blocks diff rendering

chancancode:
1. Make lines with "+"/"-" align with other lines by shifting them to the left by one character to compensate.

2. Make the "+"/"-" non-selectable and non-copyable with CSS pseudo elements.

Before:

Before

After:

After

This may want to be upstreamed to guidemaker? It could also have avoided generating those `.diff-operators` to begin with.

### DIFF
--- a/app/styles/app.css
+++ b/app/styles/app.css
@@ -127,3 +127,21 @@ a.edit-page, #search-input {
   position: relative;
   z-index: 10;
 }
+
+/* Fix diff formatting */
+
+code[data-diff] > .diff-deletion,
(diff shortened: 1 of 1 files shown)


chancancode:
Personally, I also find that the font size and line spacing on the code blocks a bit sparse. Since we want to show more of the code now with the new tutorial it made things look longer than they are. Is that something we would want to change?

Along the same lines, I would like to make "Zoey says..." closer to full-width if that's something we are open to changing.

Before:

Narrow Zoey

After:

Wide Zoey

PR #1048, opened 2019-09-08
Fix diff format (take 2)

chancancode:
Using `position: relative; left: -1ch;` had a very mild side effect that when you make text selection, the browser seems a tad confused bout how to draw the highlight color.

It doesn't affect the selection/copy/paste functionality, as far as I can tell, just looks a bit odd visually. It occured to me there may be another way to pull the text to the left without that problem so here it is.

It seems to work the same, but does not have the selection rendering issue.

Before (Safari on OS X):

Screen Shot 2019-09-08 at 3 48 27 PM

Before (Chrome on OS X):

Screen Shot 2019-09-08 at 3 49 27 PM

After (Safari on OS X):

Screen Shot 2019-09-08 at 5 30 38 PM

After (Chrome on OS X):

Screen Shot 2019-09-08 at 5 31 09 PM

### DIFF
--- a/app/styles/app.css
+++ b/app/styles/app.css
@@ -130,18 +130,17 @@ a.edit-page, #search-input {
 
 /* Fix diff formatting */
 
-code[data-diff] > .diff-deletion,
-code[data-diff] > .diff-insertion {
-  position: relative;
-  left: -1ch;
-}
 code[data-diff] .diff-deletion > .diff-operator,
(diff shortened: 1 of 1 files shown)


chancancode:
I found some issues with this approach

chancancode:
Apparently the last approach also caused some visual regressions according to the Percy, which is fixed now.

PR #1049, opened 2019-09-09
Style improvements

chancancode:
This is a spin off from https://github.com/ember-learn/guides-source/pull/1046#issuecomment-529172922, opening for discussion

### DIFF
--- a/app/styles/app.css
+++ b/app/styles/app.css
@@ -144,3 +144,22 @@ code[data-diff] .diff-insertion::before {
   margin-left: -1ch;
   background-color: rgba(93,125,93,.5)
 }
+
+/* Widens Zoey says... */
+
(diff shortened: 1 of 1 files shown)


chancancode:
The Percy diffs are the “intended” (up for discussion) result of this PR. I can’t approve them but they looked good to me.

rwjblue:
Seems generally good to me, but I'm _definitely_ no designer. Could you include the images for before/after from taht comment in the description here to make it easier to find?

chancancode:
Here are the before/after of some select pages:

Before

After

chancancode:
## Before

Before

* * *

## After

5b5f7eb4316e7aae1a44ef2fbf063b117e02f847301deef8a525b909e24e47cf

PR #1051, opened 2019-09-09
Deploy from travis

sivakumar-kailasam:
With this PR we get deploys to master on merges and preview apps for PRs if they're successful. We'll have to look at the travis deploy step logs to get the url for now.

I've reorganized the build steps so that we save a few mins in each build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,11 +46,11 @@ jobs:
       name: 'Ember test suite'
     - stage: deploy preview app
       script:
-        - npm run build --prod
+        - npm run build -- --prod
         - netlify deploy
       if: type = pull_request
     - stage: deploy to prod
       script:
-        - npm run build --prod
+        - npm run build -- --prod
         - netlify deploy --prod
       if: type IN (push) AND branch = master
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,13 +44,13 @@ jobs:
       name: 'Lint markdown'
     - script: npm test
       name: 'Ember test suite'
-    - stage: deploy to prod
-      script:
-        - npm run build --prod
-        - netlify deploy --prod
-      if: type IN (push) AND branch = master
     - stage: deploy preview app
       script:
         - npm run build --prod
         - netlify deploy
       if: branch != master
+    - stage: deploy to prod
+      script:
+        - npm run build --prod
+        - netlify deploy --prod
+      if: type IN (push) AND branch = master
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ before_install:
 stages:
   - lint
   - tests
-  - deploy preview
-  - deploy prod
+  - deploy preview app
+  - deploy to prod
 
 jobs:
   include:
@@ -45,12 +45,12 @@ jobs:
       name: 'Lint markdown'
     - stage: tests
       script: npm test
-    - stage: deploy prod
+    - stage: deploy to prod
       script:
         - npm run build --prod
         - netlify deploy --prod
-      if: branch = master
-    - stage: deploy preview
+      if: type IN (push) AND branch = master
+    - stage: deploy preview app
       script:
         - npm run build --prod
         - netlify deploy
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,10 +45,10 @@ jobs:
       name: 'Ember test suite'
     - script:
         - npm run build -- --prod
-        - netlify deploy
+        - travis_wait netlify deploy
       if: type = pull_request
     - stage: deploy to prod
       script:
         - npm run build -- --prod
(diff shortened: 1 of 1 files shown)


sivakumar-kailasam:
Yep.

chancancode:
I actually appreciate that we can get a preview right now even if the tests are failing (usually for silly reasons, like failing spell check). I suppose I can live without it... but have you considered running the deploy (preview) in parallel with the tests?

sivakumar-kailasam:
That's a fair ask. Updated the config to deploy review apps without
depending on the tests & to have that done in parallel.

*Regards,*
Siva

chancancode:
It seems like we will also lose the "Deploy preview ready!" link from the commit status, which would be a shame. (Unless I am wrong and the `netlify/ember-guides/deploy-preview` link status will still show up even though we are pushing the builds manually?)

We have a similar setup at work (we also push the deploys from Travis, though not to Netlify). I wrote a script (in Ruby) to add the deploy preview link manually using the [GitHub commit status API](https://developer.github.com/v3/repos/statuses/), which you can probably just use (if we don't mind relying on the system Ruby on Travis, which should work fine) or port to Node:

```ruby
begin
  require "bundler/inline"
rescue LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end

gemfile do
  source "https://rubygems.org"
  gem "activesupport"
  gem "jwt"
end

require "active_support/all"
require "json"
require "net/http"
require "openssl"
require "open3"

def get_token(http)
  private_pem = File.read("???.private-key.pem")
  private_key = OpenSSL::PKey::RSA.new(private_pem)

  payload = {
    # issued at time
    iat: Time.now.to_i,
    # JWT expiration time
    exp: 1.minute.from_now.to_i,
    # Integration's GitHub identifier
    iss: ???
  }

  jwt = JWT.encode(payload, private_key, "RS256")

  response = http.post(
    "/installations/???/access_tokens",
    "",
    "Accept" => "application/vnd.github.machine-man-preview+json",
    "Authorization" => "Bearer #{jwt}",
    "Content-Type" => "text/plain"
  )

  if response.code != "201"
    raise "Failed to create token: #{response.body}"
  else
    JSON.parse(response.body)["token"]
  end
end

def pending!(http, token, message)
  create_commit_status(
    http, token,
    context: "Deploy",
    state: "pending",
    description: message,
    target_url: ENV["TRAVIS_BUILD_WEB_URL"]
  )
end

def error!(http, token, message)
  create_commit_status(
    http, token,
    context: "Deploy",
    state: "error",
    description: message,
    target_url: ENV["TRAVIS_BUILD_WEB_URL"]
  )
end

def success!(http, token, message, url)
  create_commit_status(
    http, token,
    context: "Deploy",
    state: "success",
    description: message,
    target_url: url
  )
end

def deploy!(http, token)
  pending!(http, token, "Building...")

  output, status = Open3.capture2("yarn deploy")

  $stdout.print output

  if status.success?
    url = ???
    success!(http, token, "Preview available", url)
  else
    error!(http, token, "Error!")
  end
end

def create_commit_status(http, token, options)
  response = http.post(
    "/repos/ember-learn/guides-source/statuses/#{ENV['TRAVIS_COMMIT']}",
    JSON.generate(options),
    "Accept" => "application/vnd.github.machine-man-preview+json",
    "Authorization" => "token #{token}",
    "Content-Type" => "application/json"
  )

  if response.code != "201"
    raise "Failed to create commit status: #{response.body}"
  end
end

Net::HTTP.start("api.github.com", 443, use_ssl: true) do |http|
  token = get_token(http)

  if ARGV[0] == "--pending"
    pending!(http, token, ARGV[1])
  else
    deploy!(http, token)
  end
end
```

Unfortunately, it's slightly involved to setup cause you can only use the status API [as an app](https://developer.github.com/apps/). But once you have the app setup and generated the key-pairs, the script should Just Work™. There may also be a GitHub Action somewhere that does exactly this.

I don't think this is a blocker though, but it would be good to add the link back to help contributors find their preview. Otherwise, if you have to dig through the build logs to find the URL, I think the utility of the previews are reduced significantly – most people probably won't even know that there _are_ previews for their pull requests.

sivakumar-kailasam:
Thanks for the script. ATM our deploys to master are failing which is the main issue this PR is trying to address. We're aware of the functionalities we're loosing by doing this, so I'll open an issue which includes your script and will detail what needs to be done to be on parity with what we used to have.

sivakumar-kailasam:
We have around 4k files to upload during the deploy step. Netlify cli doesn't show any progress as it uploads files, so travis times out & we end up with a broken deploy again. Will close this PR in favor of a github app that can handle all these issues.

PR #1063, opened 2019-09-12
Use Ember Data decorators without parens where possible (#669)

bertdeblock:


guides/release/models/defining-models.md
```diff
@@ -166,14 +166,14 @@ When the API returns a deeply nested, read-only object or array,
 there is no need to create multiple models with `attr('hasMany')` or `attr('belongsTo')`
 relationships. This could result in a potentially large amount of unnecessary
 code. You can access these objects in the template without transforming them. This can be
-done with `attr()` (No attribute type).
+done with `@attr` (No attribute type).
```
bertdeblock: Wouldn't it be better to write `No transform specified` instead of `No attribute type`? Not part of this PR though.

jenweber:
I have tried this on the latest version of an Octane app, and it appears that the `()` is still needed. Can you double check this?

If I do the following, I get `Error while processing route: index "title" is read-only`

```javascript
import DS from 'ember-data';
const { Model, attr } = DS;

export default class SomeModelModel extends Model {
  @attr title;
}
```

```javascript

export default class ApplicationRoute extends Route {
  model() {
    this.store.createRecord('someModel', {title: 'yo'})
  }
}
```

chancancode:
Yes. This only works in ember data canary. Technically it has been merged into the beta branch too but there has not been a new beta release since. By the time of release (tomorrow?) it will work on both 3.13 stable and 3.14 beta.

jenweber:
Ok, thanks for the context! Let's merge this once that release is out.

chancancode:
Thanks @bertdeblock!

PR #1044, opened 2019-09-07
Use `@model` named argument from route templates

chancancode:
Part of emberjs/rfcs#523 and emberjs/ember.js#18363.

Would love a review to see if I missed any prose in the nearby sections that needs to be updated as well.

PR #1065, opened 2019-09-13
fix(quick-start): use props instead of args in route template code example

jessica-jordan:
This updates the code example in the Quick Start section, explaining how to pass properties to a component to use the `this.` syntax instead of argument syntax `@`.

As far as I'm aware, properties are only prefixed with an `@` when they refer to arguments passed into a component. In this code example the property `model` is used in a route's template though, therefore `this.model` should be used instead of `@model`.

### DIFF
--- a/guides/release/getting-started/quick-start.md
+++ b/guides/release/getting-started/quick-start.md
@@ -153,7 +153,7 @@ Open the `scientists` template and add the following code to loop through the ar
 <h2>List of Scientists</h2>
 
 <ul>
-  {{#each @model as |scientist|}}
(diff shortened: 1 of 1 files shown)


chancancode:
This was an intentional change (#1044), part of emberjs/rfcs#523. As of emberjs/ember.js#18363, `@model` is a full-fledged argument passed into the route template (in addition to being on the JavaScript controller class as `this.model`). However, there is a bit of a temporary problem, in that this only works on canary at the moment, and the Octane blueprints have been updated to track beta releases recently. This should be resolved next week when beta switches to 3.14.

PR #1062, opened 2019-09-11
Updates the edition guide to match the latest changes

pzuraq:
This updates the edition guide with a general editting pass,
and to make sure that all of the information within it is
accurate with the latest changes (e.g. tracked property
interop has changed slightly)

PR #1066, opened 2019-09-13
[FEAT] Adds @action, on, and fn to the edition guide

pzuraq:
Depends on #1062


guides/release/upgrading/editions.md
```diff
@@ -1144,6 +1145,230 @@ objectAt() {
...
+```handlebars
+<button {{on "click" this.doSomething}}>Click Me!</button>
+```
+
+The decorator _is_ important, as it binds the action directly to the class so it
+can reference it later on.
+
+#### The `{{on}}` Modifier
+
+The API for `{{on}}` is the same as JavaScript's native [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener). It receives the event name as the first argument, and a
```
jenweber:
```suggestion
The API for `{{on}}` is the same as JavaScript's native [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener). It receives the event name as the first argument, and a
```

jenweber:
Thank you!

PR #1067, opened 2019-09-13
[FEAT] Updates State Management Docs

pzuraq:
Updates the state management docs to match RFC 478

### DIFF
--- a/guides/release/state-management/computed-properties.md
+++ b/guides/release/state-management/computed-properties.md
@@ -1,4 +1,4 @@
-Computed properties are a pre-Octane concept in Ember. They serve the same
+Computed properties are a concept in Ember prior to Octane. They serve the same
 purpose as tracked properties and native getters, allowing users to respond to
 changes, derive state, and ultimately update the DOM. They also have built-in
 caching to prevent having to perform expensive calculations more than once.
@@ -7,7 +7,7 @@ While computed properties are no longer the recommended default, it's likely
 that you may encounter them in code that hasn't been updated to tracked
 properties just yet, either in existing applications or in the wider Ember
 ecosystem, so this guide exists both to describe how they work and can be used,
-and how they interoperate with tracked properties.
+and how they work with tracked properties.
 
 ## Computed Property Usage
 
@@ -336,12 +336,12 @@ const Person = EmberObject.extend({
 ## Computed Property Dependency Types
 
 You may have noticed that in the previous section, our computed properties were
-depending on normal, undecorated properties. This is possible in classic Ember
+depending on normal properties without decorators. This is possible in classic Ember
 if we always update those properties using Ember's `set` method, which is why
 all of the examples use it. Computed properties can depend on other types of
 values as well though. Altogether, the types of values are:
 
-- Plain, undecorated object properties
+- Plain, object properties without decorators
 - `@tracked` properties
 - `@computed` properties
 - `@dependentKeyCompat` getters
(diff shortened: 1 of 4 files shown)


guides/release/state-management/computed-properties.md
```diff
@@ -1 +1,634 @@
...
+    }
+  }
+  ```
+
+  However, you cannot _chain_ on these properties, as it is a performance
+  pitfall. You can only do 1 level of `@each` watching.
+
+### Defining Setters
+
+If you define a setter for your computed property, it'll work just like a normal
```
jenweber:
```suggestion
If you define a setter for your computed property, it will work just like a normal
```

Feedback from non-native-English-speakers is that contractions can be difficult. We do use `let's` liberally and I don't usually remove it.

guides/release/state-management/computed-properties.md
```diff
@@ -1 +1,634 @@
...
+
+    @computed('friends.@each.name')
+    get friendNames() {
+      return this.friends.map(friend => friend.name);
+    }
+  }
+  ```
+
+  However, you cannot _chain_ on these properties, as it is a performance
+  pitfall. You can only do 1 level of `@each` watching.
```
jenweber:
```suggestion
  pitfall. You can only do one level of `@each` watching.
```

guides/release/state-management/computed-properties.md
```diff
@@ -1 +1,634 @@
-Coming Soon!
+Computed properties are a concept in Ember prior to Octane. They serve the same
+purpose as tracked properties and native getters, allowing users to respond to
+changes, derive state, and ultimately update the DOM. They also have built-in
+caching to prevent having to perform expensive calculations more than once.
+
+While computed properties are no longer the recommended default, it's likely
```
jenweber:
```suggestion
While computed properties are no longer the recommended default, it is likely
```

guides/release/state-management/computed-properties.md
```diff
@@ -1 +1,634 @@
...
+   let counter = new Counter();
+
+   counter.count; // logs 'counted!'
+   counter.count; // logs nothing, the values was cached and hasn't updated
+   ```
+
+2. It will notify other "watchers", such as other computed properties and
+   templates, if any of its dependencies has updated and it needs to be
+   recalculated.
+
```
jenweber: I think maybe we need to add a third point. One of the most significant gotchas of CPs for new devs is that they only run if you are using them in the template. We should state this explicitly somewhere. We can borrow from [explanations in the existing guides](https://guides.emberjs.com/release/object-model/computed-properties/#toc_computed-properties-only-recompute-when-they-are-consumed):

```suggestion
3. A computed property will only recompute its value when it is consumed. Properties are consumed in two ways: by being accessed, like `ironMan.fullName`, or by being referenced in a handlebars template that is currently being rendered, like `{{this.ironMan.fullName}}`.

```

guides/release/state-management/computed-properties.md
```diff
@@ -1 +1,634 @@
-Coming Soon!
+Computed properties are a concept in Ember prior to Octane. They serve the same
+purpose as tracked properties and native getters, allowing users to respond to
+changes, derive state, and ultimately update the DOM. They also have built-in
```
jenweber: Pulling in a prior definition from [here](https://guides.emberjs.com/release/object-model/computed-properties/#toc_computed-properties-only-recompute-when-they-are-consumed)

```suggestion
changes, derive state, and ultimately update the DOM. 
In a nutshell, computed properties let you declare functions as properties.
You create one by defining a computed property as a function, which Ember will automatically call when you ask for the property. 
They also have built-in
```

guides/release/state-management/computed-properties.md
```diff
@@ -1 +1,634 @@
...
+caching to prevent having to perform expensive calculations more than once.
+
+While computed properties are no longer the recommended default, it's likely
+that you may encounter them in code that hasn't been updated to tracked
+properties just yet, either in existing applications or in the wider Ember
+ecosystem, so this guide exists both to describe how they work and can be used,
+and how they work with tracked properties.
```
jenweber: What do you think about explicitly providing guidance to beginners here?

```suggestion
and how they work with tracked properties.
If you are just learning Ember for the first time, you can skip this section.
```

pzuraq:
Thanks for the review, suggestions look great! Re: documenting at all, we do document native classes already, which I believe are in the same category. I agree, this could definitely be in a different "classic" section, but I think it's important since it's pretty likely users will encounter this type of code for some time.

PR #1074, opened 2019-09-15
Resync octane with master

jenweber:
Our octane branch is waaaaay out of date with `master`. This the result of a `git merge master` and resolving lots and lots of conflicts by hand. This PR appears to have many changes, however to review, you should only look at `release` files. All the others are changes to older versions of Ember Guides.

To review, I would use GitLens and look at only changes to `release`, or similarly `git diff octane -- guides/release/*`

One open question, do we want to have the `links.md` and `input-helpers.md` in the Table of Contents? I am not sure when/why they were deleted but this PR reintroduces them.


guides/release/templates/built-in-helpers.md
```diff
@@ -49,38 +49,38 @@ helper lets you create new bindings in your template.
...
 
 Account Details:
-First Name: {{capitalize this.person.firstName}}
-Last Name: {{capitalize this.person.lastName}}
+First Name: {{this.person.firstName}}
+Last Name: {{this.person.lastName}}
 ```
 
-As mentioned in the previous section we use the `concat` helper to render both
+As mentioned in the previous section, we use the `concat` helper to render both
```
pzuraq: It looks like the prose here may need to be updated, in a separate PR

guides/release/templates/index.md
```diff
@@ -188,8 +189,8 @@ Define a function in `app/helpers/sum.js` to create a `sum` helper:
 import { helper as buildHelper } from '@ember/component/helper';
 
 export function sum(params) {
```
pzuraq: We should update this to reflect the latest blueprints I think:

```js
import { helper } from '@ember/component/helper';

export default helper(function sum([first, second]) {
  return first + second;
});
```

Issue #458, opened 2019-02-15
Update Section: Testing


MelSumner:
This section is already pretty up to date, looks like someone went over it recently with the RFC-232 changes. 

Some things to add:

- [ ] explanation for why the types of tests introduced in the guides go in directories with legacy names (acceptance, integration). 
- [ ] explanation of why both Unit Tests and Container Tests both go in the unit directory.
- [ ] examples of nesting modules. (A module within a module).
- [ ] explanation of how to stub/mock a method or object in a unit or container test.
- [ ] mention addons that could help with testing
- [ ]  testing routes: include how to handle testing the asynchronous beforeModel, model, or afterModel hooks.
- [ ] how to test Query Params
- [ ] how to test Native ES Classes
- [ ] What to do if you get the "autorun is disabled, wrap in a run loop" error message?

kovalchik:
I was reading through the Octane testing section the other day and noticed that there's still a lot of usages of `this.set`. For example, in the [components testing section](https://github.com/ember-learn/guides-source/blob/octane/guides/release/testing/testing-components.md):

```js
// set the outer context to red
this.set('colorValue', 'red');
await render(hbs`<PrettyColor @name={{this.colorValue}} />`);
assert.equal(this.element.querySelector('div').getAttribute('style'), 'color: red', 'starts as red');
```

I know there's some instances where using `set` is *required*, but when setting up my tests, I was able to directly assign outer context values without any issues. Should updating the examples in the testing section to not use `get` and `set` also be a checklist item?

```js
// set the outer context to red
this.colorValue = red;
await render(hbs`<PrettyColor @name={{this.colorValue}} />`);
assert.equal(this.element.querySelector('div').getAttribute('style'), 'color: red', 'starts as red');
```

ijlee2:
@MelSumner @jenweber If it's ok with you and the Ember Learning Team, I'd like to help improve the Octane documentation for testing. I am planning to take a break from work on July 26-31 and work on personal projects.

I looked at how the Testing section currently reads from the perspective of a beginner. Some of the contents are great, but I thought the Testing section, overall, can use more polish.

I would like to reorganize existing content, create more diverse test examples (many examples only show testing what happens after `get` or `set`), and provide additional information. By doing so, I think I can address several of the items that @MelSumner listed above.

What are your thoughts?

Best,

---

Proposed structure for Testing:

- Introduction
  - Why do we need testing?
  - What tools are available to help me?
    - Ember CLI (generates blueprints)
[...]

PR #1082, opened 2019-09-16
Add section on stubs to unit testing basics

Gaurav0:
For #458

### DIFF
--- a/guides/release/testing/unit-testing-basics.md
+++ b/guides/release/testing/unit-testing-basics.md
@@ -236,10 +236,10 @@ module('Unit | Service | employees', function(hooks) {
       addJob() {}
     };
 
-    let person = new Person();
+    let person = new MockPerson();
 
     assert.equal(someThing.hire(person), 'John Smith is now an employee');
-    assert.equal(somthing.employees[0], person);
+    assert.equal(someThing.employees[0], person);
   });
 });
 ```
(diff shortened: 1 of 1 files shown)


guides/release/testing/unit-testing-basics.md
```diff
@@ -149,3 +149,97 @@ skip('skip this test', function(assert) {
...
+    return `Answer: ${z}`;
+  }
+}
+```
+
+`someComplicatedOtherMethod` might have complex behavior that you do not want failing your
+unit test for `testMethod`, because you know `testMethod` works otherwise.
+Isolating unit tests is best practice because the tests that are failing should directly
+point to the method that is failing, allowing you to quickly fix it rather than figuring
+out which method the error is in. In we stub the other method:
```
jenweber:
```suggestion
out which method the error is in. Here, we stub the other method:
```

guides/release/testing/unit-testing-basics.md
```diff
@@ -149,3 +149,97 @@ skip('skip this test', function(assert) {
...
+    person.addJob();
+    this.employees.push(person);
+    return `${person.firstName} ${person.lastName} is now an employee`;
+  }
+}
+```
+
+Here, you need to pass a person object, which could be a complex class.
+The `addJob` method in `Person` could be complex as well, perhaps requiring another class.
+Instead, create a simple object and pass it instead.
```
jenweber:
```suggestion
Instead, create a simple object and pass it:
```

PR #1086, opened 2019-09-19
Test for testing external links. (#953)

llunn:
Completes #953 
Replaces #966 

Introduces test code for checking external links in the `/release` directory.  

External link checking is not run as part of current CI tests setup (i.e. not part of npm run test:node).
Instead, two new script options have been added to `package.json` to give the option of including one of these in the travis config if desired:

1. `npm run test:node-local` which performs external link checking for all non-relative links; and
1. `npm run test:node-local-exclude-api-urls`, which performs external link checking for all non-relative links excluding those pointing to the API docs domain.

Also overtly sets up a structure for adding additional test functionality intended to be executed locally and manually (by adding new local/manual mocha tests to `node-tests/local` directory.)


llunn:
Closing to resolve CI issue with netlify deploy

PR #1096, opened 2019-09-23
[DOCS octane] Break out the upgrade guide into individual pages

pzuraq:
Breaks out the upgrade guide into individual pages, including one that hasn't yet been written on Modifiers (I'll be focusing on that next).

Since we can't have nested sections yet, and we don't want to make the Octane upgrade guide a top level section, I opted to put a dash-bullet character at the beginning of the page titles. This is much better for reading the navigation in my opinion: 

Screen Shot 2019-09-23 at 1 05 19 PM

But it does result in a bit of an awkward top level title on the page itself:

Screen Shot 2019-09-23 at 12 57 15 PM

I'm also unsure what the a11y impact of this would be. Personally, I would take the slightly awkwardness of the page title for increased clarity in navigation for now, and focus on getting the ability to have nested navigation asap (it would also be good for the new tutorial), but am open to suggestions. cc @mansona, I would love to help add the nesting to `guidemaker` if you have any guidance on that


MelSumner:
@pzuraq do you think this change proposal should go to https://github.com/ember-learn/guidemaker-ember-template?

jenweber:
Ideally we will not use the "octane" keyword in these. Links or their redirects live forever, so it would be good to keep these pages reusable.

pzuraq:
That’s actually why I put the octane keyword in each of them, so they wouldn’t accidentally overlap with future guides for future editions. These are guides targeted for a specific edition upgrade, so it makes some sense to namespace them.

Happy to remove the keyword, but I think that could be an issue when we release the next edition after Octane and need a new cheat sheet, etc.

jenweber:
We should wait to merge this until after #1104 and the file renaming.

pzuraq:
This has been updated with nesting! Ready for final review 😄

jenweber:
Looking good!

Screen Shot 2019-10-21 at 3 23 01 PM

jenweber:
I also tested the tabbing behavior and it seems good.

PR #1097, opened 2019-09-24
Rename input helper to input component in docs

ondrejsevcik:
It is confusing to mix components and helpers.

Since all examples use angle braced invocation, they are components if I'm not mistaken.


ondrejsevcik:
Done

amyrlam:
Replacing this PR with https://github.com/ember-learn/guides-source/pull/1190

PR #1108, opened 2019-09-26
Remove duplicated paragraph in tutorial

NotGrm:
I was reading chapter 7 of super rental tutorial when I saw a duplicated paragraph. I choose to remove the second one because it seems to me that it was the least complete


chancancode:
Thank you! This file is generated (noted at the top of the file). Can you send the PR to https://github.com/ember-learn/super-rentals-tutorial? If you prefer, you can do everything from the Web UI. If you go to https://github.com/ember-learn/super-rentals-tutorial/blob/master/src/chapters/07-reusable-components.md and edit the file, it will give you the option to open a PR directly from there.

NotGrm:
No problem I will do it there

PR #1109, opened 2019-09-26
Should deploy preview app only when pr has required label

sivakumar-kailasam:


chancancode:
Hm, would this actually work? If you add a label to a pull request it won’t trigger a build again right? I guess the idea is you restart the Travis build manually after adding the label?

Was the main reason that it kept timing out? Or is it because we ran out of “deploys” on Netlify or something?

jenweber:
Our preview deploys were all timing out, I think because concurrent `netlify deploy`s break. I set the Travis jobs to 1 but still see some failures and it's making our CI break down, so I am going to merge this. Even just turning off previews would be helpful, which this should essentially do.

jenweber:
Travis seems to not be running at all for this PR. @sivakumar-kailasam could you take a look?

Issue #1124, opened 2019-10-01
Zoey/Tomster should have `alt=""`


chancancode:
@MelSumner brought it to my attention.

According to [the Web Accessibility tutorial](https://www.w3.org/WAI/tutorials/images/decorative/), the Zoey/Tomster images in "Zoey/Tomster says..." are decorative, therefore should have an empty `alt` attribute (`alt=""`).

I addressed it in the tutorial in https://github.com/ember-learn/super-rentals-tutorial/pull/74, but this problem is more widespread than just the tutorial.

PR #1125, opened 2019-10-01
Removed alternate texts for Zoey & Tomster images in the release version

ijlee2:
Fix for issue #1124 .

The remaining non-empty `alt` instances happen in `release/tutorial` (will be fixed separately) and in pre-Octane Guides.

after_changes


jenweber:
Thank you!

PR #1128, opened 2019-10-02
Aren't selectionStart and selectionEnd also accepted?

bjornharrtell:
Not entirely sure but it looks like is based on this test: https://github.com/emberjs/ember.js/blob/755ea5dbe65d91e0d650707da740aa6900d0a755/packages/%40ember/-internals/glimmer/tests/integration/components/input-curly-test.js#L302


jenweber:
It looks like it fails in IE, based on the code comments. We try to stick to the "happy path" in the Guides, so potentially this should not be included here, but instead added to the API docs if it is not already there. What are your thoughts?

bjornharrtell:
Understandable, closing for now.

PR #1132, opened 2019-10-03
Test external links

llunn:
Completes #953 
Replaces #1086 

Introduces test code for checking external links in the `/release` directory.  

External link checking is not run as part of current CI tests setup (i.e. not part of npm run test:node).
Instead, two new script options have been added to `package.json` to give the option of including one of these in the travis config if desired:

1. `npm run test:node-local` which performs external link checking for all non-relative links; and
1. `npm run test:node-local-exclude-api-urls`, which performs external link checking for all non-relative links excluding those pointing to the API docs domain.

Also overtly sets up a structure for adding additional test functionality intended to be executed locally and manually (by adding new local/manual mocha tests to `node-tests/local` directory.)


llunn:
@jenweber @sivakumar-kailasam I don't know what else to try to get this PR to pass the netlify deploy line in the Travis config.  I've tried updating master and opening the pr and opening a new one.  Ideas?

locks:
Unfortunately it's timing out, so I'm not sure what we can do 🤔 Looking into it.

locks:
@llunn hi! could you rebase this PR on the latest master? CI should be addressed now.

llunn:
@locks i will be able to do this on Sunday or Monday.

llunn:
@locks I've updated my fork and the branch for the PR, and ran a test of all external links.  Results below.
image

You will see 1 x timeout at 60 seconds, 2 x 429 codes (Too many requests) and 1 x 404, which turns out to be a false negative when I tried to the url in a browser.

I can bump the timeout further, but otherwise I'm drawing blanks on how to prevent a test timeout without spamming the server (in this case the api guides server). I'm not sure how to deal with false negatives, and the 429's are not avoidable.  

Aside from the test timeout, I suspect these results are suitable for a manually run test, and the results can be manually checked?

Advice and guidance on how you or others would like to move forward.

llunn:
Happy to see this finally make it into the mix and that it has a little utility!  Hoping to give more, but it turns out that I was wrong and there are indeed only 24 hrs in a day. 😬

PR #1142, opened 2019-10-11
Fix configuration logic of `application-template-wrapper`

MrChocolatine:
- Feature **enabled**: a wrapping `<div>` is created.

ember-feature-enabled

- Feature **disabled**: no wrapping `<div>` is created.

ember-feature-disabled


jenweber:
Thank you! This was wrong for a long time and it is very important!

PR #1143, opened 2019-10-11
Update quick-start.md

RachelAntion:
when I ran the existing guide I got the following uncaught ember error
"Assertion Failed: You cannot use 'PeopleList' as a component name. Component names must contain a hyphen."

when I changed to People-List it worked as expected

### DIFF
--- a/guides/release/getting-started/quick-start.md
+++ b/guides/release/getting-started/quick-start.md
@@ -205,7 +205,7 @@ We're going to tell our component:
    provide this route's `model` as the list of people.
 
 ```handlebars {data-filename="app/templates/scientists.hbs"}
-<PeopleList @title="List of Scientists" @people={{this.model}} />
+<People-List @title="List of Scientists" @people={{this.model}} />
 ```
 
 Go back to your browser and you should see that the UI looks identical.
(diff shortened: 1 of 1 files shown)


acorncom:
@RachelAntion appreciate the PR and sorry to hear you hit snags. The rule about component names having a hyphen was relaxed in relatively recent versions of Ember / Ember CLI, are you using the most current version of Ember CLI or are you on an older one (say 3.6 or older)?

Because if you are on an older one, I'd expect you would hit problems with that particular item on the very latest version of the quick start. Do let us know!

RachelAntion:
Hi! 

I am using v 3.1.4, but that was also what came when I ran the npm install -g ember-cli command from the quick start guide.

Maybe worth putting a note or something about it, it was initially very discouraging to not be able to get the tutorial from the official ember website to work right away.

I also ran into a problem where assert.dom was not resolving and I had to run npm install --save-dev qunit-dom to get it to work.

I'll let you know if I find anything else as I go through

locks:
Thanks for the feedback @RachelAntion! We're going to make sure to add some instructions so the reader makes sure they have the correct versions :)

Issue #1140, opened 2019-10-10
Document nested components, `<Buttons::Info>`


jenweber:
Some time ago, we released support for components that are nested in the directory structure. But this isn't shown in the Guides! 

For this Issue, please find a section of the Guides where we should document this feature. You should do your work from the `octane` branch at first, and open a PR with a draft. After you get some feedback, you should copy the resulting explanation into all versions where this feature was available, in addition to `release`.

This should be a very short entry, just a couple sentences to let people know of the capability.

jenweber:
Closed by #1209 and previous work

PR #1145, opened 2019-10-15
Document nested component invocation #1140

TomaszWegrzyn:
First draft of documentation for nested component invocation(#1140). Wasn't sure where to add it, but I think the best place is close to the end of _Component Templates_.


TomaszWegrzyn:
> I believe you meant to say "double **colon** syntax"

Right, thank you for spotting that! 👍

PR #1156, opened 2019-10-25
Note that it is ok to leave outlet in templates

jenweber:


guides/release/getting-started/quick-start.md
```diff
@@ -79,6 +79,7 @@ Ember detects the changed file and automatically reloads the page for you in the
 You should see that the welcome page has been replaced by "PeopleTracker".
 You also added an `{{outlet}}` to this page,
 which means that any nested route will be rendered in that place.
+It is ok leave the `{{outlet}}` at the bottom of your templates while you are learning.
```
locks: Did this come from real-world feedback? Seems a strange aside.

jenweber:
@locks yeah, I was helping someone new through the QuickStart. I should have added a note in the PR comments about it.

guides/release/getting-started/quick-start.md
chancancode: For what it’s worth, I also find this a little strange. It’s not clear why the person reading this would think they should remove this particular piece of code, compared to any other pieces of code/markup that we told them to add. It may be better to dig into why they think that and resolve it from the root cause (eg if some text implied that in the first place, we should fix that; or if it’s because it was unclear why this was added/needed, then maybe it’s better to just add a line explaining what it does)

jenweber: I removed this line in order to get the other changes merged in.

PR #1163, opened 2019-11-03
Update ember-core-concepts.png to use Octane syntax (and match svg)

muziejus:
I created an svg version of the core concepts image a long time ago, but I didn't (I don't think) also create a png. Now I have, and now the image uses updated Octaney syntax.

I didn't change the router line into class syntax because the line would be too long. Will take advice on what to do, instead!


muziejus:
Also, I hadn't even noticed that @chancancode had updated the svg to include colocated components. Thanks!

rwjblue:
Changed the target branch and closed/reopened to kick off another CI run.

rwjblue:
> I didn't change the router line into class syntax because the line would be too long. 

I don't think we want to change the `Router.map` line anyways, in the new application blueprints we have the `export default class Router extends EmberRouter {}` on a separate line from `Router.map()` line so it is probably fine to leave just the `Router.map` part in the displayed snippet.

rwjblue:
Can you update the route code though? To:

```js
export default class RentalsRoute extends Route {
  model() {
  }
}
```

edit: as I re-read your original comment, maybe this was the one you were saying was too long?

muziejus:
@rwjblue Correct. 

`export default class RentalsRoute extends Route {`

doesn't fit on one line within that space. I didn't want to radically alter the look of this image to make it fit. But it could be done a bunch of ways.

amyrlam:
Made a new PR as this was stale: https://github.com/ember-learn/guides-source/pull/1460 let me know what you think!

Not sure how you about the .png generation or what it's for, did that come from Adobe Illustrator?

amyrlam:
let's open a new pr (convo on https://github.com/ember-learn/guides-source/pull/1460)

PR #1169, opened 2019-11-08
Possible typo: missing self-close?

rictic:
There isn't a `</ToDo>` and it hasn't talked about yielding yet, so this looked like a typo to me.


amyrlam:
@rictic Thanks for the PR! In this case it seems like a literal `TODO` aka placeholder of work in progress, similar to L368. I'm thinking the content of the letter will be added there. I'm going to close the PR but appreciate the help!

PR #1171, opened 2019-11-12
Fix broken markdown on REST Adapter page

thejonrichmond:
There's a bug on the REST Adapter page that has to do with rendering markdown on the page. There was an extra space before the backticks that close out the last code block in the relationships section of that page (link [here](https://guides.emberjs.com/release/models/the-rest-adapter/#toc_relationships)), which messes up the formatting on the rest of the page.

### DIFF
--- a/guides/v1.10.0/models/the-rest-adapter.md
+++ b/guides/v1.10.0/models/the-rest-adapter.md
@@ -207,7 +207,7 @@ the `keyForRelationship` method.
       return key + 'Ids';
    }
  });
- ```
+```
 
 #### Sideloaded Relationships
 


amyrlam:
@thejonrichmond thanks for the fix! i see that bug at this link: https://guides.emberjs.com/v1.10.0/models/the-rest-adapter/

image

thejonrichmond:
@amyrlam yeah it's kind of strange... I see that this formatting issue doesn't exist in the other versions of `v1.x` for this page. I think I was seeing it on `release` because I would search for the rest adapter page and would find `v1.10.0`, and then would click the button at the top to see the newest version, and the bug would exist on the `release` version. It seems to have to do with starting on that broken page. I tried the same thing from `v1.11.0` (which isn't broken) and didn't see the same behavior.

thejonrichmond:
I did a little more digging and saw this error in the console when trying to navigate to the most recent version of this page... I guess it's trying to find `v3.13.0` of the rest adapter page, which doesn't exist, so it's keeping the `v1.10.0` content on the page and updating the URL. I'm unsure what to do about it. 

```
Error: Ember Data Request GET /content/v3.13.0/models/the-rest-adapter.json returned a 404
```

amyrlam:
@thejonrichmond 

CI didn't go green on a rerun, but it's passed! will merge this in

yes, the visual error with the ``` is only on https://guides.emberjs.com/v1.10.0/models/the-rest-adapter/

i'm not sure exactly what you are seeing, can you elaborate? we can continue the convo here even though i'll merge this in

gif of me going from https://guides.emberjs.com/v1.10.0/models/the-rest-adapter/ to https://guides.emberjs.com/release/models/the-rest-adapter/:
rest

thejonrichmond:
@amyrlam Sorry for the confusing explanation! I guess what I was getting at is that clicking on "View v3.13.0" updates the page state to make it look like you're on a v3.13.0 version of the guides (both in the URL and the version dropdown on the left), when the REST Adapter page doesn't exist in version 3. I didn't realize there wasn't a v3.13.0 version of that page until I started working on this issue tbh. 

I'm not sure what (if anything) to do about that though.

amyrlam:
@thejonrichmond That was super confusing! I totally see what you mean now. 

* If you go from the v1.10.0 link https://guides.emberjs.com/v1.10.0/models/the-rest-adapter/, then click "View v3.13.0", the url https://guides.emberjs.com/release/models/the-rest-adapter/ "falsely" exists
* When you go directly to https://guides.emberjs.com/release/models/the-rest-adapter/, it does not exist. 

I filed this issue: https://github.com/ember-learn/guides-source/issues/1179

I'm not sure what's wrong yet, but will try to help figure this out!

PR #1172, opened 2019-11-12
Fix typo

kategengler:


amyrlam:
@jenweber merging this in. i am guessing the percy failure is due to maxing out on builds

PR #1174, opened 2019-11-13
update to dot notation in ember-data section

sly7-7:


sly7-7:
@runspired Could you double check please ? Also I wonder if we should always mention `await` or `.then` in the docs

PR #1177, opened 2019-11-14
add tracked in the classNameBindings replacement

sly7-7:


jenweber:
I added and removed a comma just to get the percy tests to re-run. I am not sure why they totally failed, but it's almost certainly not caused by this PR.

PR #1190, opened 2019-12-07
Rename input helper to input component in docs

amyrlam:
It is confusing to mix components and helpers. Since all examples use angle braced invocation, they are components.

Replaces https://github.com/ember-learn/guides-source/pull/1097 (PR got stale)

cc @ondrejsevcik (added you as co-committer)


locks:
I think we should keep the filename the same so the diff is easier.

amyrlam:
@locks went that route based on the old PR https://github.com/ember-learn/guides-source/pull/1097 ... which got out of date / was faster to make a new PR and add @ondrejsevcik as co-committer than rebase it

> It is confusing to mix components and helpers.
> Since all examples use angle braced invocation, they are components if I'm not mistaken.

> Hi, thanks for bringing this up. We should indeed refer to them as components. However, anytime we rename a file, we need to create a redirect for the old file name. If you search this repository for "redirect," you will find out how to do it. This way, we don't kill links in blog posts and such.

amyrlam:
thanks! has the known node.js 12 netlify timeout error, so is good to merge

PR #1191, opened 2019-12-07
[FEAT] Adds ToC Headings

pzuraq:
Adds the ability to have ToC headings, which allows us to organize the ToC visually.

Screen Shot 2019-12-07 at 1 18 21 PM

Unlike the previous nesting work, this is purely presentational. Headings do not link to any page, and they do not need to be toggled. This just allows us to bucket the main sections more clearly.

Could use help with the styling, I think I landed on something _ok_ but it could use more work.

### DIFF
--- a/app/helpers/includes.js
+++ b/app/helpers/includes.js
@@ -1,0 +1,5 @@
+import { helper } from '@ember/component/helper';
+
+export default helper(([string = '', match]) => {
+  return string.indexOf(match) !== -1;
+});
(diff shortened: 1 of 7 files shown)


pzuraq:
This work could also be pulled upstream if it'd be useful in Guidemaker.

pzuraq:
Tests passed, netlify deploy stalled out. Merging for now, and we can work on upstreaming later on.

PR #1192, opened 2019-12-08
Tutorial URLs reorg

### LINKED ISSUE #1111: Settle on the final tutorial paths!
Things to fix/decide on before shipping Octane guides!

* We need to make /tutorial work (redirects to part 1 intro probably)
* Do we want to use the Nested Nav feature @pzuraq worked on?
* Do we want to keep the `00-...`, `01-...` prefixes?

For the last point, I did it so that they would sort correctly in lexicographic order, and I don't have to worry about specifying the order externally in [my tool](https://github.com/ember-learn/super-rentals-tutorial). However, I think I'll have to add that feature soon regardless, so we could remove them if we prefer.

The main benefit is it's easy to tell the order when viewing the files (in GitHub, or in your editor, etc). The main downside is that if we want to reorder the chapters later or insert chapters in between, we would have to rename a bunch of files and set up redirects. It's not a big deal to me either way.

chancancode:
* Remove number prefixes
* Use nested nav feature to organize part 1 and part 2
* Move images into more conventional location
* Add redirect for `/tutorial` to `/tutorial/part-1`

Known issues:

* Automated builds have not been updated yet

Closes #1111


pzuraq:
Looks good, tests passed but Netlify deploy failed. Going to merge (we should consider making that task optional)

PR #1194, opened 2019-12-08
Update the upgrade guide

jenweber:
To review:
Please read the entire `upgrading/editons/index.md` page and see if it's accurate/true. I copied the old contents and then made significant edits to them, including linking to codemods and trying to give sensible advice for whether to use Native Class codemods. If you want to make edits to this PR directly instead of leaving comments, that is ok by me.

Changes:
- delete some unusued files
- move Octane-specific info out of "How to upgrade" and put it in the index page for the Octane section
- Give advice for native class codemods
- add links to codemods
- split apart learning vs implementing steps for upgrading
- minor wording edits

I found that I did not want to remove any upgrading content yet, even if it repeats some things in the main guides. I feel like I would want it to be there, if I put myself in the shoes of someone who is new to Octane. We can trim it down in a later release. 

Things that are still missing from upgrade guides is a note about generator defaults, setEdition, and whatever the `octanify` command is.


guides/release/upgrading/current-edition/index.md
```diff
@@ -85,3 +73,65 @@ strictly.
...
+
+### Optional features in Octane
+
+A fully-Octane app has the following configuration in `config/optional-features.json`:
+
+```json
+{
+  "application-template-wrapper": false,
+  "jquery-integration": false,
+  "template-only-glimmer-components": true
```
pzuraq:
```suggestion
  "template-only-glimmer-components": true,
  "default-async-observers": true
```

pzuraq:
Tests passed, Netlify deploy failed. Merging.

PR #1197, opened 2019-12-08
Adds back Application Concerns

pzuraq:
We missed this section when updating, adding it back for the time being.


pzuraq:
Tests passing, Netlify failing, merging

PR #1205, opened 2019-12-11
[Octane] Slight tweaks to the new component guides

acorncom:
These are some very slight editorial tweaks to what otherwise reads quite nicely. My goal here was to reduce jargon and simplify the language a tad without over-simplifying the items we're teaching

### DIFF
--- a/guides/release/components/component-arguments-and-html-attributes.md
+++ b/guides/release/components/component-arguments-and-html-attributes.md
@@ -23,9 +23,9 @@ attributes).
   <div class="cta-note">
     <div class="cta-note-body">
       <div class="cta-note-message">
-        You may have noticed that the <code>is-active</code> class on the
+        You may notice that the <code>is-active</code> class on the
         received message avatar from the previous chapters is missing here.
-        We'll return to this in the next chapter on
+        We'll cover that in the next chapter on
         <a href="../conditional-content">Conditional Content</a>.
       </div>
     </div>
@@ -66,19 +66,9 @@ the _browser_ what to do, it's telling your custom tag what to do.
       <div class="cta-note-heading">Zoey says...</div>
       <div class="cta-note-message">
         <p>
-          You might be wondering why Ember doesn't use attribute syntax for the
-          component syntax. In the next section, we'll learn that you can put normal
-          HTML attributes on component tags, and they will end up on the HTML
-          element that the component creates.
-        </p>
-        <p>
-          So when you use a normal attribute name, you're putting a normal HTML
-          attribute on an element. When you use an argument (which starts with
-          <code>@</code>), you're providing data to a component.
-        </p>
-        <p>
-          This also means that arguments can pass any kind of data to the
-          component, even though HTML attributes are always strings
+          You might be wondering why Ember uses the `@` syntax for its
+          components instead of normal HTML attribute syntax. We'll learn why
+          in the next section.
         </p>
       </div>
     </div>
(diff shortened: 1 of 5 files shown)


guides/release/components/component-arguments-and-html-attributes.md
```diff
@@ -23,9 +23,9 @@ attributes).
   <div class="cta-note">
     <div class="cta-note-body">
       <div class="cta-note-message">
-        You may have noticed that the <code>is-active</code> class on the
+        You may notice that the <code>is-active</code> class on the
```
acorncom: They likely haven't until we draw their attention to it here ... 😉

guides/release/components/component-arguments-and-html-attributes.md
```diff
@@ -66,19 +66,9 @@ the _browser_ what to do, it's telling your custom tag what to do.
...
-          element that the component creates.
-        </p>
-        <p>
-          So when you use a normal attribute name, you're putting a normal HTML
-          attribute on an element. When you use an argument (which starts with
-          <code>@</code>), you're providing data to a component.
-        </p>
-        <p>
-          This also means that arguments can pass any kind of data to the
-          component, even though HTML attributes are always strings
```
acorncom: Although interesting, to the common user the example that immediately follows this helps clarify this in a "oh, that's nice and simple" whereas this starts feeling like too much explanation in an aside

guides/release/components/component-arguments-and-html-attributes.md
```diff
@@ -134,15 +124,15 @@ the avatar component now, and they will all end up on the element that has
       <div class="cta-note-heading">Zoey says...</div>
       <div class="cta-note-message">
         <p>
-          <code>...attributes</code> syntax follows the same rules as
-          object-spread syntax in JavaScript. If it appears <em>after</em> an attribute,
```
acorncom: Here too, I don't find the explanation of how it's "just like object-spread syntax", to necessarily aid the general reader. It's possible that's because I'm not working to grok the true mental model that underlies templates here, but I'm not sure we need that at this juncture for folks?

guides/release/components/conditional-content.md
```diff
@@ -72,6 +72,17 @@ _public API_ of the component, so an argument makes sense.
 So, we want to add the `is-active` class if an argument, like say `@isActive`,
 is passed in and is truthy.
 
+```handlebars {data-filename="app/components/avatar.hbs"}
```
acorncom: I moved this up so the reader could go straight to the example before we explain our falsy approach in templates. It was jarring otherwise

guides/release/components/helper-functions.md
```diff
@@ -61,7 +61,7 @@ Since the title is just the `@username` plus some extra stuff, We can replace `@
 </section>
 ```
 
-However, to get the first initial of the string, we'll need to use JavaScript. We'll write a helper function that gives us a substring from inside a string.
```
acorncom: We say the same sort of substring stuff in a sentence right at the beginning of the next section

guides/release/components/helper-functions.md
```diff
@@ -97,7 +97,7 @@ function substring(args) {
 export default helper(substring);
 ```
 
-**This is how we'll normally write helpers in Ember**.
+**This is how we normally write helpers in Ember**.
```
acorncom: Simpler to just state it as fact

guides/release/components/helper-functions.md
```diff
@@ -238,7 +238,7 @@ Now let's say your template is starting to get a bit cluttered and you want
 to clean up the logic in your templates. This can be achieved with the `let`
 block helper.
 The [`{{let}}`](https://www.emberjs.com/api/ember/release/classes/Ember.Templates.helpers/methods/let?anchor=let)
-helper lets you create new bindings in your template.
+helper lets you create new bindings (or temporary variables) in your template.
```
acorncom: Bindings seems jargony here

PR #1208, opened 2019-12-11
Remove duplicate restriction about valid HTML elements in <body>

bartocc:
The `Restrictions` section listed 3 items:

```
- Only valid HTML elements in a `<body>` tag can be used
- No `<script>` tags
- You can only use HTML that is valid in `<body>`
```

I believe the first and third ones are identical and the list can be reduced to:

```
- Only valid HTML elements in a `<body>` tag can be used
- No `<script>` tags
```


jenweber:
Thank you!

PR #1217, opened 2019-12-20
docs: update Array method when compared to EmberArray

MrChocolatine:
The method `includes()` used on the native JS Array should be changed to something else, to a method that really does not exist.

1. You are trying to show that this method does not exist but at this time [native JS Array does implements the `includes()` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes).

2. The snippet currently displayed is incorrect and should return `true`:
```js
let islands = ['Oahu', 'Kauai'];
islands.includes('Oahu');
// => true
```

### DIFF
--- a/guides/release/configuring-ember/disabling-prototype-extensions.md
+++ b/guides/release/configuring-ember/disabling-prototype-extensions.md
@@ -76,15 +76,15 @@ required interfaces using the convenience method `Ember.A`:
 import { A } from '@ember/array';
 
 let islands = ['Oahu', 'Kauai'];
-islands.includes('Oahu');
-// => TypeError: Object Oahu,Kauai has no method 'includes'
+islands.pushObject('Maui');
+// => TypeError: Object Oahu,Kauai has no method `pushObject`
 
 // Convert `islands` to an array that implements the
 // Ember enumerable and array interfaces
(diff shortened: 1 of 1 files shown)


guides/release/configuring-ember/disabling-prototype-extensions.md
```diff
@@ -76,15 +76,15 @@ required interfaces using the convenience method `Ember.A`:
 import { A } from '@ember/array';
 
 let islands = ['Oahu', 'Kauai'];
-islands.includes('Oahu');
-// => TypeError: Object Oahu,Kauai has no method 'includes'
+islands.without('Oahu');
```
rwjblue: I'd rather not use `.without` as the replacement here. Can we use either `.pushObject` (there are more Hawaiian islands 😛) or swap the items to small objects and use something like `.findBy('isBigIsland', false)`?

My concern with `.without` is that it isn't **really** present due to anything with `@ember/array` it's more of a fundamental internal `@ember/object` thing, so using it as the example is a bit more confusing than something you _would_ expect to call on an `@ember/array`.

rwjblue:
Thank you!

PR #1224, opened 2019-12-21
Try to fix missing netlify reporting

jenweber:
The goal of this PR is for Travis to show a failure if the build fails, but otherwise display success, even if netlify times out.

If this works, I think we can repurpose the should-deploy script to make deploys skippable.


jenweber:
I pushed some broken code and confirmed that if the build exits with 1, the test still shows up as a failure, as it should. If the netlify deploy times out, the job will appear to pass. This is what I was hoping for.

PR #1228, opened 2019-12-22
Use decorator to inject store service in example

laurmurclar:
The example given isn't syntactically correct and should be updated to use decorators as its more Octane-y :)


rwjblue:
Thank you @laurmurclar!

PR #1235, opened 2019-12-23
Typo in example's address

maskott:
The address of this example is incorrect. Should be under `app/templates/components` directory.


locks:
Thank you for submitting @maskott! Starting with 3.15 the default is component template co-location, so the text seems to be correct. You can read more in the [respective RFC](https://emberjs.github.io/rfcs/0481-component-templates-co-location.html).

PR #1238, opened 2019-12-24
Fixed a typo (lifecyle -> lifecycle)

ijlee2:
I noticed that the title for https://guides.emberjs.com/release/components/template-lifecycle-dom-and-modifiers/ has a typo.


ijlee2:
Thanks for reviewing, @locks! Do you know how I can fix the Travis CI error? It looks like it errored because of a time-out.

PR #1243, opened 2019-12-28
Use ember-cli-htmlbars over deprecated ember-cli-htmlbars-inline-precompile

mszabo-wikia:
Some of the integration test docs are using the deprecated `ember-cli-htmlbars-inline-precompile` in code examples. This patch updates the docs to use the preferred `ember-cli-htmlbars` instead.

### DIFF
--- a/.local.dic
+++ b/.local.dic
@@ -53,6 +53,7 @@ ember-cli-deprecation-workflow
 ember-cli-cjs-transform
 ember-cli-document-title
 ember-cli-head
+ember-cli-htmlbars
 ember-cli-mirage.
 ember-cli-tutorial-style
 ember-debug-handlers-polyfill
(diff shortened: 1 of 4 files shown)


ijlee2:
@mszabo-wikia Thanks for the updates!

To all, can we also update the hbs-import statements in Test Types and Testing Helpers sections? The Test Types section also has an external link to `htmlbars-inline-precompile`. If we can change that link to `ember-cli-htmlbars`, I think it'd be great!

mszabo-wikia:
👍 Good catch, thanks! I'll update those places in a moment

rwjblue:
@mszabo-wikia Looks like the only remaining issue is a markdown linting error:

```
guides/release/testing/test-types.md
  136:203-136:221  warning  `ember-cli-htmlbars` is misspelt  ember-cli-htmlbars  retext-spell
```

I'm not exactly sure why `ember-cli-htmlbars` fails spelling tests but `htmlbars-inline-precompile` doesn't, but I believe that adding `ember-cli-htmlbars` to [.local.dic](https://github.com/ember-learn/guides-source/blob/master/.local.dic) will fix it.

rwjblue:
I've just pushed that change, 🤞 waiting on CI now...

rwjblue:
Thanks again!

mszabo-wikia:
Thanks for the quick review! :)

PR #1264, opened 2020-01-09
Update dependecies

mansona:
Updating dependencies using `ember-cli-update` and `npm-check -u`.

This also incorporates a fix that will finally stop the 40s rebuild times and crazy seemingly infinite loop that happened when it was rebuilding 🙈(it wasn't infinite but it was **very** long!) 

Sorry for the delay 👍


jenweber:
I have spent some time debugging this. Here is what needs to be done:

1. I turned off some template lints for things that need to be fixed in guidemaker-ember-template
2. Fixed a whole bunch of URLs that for some reason, were not being tested
3. Inserted a `<!-- -->` in any file that ended in a code block

PR #1267, opened 2020-01-12
Add link to full list of void HTML elements

tomdale:
Adds a link to the HTML5 specification with the full list of void elements.

Also fixes a syntax bug where Markdown backticks were used (and not rendered as intended) inside the HTML block, replacing them with `code` tags. Additionally, fixes the CSS to remove the incorrect orange background styling of `code` elements inside the red background of the Zoey CTA.

### DIFF
--- a/app/styles/app.css
+++ b/app/styles/app.css
@@ -155,6 +155,10 @@ code[data-diff] .diff-insertion::before {
   line-height: 1.75em;
   margin: 0.5em 0;
 }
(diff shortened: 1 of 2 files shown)


tomdale:
Templates are HTML - Components - Ember Guides 2020-01-11 16-21-47

jenweber:
Thank you! This looks better and is more informative.

PR #1271, opened 2020-01-13
Prebuild assets in order to decrease deploy time and enable previews, for versions less than 3.12

mansona:
This is an initial experiment to see if the partial build feature that we discussed at a learning team meeting will work and allow us to go back to using Netlify only for builds and branch previews 🎉


jenweber:
## Caveats

- icons don't work on Percy
- If we make big changes, we need to rebuild - changing above the fold md content in pre `2.18`, any time we update the styleguide version, any time we make changes to the app itself

TODO outside of this PR
- document the build steps
- turn automatic deploys back on for previews and master

PR #1272, opened 2020-01-14
Update application-template-wrapper documentation

skaterdav85:
I think the enable/disable bits on the `application-template-wrapper` feature are backwards.


locks:
The Travis failure is unfortunately expected, merging.

PR #1279, opened 2020-01-21
Fix example for customizing the namespace of an adapter

YoranBrondsema:


locks:
Thanks for the help! Tracking this change at https://github.com/ember-learn/guides-source/pull/1281.

Issue #1266, opened 2020-01-10
"First" page in every group takes a few seconds before loading when clicked


jacobq:
Steps to reproduce:

1. Navigate to https://guides.emberjs.com/release/
2. Click on a group to expand it, e.g. "CORE CONCEPTS" > "Components"
3. Optional: open dev tools to monitor network activity
4. Click the first link, e.g. "Templates are HTML"
5. Wait for page to load -- takes longer than expected, apparently due to 404 waiting for https://guides.emberjs.com/content/v3.15.0/components.json

Note 1: other pages (not the first link in their respective group) do not exhibit this behavior
Note 2: loading the problematic pages directly by their URL seems to avoid the problem (request still happens but doesn't delay page rendering)
Note 3: tested on 1/10/2020 in Chrome 79 on Linux x64, but I doubt that matters as it seems to be either an application or hosting problem.

jacobq:
Also, Ember Inspector's "Promises" tracking tool shows a rejection "DS: RESTAdapter#ajax GET to /content/v3.15.0/components.json"

diamondo25:
Experienced this issue too. Not a good experience if you want to start using Ember, as it looks like the page link doesnt work.

marcelotedeschi:
I experienced this issue as well and took a deeper look into what is (potentially) happening.

Essentially, from what I understand, this is an upstream "issue" (though apparently the intended design), from [guidemaker](https://github.com/empress/guidemaker).

More specifically, it happens [here](https://github.com/empress/guidemaker/blob/master/app/routes/version/show.js#L49). When the user navigates to a `version.show.index` ("first page" of sections) route, there is a forced transition to the parent route, that is, replacing the `index` path. Consequently ([L60](https://github.com/empress/guidemaker/blob/master/app/routes/version/show.js#L60)), the record query cannot find the file because the required file is _actually_ called `index.js` and not the parent route `getting-started.json`


![Screen Shot 2020-02-14 at 11 50 32 AM](https://user-images.githubusercontent.com/5921389/74524911-3a119500-4f20-11ea-914c-37f36ce46f8b.png)

Nevertheless, that is the "intended design", since there's a [catch block (L72)](https://github.com/empress/guidemaker/blob/master/app/routes/version/show.js#L64-L72) which then queries for the correct `${path}/index` file. From what I can gather, the only useful outcome of this design is not to have the `/index` on the page URL (but I'm probably not seeing a bigger picture here).

[...]

PR #1290, opened 2020-02-04
index routes take 6s to load in prod - update guidemaker

jenweber:
Our index routes were 404s, which caused very long or failing requests for subsequent resources. This was fixed in guidemaker 2.2.0

Fixes https://github.com/ember-learn/guides-source/issues/1312
Fixes https://github.com/ember-learn/guides-source/issues/1266

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,59 @@
         "@babel/highlight": "^7.0.0"
       }
     },
+    "@babel/compat-data": {
+      "version": "7.8.6",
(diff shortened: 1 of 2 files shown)


jenweber:
It looks like we have Fastboot problems in the preview deploy. We should hold off on merging until we figure out what's wrong and fix it.

jenweber:
Closed, there is a bug blocking the upgrade and we'll need to do a patch release of guidemaker first.

MelSumner:
@jenweber can you re-base this please

PR #1295, opened 2020-02-06
Guidemaker version update

jenweber:
Testing to see if my PR here fixes the issue described in https://github.com/ember-learn/guides-source/pull/1290 and https://github.com/empress/guidemaker/pull/38

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7269,9 +7269,9 @@
               "dev": true
             },
             "resolve": {
-              "version": "1.14.2",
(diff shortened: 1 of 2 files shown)


jenweber:
The pattern seems to be that the build fails when an index.md in the release version is a redirect.

jenweber:
Errors may be related to #1301, with any luck

mansona:
This does indeed solve the problem for all index routes (apart from multi-level routes 😫 like in the tutorials)  We should merge this and then fix the tutorials later because it's still a very good improvement 👍

PR #1328, opened 2020-03-07
Update configuring-ember/index.md

MelSumner:
Adding initial language to indicate that configuring your app is not necessarily required.


locks:
💪

PR #1340, opened 2020-03-16
code shell: fixed ember-cli version

vladimirschneider:


locks:
Hi! As mentioned in the file and the GitHub warning, you edited a file that is generated from the tutorial source files. Could you open this in https://github.com/ember-learn/super-rentals-tutorial/blob/master/src/markdown/tutorial/part-1/01-orientation.md instead? Thank you!

PR #1342, opened 2020-03-16
Update helper-functions.md

anonimusprogramus:
Use namespaced Index component—the sample already has <Message::Avatar> and <Message::Username>.

### DIFF
--- a/guides/release/components/helper-functions.md
+++ b/guides/release/components/helper-functions.md
@@ -4,7 +4,7 @@ Ember's template syntax limits what you can express to keep the structure of you
 
 For instance, let's take a look at a generic message component from a messaging app.
 
-```handlebars {data-filename="app/components/message.hbs"}
+```handlebars {data-filename="app/components/message/index.hbs"}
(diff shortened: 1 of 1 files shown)


jenweber:
Hi @pzuraq and @wycats, I would like to hear your thoughts on what fits best with the Octane mental model - using index, using `message`, or making the parent component named something else like `container` and nest it inside `message`.

@anonimusprogramus thanks for bringing this up! I'll wait for more feedback before merging or suggesting a change.

chancancode:
Personally, I don't see a lot of value in making this change. IMO, it's fine to have a `message.hbs` next to a folder `message/*`. It's fine to move it in and make it the index too, but I don't think that is superior, and so this just add another thing to explain when it is not the point of this chapter. That being said, if there is a good motivation to make this change that I missed, I am not against it either. Mostly I just think it's a bit arbitrary.

anonimusprogramus:
It's fine with me, thanks for considering.

PR #1346, opened 2020-03-17
Update template-lifecycle-dom-and-modifiers.md

anonimusprogramus:
adding import { tracked } from '@glimmer/tracking';


guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -449,6 +449,7 @@ We could then use the `modal` component this way:
 
 ```js {data-filename="app/components/sidebar.js"}
 import Component from "@glimmer/component";
+import { tracked } from '@glimmer/tracking';
```
jenweber:
```suggestion
import { tracked } from "@glimmer/tracking";
```

Good catch! I think these should be double quotes to match the others.

locks:
Travis CI seems to be acting up.

PR #1350, opened 2020-03-21
fixed typo on Pushing Records page

brentzc:


locks:
Thank you for getting all the relevant versions in the same PR :)

PR #1352, opened 2020-03-23
(maint) grammar and consistency edit

jtappa:
A small edit to fix grammar issue of "class's" to "classes" and then to make the two notes on the page consistently styled.

I also noticed that the notes on this page are using markdown's "quote" syntax but are appearing as normal/not styled any differently on the page. Wondering if those styles were just never edited intentionally, or if the styleguide repo also maybe needs some additional styling for notes that aren't the cta ones from the mascot? ¯\_(ツ)_/¯ either way that's something that can be addressed separately from this PR.


jtappa:
I'm not familiar with percy, but it looks like only one of the screencaps it includes is based on changes I made. The two visual changes in this PR would be a note is now italic to match the other one, and one grammar/word change.

PR #1356, opened 2020-03-25
Fix a tiny typo in getting-started/quick-start.md

monitron:
Changed `Mousevent` to correct `MouseEvent`


locks:
Thanks for the help!

PR #1357, opened 2020-03-26
Fix pluralization in provider-components.md

allison-strandberg:
Change "a tests" to "a test".


allison-strandberg:
Didn't notice this was a generated file: I'll edit the original.

PR #1358, opened 2020-03-26
Update helper-functions.md

zapalagrzegorz:
According to MDN as of String.prototype.substr:
Warning: Although String.prototype.substr(…) is not strictly deprecated (as in "removed from the Web standards"), it is considered a legacy function and should be avoided when possible. It is not part of the core JavaScript language and may be removed in the future. If at all possible, use the substring() method instead.


locks:
Hi! Did you accidentally close the R? It seems like a change we want to make :)

zapalagrzegorz:
Hey, I've editied it and opened new one. https://github.com/ember-learn/guides-source/pull/1359

Issue #1362, opened 2020-03-27
Dropdown doesn't mark 3.16 as LTS


mehulkar:
Seen here: https://guides.emberjs.com/release/

<img width="408" alt="Screen Shot 2020-03-27 at 12 38 40 PM" src="https://user-images.githubusercontent.com/490968/77793826-000bd680-7028-11ea-8e26-49430629ccc2.png">

PR #1364, opened 2020-03-28
update LTS versions to account for 3.16

locks:
I opted to list only the current active LTS. Should we keep every version?

Fixes https://github.com/ember-learn/guides-source/issues/1362.

### DIFF
--- a/guides/versions.yml
+++ b/guides/versions.yml
@@ -42,7 +42,5 @@ allVersions:
   - "v3.17.0"
 currentVersion: "v3.17.0"
 ltsVersions:
-  - "v2.18.0"
-  - "v3.4.0"
-  - "v3.8.0"
   - "v3.12.0"
+  - "v3.16.0"


acorncom:
@Locks seems like it might be good to leave the annotation on some of the older ones but not sure it’s needed back to 2.18

amyrlam:
@locks @acorncom shall we merge in as-is? my comment was just a suggestion

locks:
@acorncom @amyrlam the team hasn't decided if we should only list the active ones or all LTS versions. I don't think a middle ground would make sense personally, up to hearing more arguments tho.

amyrlam:
cool makes sense 👍, thought you were holding cause of me

guides/versions.yml
```diff
@@ -42,7 +42,5 @@ allVersions:
   - "v3.17.0"
 currentVersion: "v3.17.0"
 ltsVersions:
-  - "v2.18.0"
-  - "v3.4.0"
-  - "v3.8.0"
```
rwjblue, line 47: FWIW, v3.8.0 is still supported until 2020-04-13.

rwwagner90:
I think it is useful to know old LTS versions, so if you are behind on your Ember version, you can target whatever LTS is closest to your next version bump. Just my 2 cents.

jenweber:
My 2 cents are that we should only show active LTS in this dropdown.

rwjblue:
I think that seems good (better than confusing folks about which LTS to choose), but I wanted to confirm a couple of things:

* We would still list these older LTS versions in the drop down, they just wouldn't be _shown_ as an LTS, right? E.g. after this lands you could still choose the 3.4.0 version of the guides/api docs, you just wouldn't see "LTS" next to the version number.
* Is anyone opposed to listing the LTS versions _somewhere_ (the logical answer here to me is `/releases`)?

locks:
We still show all the released minor versions in the dropdown, this is merely about the "LTS" denotation, as you mention.
This change was motivated by the recent change to https://emberjs.com/releases/lts/ to only include active LTS versions in the schedule table and then list the remaining in prose, along with the EOF dates.

guides/versions.yml
locks, line 47: This resolved itself 😂 I guess we should wait until this week's release.

jenweber:
We talked about this in the learning team meeting, and I think the conclusion was to just show active LTS here. People who want to see older LTS versions should reference the Releases page.

PR #1365, opened 2020-03-29
chore: add examples for aria labels for Input

rajasegar:
Fixes #1353

PR #1366, opened 2020-03-29
Further modernize core-concepts.svg

### LINKED ISSUE #1311: further modernize core-concepts.svg
Continue the great work in https://github.com/ember-learn/guides-source/pull/631.

The [core concepts.svg](https://github.com/ember-learn/guides-source/blob/0cdd97b6518d42c7e6b04a4f8d38c7fbbeaa09de/public/images/ember-core-concepts/ember-core-concepts.svg) file has some snippets that need updating:
Screen Shot 2020-02-19 at 15 25 48

- `this.model` should be `@model`
- The route example should use native classes

rajasegar:
Fixes #1311.

image

### DIFF
--- a/.percy.yml
+++ b/.percy.yml
@@ -1,0 +1,6 @@
+version: 1
+agent:
+  asset-discovery:
+    network-idle-timeout: 250 # ms
+    page-pool-size-min: 8 # pages
+    page-pool-size-max: 15 # pages
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,5 +28,5 @@ script:
   - npm run lint:md
   # if there's a forked PR with the `branch` set to `master`, we want to unset that for Percy
   # to avoid resetting the Percy projects baseline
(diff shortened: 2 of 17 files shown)


rwjblue:
The end of the longer line seems to be getting cut off due to the length of the line with `export default class RentalRoute extends Route {`.

IMHO, we should drop the `export default` bit as this diagram is already not showing _all_ content, and the spirit is still the same:

```js
class RentalRoute extends Route {
  async model() {
  }
}
```

rajasegar:
@rwjblue Removed export default

rwjblue:
The updates to the SVG look great thank you @rajasegar! I think this accidentally includes the changes from https://github.com/ember-learn/guides-source/pull/1365 though, mind rebasing those commits out so we can land the SVG fixes?

rajasegar:
@rwjblue Looks like my branch and origin have diverged. I will close this PR and open a new one cherry-picking only the SVG changes

PR #1367, opened 2020-03-29
Guides angle bracket component name fix

somasuna:
Angle bracket component name fix. Have updated based on the component name used in the before code.


sivakumar-kailasam:
Thanks @somasuna 🙌

Issue #1372, opened 2020-04-01
Hide navbar search icon from Percy 


amyrlam:
Percy shows the es-navbar magnifying glass as a change in every build: 

![image](https://user-images.githubusercontent.com/1372946/78099265-33d35d00-7396-11ea-8c62-6da7a5c72d3c.png)

from https://percy.io/Ember/guides-app/builds/4695682?utm_campaign=Ember&utm_content=guides-app&utm_source=github_status_public

---

I'm not sure where to make this edit in the guides-source repo?

![image](https://user-images.githubusercontent.com/1372946/78099513-f58a6d80-7396-11ea-996a-64b601e21e97.png)

Confused why I don't see `es-navbar` or `EsNavbar` or `navbar` in a "source of truth" spot in the repo or even `Search the guides` text?

---

I know how to hide certain CSS in Percy: https://docs.percy.io/docs/percy-specific-css#section-hiding-regions-with-percy-specific-css

Let me know if you have any ideas!

amyrlam:
Adding a note for posterity cc @jenweber:

This is a known issue on Percy's side and there is nothing that we can do on our end. There is an Intercom thread on this subject where @mansona was working with them

amyrlam:
https://github.com/ember-learn/guidemaker-ember-template/blob/master/app/templates/application.hbs here is the navbar input

i can see if Percy-specific CSS carries over from one repo to another

amyrlam:
This seems fixed now! https://github.com/ember-learn/guides-source/pull/1396

PR #1374, opened 2020-04-01
deps: Upgrade Percy SDK & tweak asset discovery settings

Robdel12:
## What is this?

This PR updates the Percy Ember SDK to the latest version. With that we're updating the underlying package that does asset discovery (`@percy/agent`). In newer versions asset discovery response caching is on by default, so we cache all network requests.

The search svg was missing in random snapshots because the test server was failing to respond when the asset discovery service would request the asset (to save). By enabling caching we can reduce the burden on the server and always have that svg in the cache (along with all other assets that load with a 200).

I've ran a number of test builds back to back locally and it looks pretty good. Curious to see how this works out over some time too 👍 

This should address #1372

### DIFF
--- a/.percy.yml
+++ b/.percy.yml
@@ -1,0 +1,6 @@
+version: 1
+agent:
+  asset-discovery:
+    network-idle-timeout: 250 # ms
+    page-pool-size-min: 8 # pages
+    page-pool-size-max: 15 # pages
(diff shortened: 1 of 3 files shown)


Robdel12:
The diff is correcting the missing icon from the baseline. Once this is merged and `master` runs with a new baseline I'd expect future branches (with these settings) to be much more stable

amyrlam:
thanks @Robdel12 and @mansona for the heads up!

PR #1379, opened 2020-04-05
Helper Functions cleanup

skaterdav85:


guides/release/components/helper-functions.md
```diff
@@ -9,7 +9,7 @@ For instance, let's take a look at a generic message component from a messaging
   @title={{@avatarTitle}}
   @initial={{@avatarInitial}}
   @isActive={{@userIsActive}}
-  class="{{if @isCurrentUser "current-user"}}"
+  class={{if @isCurrentUser "current-user"}}
```
amyrlam, line 12: 👍 thanks for fixing!

guides/release/components/helper-functions.md
```diff
@@ -73,7 +72,7 @@ To implement the helper, we write a JavaScript function that takes its arguments
 arguments, which we'll discuss next.
 
 ```js {data-filename="app/helpers/substring.js"}
-import { helper } from "@ember/component/helper";
+import { helper } from '@ember/component/helper';
```
amyrlam, line 75: @jenweber @locks should we single or double quote as standard in .js / .hbs? just checking before merging

lgtm otherwise!

locks: I think the default linting rule is single quotes for .js and double quotes for .hbs, so this change should be good!

amyrlam: thank you!

Issue #1381, opened 2020-04-08
Show how to add a page title to routes


jenweber:
Every route in an Ember app should have a Page Title, which helps for accessibility and SEO. In a Route's JavaScript file, you can do something like this:

```
  activate() { 
    document.title = 'Page Title Goes Here'; 
   },

  deactivate() {
     document.title = '';  
  }
```

To close this issue:
- Keep it short and sweet
- Add a code example to https://guides.emberjs.com/release/accessibility/page-template-considerations/
- Explain that this is the simplest approach that is a great first step for most apps. There are addons available that help do this in a more dynamic and maintainable way
- Note that the example above does not work with Fastboot apps
- Show how to test the page title (Ava will show an example in the comments below)

jgwhite:
🤚 I’d like to tackle this

sharpshark28:
Could have an example for how to write a simple test.

```
await visit('my-page');
assert.equal(document.title, 'Page Title Goes Here');
```

locks:
Related PR is merged. Thanks!

PR #1383, opened 2020-04-09
Show how to add a page title to routes (#1381)

jgwhite:
Addresses #1381

### DIFF
--- a/guides/release/accessibility/page-template-considerations.md
+++ b/guides/release/accessibility/page-template-considerations.md
@@ -59,7 +59,7 @@ This technique is a reasonable first step, but has limitations:
 - It doesn't work when the page is rendered server-side with FastBoot.
 - It doesn't provide any conventions for constructing nested page titles.
 - It doesn't automatically apply the site title (though you can imagine how to add that).
-- It may not be absolutely robust if data in your app changes *a lot* (imagine a realtime app).
+- It may not be absolutely robust if data in your app changes *a lot* (imagine a real time app).
 
 When your needs become more complex, the following addons facilitate page titles in a more dynamic and maintainable way (including FastBoot support):
 
--- a/guides/release/accessibility/page-template-considerations.md
+++ b/guides/release/accessibility/page-template-considerations.md
@@ -31,7 +31,6 @@ export default class Title extends Helper {
     document.title = this.original;
   }
 }
-
 ```
 
 We can use this helper to set the page title at any point in any template.
@@ -57,11 +56,12 @@ This title will take effect when we enter the “post” route, and the line of 
 
 This technique is a reasonable first step, but has limitations:
 
-- It does *not* work when the page is rendered server-side with FastBoot.
-- It doesn’t provide any conventions for constructing nested page titles.
-- It doesn’t automatically apply the site title (though you can imagine how to add that).
+- It doesn't work when the page is rendered server-side with FastBoot.
+- It doesn't provide any conventions for constructing nested page titles.
+- It doesn't automatically apply the site title (though you can imagine how to add that).
+- It may not be absolutely robust if data in your app changes *a lot* (imagine a realtime app).
 
-When your needs become more complex, the following addons faciliate page titles in a more dynamic and maintainable way (including FastBoot support):
+When your needs become more complex, the following addons facilitate page titles in a more dynamic and maintainable way (including FastBoot support):
 
 - [ember-page-title](https://github.com/adopted-ember-addons/ember-page-title)
 - [ember-cli-head](https://github.com/ronco/ember-cli-head)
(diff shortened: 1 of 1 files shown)


jgwhite:
I’ve added this content as discussed in the a11y strike team meeting, but in working on it with my colleague @zwhitfield3 we’re somewhat concerned that although this *is* the simplest place to start it might raise more questions that it answers. For example, I can imagine someone using this code, but soon needing to incorporate data from the model in the page title, and running into caveats with the `activate` approach (or, indeed, any of the model hooks). The more I look at it, the more I think it would be better to simply promote the existing addons until we have the unified ember-title approach available.

That said, I think the testing snippet is still valuable no matter what, so perhaps we can keep that?

Keen to hear your thoughts.

amyrlam:
@jgwhite 👋 I was at the a11y meeting this week and am on the Learning Team can help get this PR merged. (Have been fixing a CI issue on this repo that should be fixed, if you can do a rebase, from the changes in https://github.com/ember-learn/guides-source/pull/1395.) 

As for the content, I think just adding the testing portion is fine. Or we could keep it but massage the verbiage e.g. instead of "the simplest" to say something like "a simple way, with some gotchas". Will post this PR in #st-a11y.

Also, feel free to tag @ ember-learn/learning-core-team as a reviewer in the future if you want / crosspost in #dev-ember-learning, I can also assist as should be able to make a11y meetings

MelSumner:
@jenweber in the other sections of the guides where I talk about A11y, I talk about how to do it in Ember right now (even if it's complex or not ideal) and then at the end list other addons that can help. It might be good to be consistent.

jgwhite:
Thanks everyone for your input. I’ve taken it all into account and taken it in a slightly new direction.

@amyrlam mind giving this another look?

guides/release/accessibility/page-template-considerations.md
```diff
@@ -15,14 +15,76 @@ Consider this format:
...
+
+  compute([title]) {
+    document.title = title;
+  }
+
+  willDestroy() {
+    document.title = this.original;
+  }
+}
+
```
amyrlam, line 34: Hard to apply in GitHub suggestion, but copy edit suggestion - delete blank line before ```

guides/release/accessibility/page-template-considerations.md
```diff
@@ -15,14 +15,76 @@ Consider this format:
...
+This title will take effect when we enter the “post” route, and the line of code in our `willDestroy` hook will take care of restoring the former title when we return to the “posts” route.
+
+This technique is a reasonable first step, but has limitations:
+
+- It does *not* work when the page is rendered server-side with FastBoot.
+- It doesn’t provide any conventions for constructing nested page titles.
+- It doesn’t automatically apply the site title (though you can imagine how to add that).
+
+When your needs become more complex, the following addons faciliate page titles in a more dynamic and maintainable way (including FastBoot support):
```
amyrlam, line 64:
```suggestion
When your needs become more complex, the following addons facilitate page titles in a more dynamic and maintainable way (including FastBoot support):
```

guides/release/accessibility/page-template-considerations.md
```diff
@@ -15,14 +15,76 @@ Consider this format:
...
+
+<h1>{{@model.title}}</h1>
+```
+
+This title will take effect when we enter the “post” route, and the line of code in our `willDestroy` hook will take care of restoring the former title when we return to the “posts” route.
+
+This technique is a reasonable first step, but has limitations:
+
+- It does *not* work when the page is rendered server-side with FastBoot.
+- It doesn’t provide any conventions for constructing nested page titles.
```
amyrlam, line 61:
```suggestion
- It doesn't provide any conventions for constructing nested page titles.
```

Not familiar with this linter yet, but seems to be giving a warning https://travis-ci.org/github/ember-learn/guides-source/builds/676236275#L545

I will set up these linters locally after this!

guides/release/accessibility/page-template-considerations.md
```diff
@@ -15,14 +15,76 @@ Consider this format:
...
+- It doesn’t automatically apply the site title (though you can imagine how to add that).
```
amyrlam, line 62:
```suggestion
- It doesn't automatically apply the site title (though you can imagine how to add that).
```

guides/release/accessibility/page-template-considerations.md
jgwhite: LOL, good spot!

jgwhite:
Thanks for the review @amyrlam. I’ll attend to your feedback as soon as possible.

jgwhite:
@amyrlam I’ve applied all those suggestions and added an extra caveat (might be overthinking this 😅)

jgwhite:
> warning  `realtime` is misspelt; did you mean `mealtime`?

These listing rules are strict!

PR #1385, opened 2020-04-09
`on` does not prevent default like `action` did

eramod:
I think it would be helpful to point out this difference in the `action` and `on` modifiers. I encountered this issue when I tried to convert a component to Octane and it took a very long time to figure out the problem.

### DIFF
--- a/guides/release/upgrading/current-edition/action-on-and-fn.md
+++ b/guides/release/upgrading/current-edition/action-on-and-fn.md
@@ -133,6 +133,8 @@ export default class ExampleComponent extends Component {
 }
 ```
 
+Note that the `{{action}}` modifier called `event.preventDefault()` under the hood, but the `{{on}}` modifier does not, so if you need to do anything other than the default action for a particular event, you must call `event.preventDefault` within the action. 
+
 This is a replacement for `{{action}}` when it is used as a modifier:
 
 ```handlebars
--- a/guides/release/upgrading/current-edition/action-on-and-fn.md
+++ b/guides/release/upgrading/current-edition/action-on-and-fn.md
@@ -138,7 +138,7 @@ export default class ExampleComponent extends Component {
     <div class="cta-note-body">
       <div class="cta-note-heading">Zoey says...</div>
       <div class="cta-note-message">
-        The `{{action}}` modifier called `event.preventDefault()` under the hood, but the `{{on}}` modifier does not, so if you need to do anything other than the default action for a particular event, you must call `event.preventDefault` within the action.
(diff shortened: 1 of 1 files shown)


locks:
Thank you for the suggestion! I think we could use a "Zoey says" in here (see https://guides.emberjs.com/release/components/#toc_self-closing-tags for an example).

locks:
You can copy the "Zoey says" markup from the raw content of https://github.com/ember-learn/guides-source/blob/master/guides/release/components/index.md!

guides/release/upgrading/current-edition/action-on-and-fn.md
```diff
@@ -133,6 +133,8 @@ export default class ExampleComponent extends Component {
 }
 ```
 
+Note that the `{{action}}` modifier called `event.preventDefault()` under the hood, but the `{{on}}` modifier does not, so if you need to do anything other than the default action for a particular event, you must call `event.preventDefault` within the action.
```
locks, line 136:
```suggestion
<div class="cta">
  <div class="cta-note">
    <div class="cta-note-body">
      <div class="cta-note-heading">Zoey says...</div>
      <div class="cta-note-message">
        The `{{action}}` modifier called `event.preventDefault()` under the hood, but the `{{on}}` modifier does not, so if you need to do anything other than the default action for a particular event, you must call `event.preventDefault` within the action.
      </div>
    </div>
    <img src="/images/mascots/zoey.png" role="presentation" alt="Ember Mascot">
  </div>
</div> 
```

locks:
Oh you are right, forgot that catch: https://deploy-preview-1385--ember-guides.netlify.app/release/upgrading/current-edition/action-on-and-fn/#toc_the-on-modifier

guides/release/upgrading/current-edition/action-on-and-fn.md
```diff
@@ -133,6 +133,18 @@ export default class ExampleComponent extends Component {
 }
 ```
 
+<div class="cta">
+  <div class="cta-note">
+        The `{{action}}` modifier called `event.preventDefault()` under the hood, but the `{{on}}` modifier does not, so if you need to do anything other than the default action for a particular event, you must call `event.preventDefault` within the action.
```
amyrlam, line 141:
```suggestion
        The <code>{{action}}</code> modifier called <code>event.preventDefault()</code> under the hood, but the <code>{{on}}</code> modifier does not, so if you need to do anything other than the default action for a particular event, you must call <code>event.preventDefault</code> within the action.
```

amyrlam:
Fixed the code display, will merge in. Thanks @eramod for the update, as well as @locks and @chancancode! 

image

PR #1387, opened 2020-04-11
Test broken CI

amyrlam:
See if this fixes the .travis.yml, which currently cannot be parsed, see: https://travis-ci.org/github/ember-learn/guides-source/requests

Relates to:
https://github.com/ember-learn/guides-source/pull/1376
https://github.com/ember-learn/guides-source/pull/1375
https://github.com/ember-learn/guides-source/pull/1374

PR #1388, opened 2020-04-11
fix: Correct Travis config for forked branch parsing

Robdel12:
It looks like starting a bash command with `[` breaks the travis config
parser. Using their parser (https://config.travis-ci.com/explore) as a debugger
it looks like adding single quotes around the command will parse correctly

Sorry about that! Lets not merge until we know CI works :) (I assumed it didn't run due to concurrency issues or something like that) 

[edit: locks] Closes https://github.com/ember-learn/guides-source/pull/1387.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,5 +28,5 @@ script:
   - npm run lint:md
   # if there's a forked PR with the `branch` set to `master`, we want to unset that for Percy
   # to avoid resetting the Percy projects baseline
-  - [ "$TRAVIS_PULL_REQUEST_SLUG" != "ember-learn/guides-source" ] && [ "$TRAVIS_PULL_REQUEST_BRANCH" == "master" ] && export PERCY_BRANCH="fork-${TRAVIS_BUILD_NUMBER}"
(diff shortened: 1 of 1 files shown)


Robdel12:
Looks like this does the trick. :D  https://percy.io/Ember/guides-app/builds/4864877 The branch in Percy is `fork-3209` even though I'm PR'ing from the master branch of my fork. Previously this would have reset your projects Percy baseline (since it would report the branch as `master`)

amyrlam:
@Robdel12 @locks thank you so much! sorry i did not realize from my googling that quotes would be a fix 🤦‍♀️

Robdel12:
No problem! You pointed me in the right direction with the config debugger :D

amyrlam:
FWIW it could have been done this way, forgot to cancel the support ticket:

> Could you please re-arrange your command line as below? Since you are trying to put a condition to one of your script line I assume if condition would work for your in this case.

```
...
script:
  - |
    if [ "$TRAVIS_PULL_REQUEST_SLUG" != "ember-learn/guides-source" ] && [ "$TRAVIS_PULL_REQUEST_BRANCH" == "master" ] && ; then
     export PERCY_BRANCH ="fork-${TRAVIS_BUILD_NUMBER};
...
```

Robdel12:
Heh, yeah, I went for a one liner rather than expanding the if out. Either works 👍

amyrlam:
🤘

amyrlam:
To close the loop, for some reason had to change the syntax https://github.com/ember-learn/guides-source/pull/1395/files, but working now!

PR #1390, opened 2020-04-11
fix link from quick start to super rentals tutorial

locks:
Fixes https://github.com/ember-learn/guides-source/issues/1389.

### DIFF
--- a/guides/release/getting-started/quick-start.md
+++ b/guides/release/getting-started/quick-start.md
@@ -390,7 +390,7 @@ Now that your app is deployed, what should you do next?
 ### Advance to the next level
 
 There is an official, free tutorial here in the Guides that delves deeper into some of the features you used today.
-[Give it a try!](./tutorial)
+[Give it a try!](../../tutorial)
 
 ### Explore the ecosystem
 
--- a/guides/release/getting-started/quick-start.md
+++ b/guides/release/getting-started/quick-start.md
@@ -390,7 +390,7 @@ Now that your app is deployed, what should you do next?
 ### Advance to the next level
 
 There is an official, free tutorial here in the Guides that delves deeper into some of the features you used today.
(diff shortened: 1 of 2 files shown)


amyrlam:
I think this will work after a rebase, fix from https://github.com/ember-learn/guides-source/pull/1395

guides/release/getting-started/quick-start.md
```diff
@@ -390,7 +390,7 @@ Now that your app is deployed, what should you do next?
 ### Advance to the next level
 
 There is an official, free tutorial here in the Guides that delves deeper into some of the features you used today.
-[Give it a try!](./tutorial)
+[Give it a try!](./../../tutorial/part-1)
```
jenweber, line 393:
```suggestion
[Give it a try!](../../tutorial/part-1/)
```

I updated this PR with the contents of master and tests are failing for a new reason. Try this to see if it gets past the link checker ^^^

guides/v3.16.0/getting-started/quick-start.md
```diff
@@ -390,7 +390,7 @@ Now that your app is deployed, what should you do next?
 ### Advance to the next level
 
 There is an official, free tutorial here in the Guides that delves deeper into some of the features you used today.
-[Give it a try!](./tutorial)
+[Give it a try!](../../tutorial/part-1)
```
jenweber, line 393:
```suggestion
[Give it a try!](../../tutorial/part-1/)
```

jenweber:
@locks it turns out the fix was good to get tests passing, however it was only applied to 1 of the 2 links that were broken. I updated the second link and we should be good now. I'll merge when it's green. Thanks for fixing this and being persistent about fixing it.

The side problem is that tests didn't catch that these links were incorrect in the first place, but that's a separate issue.

PR #1392, opened 2020-04-13
Clean up conditional-content.md

skaterdav85:


guides/release/components/conditional-content.md
```diff
@@ -105,7 +105,7 @@ is passed in and is truthy.
 <aside ...attributes>
   <div
     class="avatar {{if @isActive "is-active"}}"
-    title="{{@title}}"
+    title={{@title}}
```
rwjblue, line 108: I'd probably still keep these quotes personally. I'm curious, what made you remove them?

skaterdav85: Wouldn't this fail `no-unnecessary-concat`?

jenweber:
It looks like the recommended default was changed in the fall to exclude `no-unnecessary-concat`:  https://github.com/ember-template-lint/ember-template-lint/pull/899/files 

In general I think we mostly want our code samples to only cover the recommended set, so I wouldn't want to go update every example. However it's fairly harmless either way, leaning towards being beneficial for older apps, so I'll merge this.

jenweber:
P.S. Thanks for helping out with cleaning up some guides content! If you find anything else that seems amiss, let us know!

skaterdav85:
@jenweber Interesting, I missed this update. Why would it be recommended for older apps and not newer apps?

jenweber:
I’m not sure, but perhaps there is an issue or PR on it. I did notice there is a “stylistic” set of configs too, so maybe it has something to do with a shift towards that.

skaterdav85:
I'm also curious if this is just a stylistic lint rule or if it has any performance implications (although maybe negligible). I know when I first came across this rule a long time ago, I thought that it maybe existed because doing `attribute="{{foo}}" would result in an extra concatenation operation, as the name of the rule suggests.

PR #1394, opened 2020-04-14
Use app.import in docs

ondrejsevcik:
To avoid the confusion between native import and app.import function


locks:
Merging as tests are passing in CI although something in the CI script is erroring out.

PR #1413, opened 2020-04-29
Use ember-template-lint instead of ember-cli-template-lint

amyrlam:
ember-cli-template-lint is deprecated and was giving this error in local dev:

```
➜  guides-source git:(master) ✗ ember serve
Cannot find configuration for extends: octane
Invalid rule configuration found: no-curly-component-invocation

Build successful (38366ms) – Serving on http://localhost:4200/
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8296,18 +8296,6 @@
         }
       }
     },
-    "@glimmer/compiler": {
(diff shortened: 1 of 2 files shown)


amyrlam:
thanks!

PR #1414, opened 2020-04-29
Update docs to mention Watchman

amyrlam:
I'm on a fresh machine and ran into this error

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ ember serve
 
 Visit [http://localhost:4200](http://localhost:4200)
 
+Note: If you get an error on Mac OSX mentioning `"Error: EMFILE: too many open files, watch"`, try installing Watchman. Install [Homebrew](https://brew.sh/) if you don't have it. Then in you terminal, run `brew update`, then `brew install watchman`.
+
 ## Running tests
 
 Use `npm` to run tests instead of `ember`, since we have additional
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ ember serve
 
 Visit [http://localhost:4200](http://localhost:4200)
 
-Note: If you get an error on Mac OSX mentioning `"Error: EMFILE: too many open files, watch"`, try installing Watchman. Install [Homebrew](https://brew.sh/) if you don't have it. Then in you terminal, run `brew update`, then `brew install watchman`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,6 +53,8 @@ ember serve
 
 Visit [http://localhost:4200](http://localhost:4200)
 
+Note: If you get an error on Mac OSX mentioning `"Error: EMFILE: too many open files, watch"`, try installing Watchman. Install [Homebrew](https://brew.sh/) if you don't have it. Then in you terminal, run `brew update`, then `brew install watchman`.
```
locks, line 56:
```suggestion
Note: If you get an error on Mac OSX mentioning `"Error: EMFILE: too many open files, watch"`, try installing Watchman. Install [Homebrew](https://brew.sh/) if you don't have it. Then in you terminal `brew install watchman`.
```

brew nowadays does an update before trying to install, so it shouldn't be necessary!

chancancode:
hmmm I thought this is mostly not needed anymore, see https://github.com/ember-cli/watch-detector/pull/8, did that turn out to be not true? cc @rwjblue

locks:
FWIW, I recently had to re-install watchman to keep some projects from going into an infinite "file changed" loop.

amyrlam:
Keep me posted... to clarify, the error I got was `"Error: EMFILE: too many open files, watch"`, not `"Could not start watchman" ` (I have gotten that before, but not recently)

chancancode:
@amyrlam what macOS version were you using when you encountered the problem? this seems like a classic "macOS has its default ulimit set too low" problem, I wonder if part of the watchman install process just increased that for you (in which case maybe we should just document that)

chancancode:
Anyway, it's not a blocker since this is just in the README of the guides repo itself, not in the actual documentation. If this helps people I think it's fine to merge (we may want to think more carefully if we were to recommend the readers do the same). Just curious if we made the right call in stop recommending watchman.

amyrlam:
@chancancode am online so replying instantly! Machine is ~1 year old, clean install, macOS Catalina v10.15.4

amyrlam:
i get what you mean, sounds good

i've been meaning to revisit content from scratch, and will see how it goes

PR #1415, opened 2020-04-29
Fix formatting

amyrlam:
Add manual line break between closing ``` and ## header. Not sure if this is the best solution, but wanted to send in a fix, let me know!

WAS:
image

IS:
image


locks:
We have this problem throughout the entires Guides I believe, we might need to fix this at the styleguide leve. cc @mansona

jenweber:
I think it would be ok to add a line of CSS to guidemaker-ember-template, as a stopgap until the new styleguide gets applied.

jenweber:
@amyrlam does it make sense for us to close this?

amyrlam:
Thanks for the nudge, I filed an issue: https://github.com/ember-learn/guidemaker-ember-template/issues/45, closing

Issue #1411, opened 2020-04-25
guides still mention `defaultSerializer` but it's deprecated


sdhull:
The guides mention [`defaultSerializer` here](https://guides.emberjs.com/release/models/customizing-adapters/#toc_authoring-adapters) but it's been [deprecated](https://deprecations.emberjs.com/ember-data/v3.x/#toc_ember-data:default-serializers).

I'm not sure how to update this section of the guide to remove the mention of `defaultSerializer`. I guess if you're authoring adapters, you need to include instructions for setting up appropriate, corresponding serializers as well.

rwjblue:
Good catch. We should definitely remove the deprecated reference.

PR #1420, opened 2020-05-01
Remove documentation of deprecated defaultSerializer functionality

locks:
Fixes https://github.com/ember-learn/guides-source/issues/1411.

Removed documentation relative to the deprecated `defaultSerializer` property of Ember Data's Adapter.
I have also backported the changes to the affected versions in a separate commit, in case the team decides it should not be applied. I added that commit because I consider a bug to recommend code that will result in deprecation warnings.


jenweber:
Thank you!

Issue #1416, opened 2020-04-30
Make sure all button elements have an explicit type


locks:
By default, ember-template-lint will emit a linting error if a button element in a template does not have an explicit type. That means this would fail linting:

```handlebars
<button>
```

But this would not:

```handlebars
<button type="button">
```

We need to make sure to add an explicit type attribute to all button elements so that any user copying it into their app does not fail linting.
Make sure that the example is not showing a linting error on purpose, and that the button type is correct for the example (e.g. it might be `submit` if it's the submit button of a form).

This work should be done for the [**release**](https://github.com/ember-learn/guides-source/tree/master/guides/release) content.

locks:
Assigned to @roderickdevelopers.

roderickdevelopers:
Just an update on the linting error for button type=""
I started with 3.17 and worked my way back. 
Hopefully this helps a little bit

Ember Cli 3.17 Gives the lint error as expected 
![Ember-cli-3 17](https://user-images.githubusercontent.com/8920079/81091690-935ce500-8ecd-11ea-8037-45c03bdc639f.png)

Ember Cli 3.16 Gives the lint error as expected
![ember-cli-3 16](https://user-images.githubusercontent.com/8920079/81091733-9f48a700-8ecd-11ea-9050-93dff4938a83.png)

Ember Cli 3.15 Gives the lint error as expected
![Ember-cli-3 15](https://user-images.githubusercontent.com/8920079/81091762-a96aa580-8ecd-11ea-9533-89986b65d943.png)

Ember Cli 3.14 Does not give the lint error as expected
![Ember-cli-3 14](https://user-images.githubusercontent.com/8920079/81091793-b2f40d80-8ecd-11ea-91e0-3e3ce5695348.png)

ijlee2:
@locks I checked today that in the `guides/release` folder, all buttons have a type. I think we can close this issue now. Thanks again, @roderickdevelopers!

Resolved in:

- #1425 
- #1431 
- #1433

roderickdevelopers:
No problem!! Glad I could help

PR #1425, opened 2020-05-04
1st wave of button type fix (#1416)

roderickdevelopers:
I edited 4 files for the button type fix. I will do others after this one is good. I would like to only edit 3 or 4 at a time. @locks maybe add an item to explain if the developer would need to specify the button type.

### DIFF
--- a/guides/release/accessibility/components.md
+++ b/guides/release/accessibility/components.md
@@ -6,7 +6,7 @@ Browsers have implemented the spec in a way that provides functionality for free
 For example, consider this code sample: 
 
 ```html
-<button>Submit Form</button>
+<button type="submit">Submit Form</button>
 ```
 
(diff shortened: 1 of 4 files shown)


guides/release/accessibility/components.md
```diff
@@ -6,7 +6,7 @@ Browsers have implemented the spec in a way that provides functionality for free
 For example, consider this code sample: 
 
 ```html
-<button>Submit Form</button>
+<button type="submit">Submit Form</button>
```
locks, line 9: Can you revert this change? Since it's in the accessibility section of the guides I'm inclined to keep the current example as it's likely purposeful.

MelSumner: This looks good to me @locks

PR #1428, opened 2020-05-05
Remove glossary

locks:
We do not actively refer to the glossary or update it, so I propose to remove it to have less content prone to rot.


jenweber:
I think I’d prefer to see a PR that adds to this or links to it, over its removal. It’s not well utilized at the moment, but it’s a great place to connect Zoey Says callouts to.

Ifbthings in it seem prone to falling out of date, we could remove those sentences.

locks:
@MelSumner retracting this PR and respective agenda item since there's no desire for discussion.

PR #1430, opened 2020-05-05
implement "update prebuilt guides" script

mansona:
I'm just pushing this to share the work that I have done with the team. I'm pretty sure there is a blocker that will prevent us from shipping this 🤔


locks:
@mansona poke

meanking:
Hi, why don't consider these pull requests?

PR #1302, opened 2020-02-13
Automate search index builds

sivakumar-kailasam:
This should address https://github.com/ember-learn/guides-source/issues/487. It's been a while since we've had that issue open. I suggest we begin with these changes as a starting point & improve on it to address any other concerns we may have about it.

Once this PR is merged, we can trigger rebuilds whenever needed by creating tags.

PR #1432, opened 2020-05-05
Move manual search index updating to a guided script

locks:
I took some inspiration from https://blog.danslimmon.com/2019/07/15/do-nothing-scripting-the-key-to-gradual-automation/ and did a thing.
This moves the instructions regarding updating the search index for the Guides into a guided script, automating the steps that can be automated.
I tried to make it as non-destructive and straightforward as I knew how.

It isn't fancy in terms of error-recovery and such, but if we use the script I believe there's more likelihood of keeping it up to date, as the script would not work otherwise.

Opening for feedback!

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -34,35 +34,6 @@ It is required that all maintainers use 2FA (two factor authentication). These a
 
 ### Updating the guides search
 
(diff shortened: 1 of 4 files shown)


jenweber:
@locks there is an open PR from Siva #1302 that tried to do something similar. Could you look it over and see if there are any takeaways to incorporate? In the meantime, I will run this locally (minus actually pushing to algolia and let you know if I spot anything, but in a visual read it looks good/complete.

locks:
@jenweber I had taken a look but since it's supposed to automate the job in CI the angle is a bit different. The steps on this script is roughly the steps in the GitHub specification!

PR #1436, opened 2020-05-18
Correct "lets" to 'let's"

CodingItWrong:
## Current

On Components > Looping Through Lists, a sentence reads

>  Next, lets add the component class:

## Expected

"lets" here is intended to be a contraction for "let us", but that should be written "let's"

## Change

Updates to the contraction form "let's"


rwjblue:
Thanks @CodingItWrong!

PR #1438, opened 2020-05-18
Fix: Tests fails in 'Integration | Component | rental/image clicking on the component toggles its size' due to incorrect asset location

avocadoras:
The location passed into the component in its test file does not match the location in the project (`assets/images/` instead of `assets/`). 

This causes the test to fail sometimes and it's hard to debug the source of the bug.


avocadoras:
My bad wrong place

PR #1452, opened 2020-06-11
small clarification on implementing Ember.onerror

nvdk:
In modern apps it's required to import ember now, and it's not an import new users will be used to, so it seems logical to add it here. I also suggested to add it in app/app.js, but if there is a better place let me know and I'll update it :).

### DIFF
--- a/guides/release/configuring-ember/debugging.md
+++ b/guides/release/configuring-ember/debugging.md
@@ -78,7 +78,8 @@ Ember.ENV.LOG_STACKTRACE_ON_DEPRECATION = true;
 
 ### Implement an Ember.onerror hook to log all errors in production
 
-```javascript
+```javascript {data-filename=app/app.js}
+import Ember from 'ember';
 import fetch from 'fetch';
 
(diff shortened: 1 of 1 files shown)


guides/release/configuring-ember/debugging.md
```diff
@@ -78,7 +78,8 @@ Ember.ENV.LOG_STACKTRACE_ON_DEPRECATION = true;
 
 ### Implement an Ember.onerror hook to log all errors in production
 
-```javascript
+```javascript {data-filename=app/app.js}
+import Ember from 'ember';
 import fetch from 'fetch';
```
jaredgalanis, line 84: @nvdk thanks for making this change 😄. What do you think about adding an ellipsis below the imports but before the `onerror` code to demonstrate that this hook could be introduced in a variety of types of files and that there's other code not displayed?
 
```suggestion
// ...
```

nvdk:
I've added the ellipsis, thank you for the suggestion @jaredgalanis.

Issue #1443, opened 2020-05-23
Remove references to log bindings and observersFor


jenweber:
In https://guides.emberjs.com/v3.12.0/configuring-ember/debugging/#toc_observers--binding, we talk about a feature that appears to have been removed in 3.1.

To close this ticket, remove the incorrect information from Guides versions 3.1 up to the present (aka the release folder).

See https://github.com/emberjs/ember.js/issues/18929 for context before you start. Thanks!

ijlee2:
I can work this today.

PR #1455, opened 2020-06-13
Removed outdated references to observersFor and LOG_BINDINGS

ijlee2:
This PR addresses the issue https://github.com/ember-learn/guides-source/issues/1443. I removed the section called `Observers / Binding` in the `guides` directory, in all versions between 3.1 and 3.18 (current release).


ijlee2:
Question. I noticed that the `public` folder also contains references to `observersFor` and `LOG_BINDINGS`, for versions 3.1 to 3.11. Did I need to remove these as well?

ijlee2:
I confirmed with @jenweber that the files in `public` directory are generated automatically. Updating the files in `source` should be enough so I believe we can merge this PR (and open a new one if needed).

PR #1460, opened 2020-06-23
Manually update svg code

amyrlam:
Ref: https://github.com/ember-learn/guides-source/pull/1163

I was looking at the old PR list, and made this change. Was tricky to rebase the svg so seemed faster to achieve this way. Let me know what you think! cc @muziejus @MelSumner @rwjblue 

I made the Route Handler font smaller to fit `export default class RentalRoute extends Route {` on one line. 

IS:
image

I don't know Adobe Illustrator, or how to do this in code yet, but maybe we could reduce the padding on the left-hand side:

image

### DIFF
--- a/public/images/ember-core-concepts/ember-core-concepts.svg
+++ b/public/images/ember-core-concepts/ember-core-concepts.svg
@@ -19,6 +19,7 @@
 	.fill-maize{fill:#C09A3F;}
 	.fill-dark-red{fill:#9B5436;}
 	.fill-light-blue{fill:#3095C3;}
+	.text-10{font-size:10px;}
 </style>
 <g id="Background">
 	<rect class="st0" width="116.5" height="540.9"/>
@@ -116,11 +117,11 @@
     <!-- line 3 -->
     <tspan x="0" y="28.8">});</tspan>
   </text>
-	<text transform="matrix(1 0 0 1 295.5 170.3959)" class="font-monospace text-12">
+	<text transform="matrix(1 0 0 1 295.5 170.3959)" class="font-monospace text-10">
     <!-- line 1 -->
-    <tspan class="fill-dark-red" x="0" y="0"> class </tspan><tspan class="fill-maize">RentalRoute </tspan><tspan class="fill-dark-red">extends </tspan><tspan class="fill-maize">Route</tspan> {
+    <tspan class="fill-dark-red" x="0" y="0"> export default class </tspan><tspan class="fill-maize">RentalRoute </tspan><tspan class="fill-dark-red">extends </tspan><tspan class="fill-maize">Route</tspan> {
     <!-- line 2 -->
-    <tspan x="14" y="14.4" class="fill-dark-red">async </tspan><tspan class="fill-dark-blue">model</tspan>() {
+    <tspan x="14" y="14.4" class="fill-dark-blue">model</tspan>() {
     <!-- line 3 -->
     <tspan x="28" y="28.8" class="fill-maize">return </tspan><tspan class="fill-dark-red">this</tspan>.store.<tspan class="fill-dark-blue">findAll</tspan>(<tspan class="fill-light-blue">&apos;rental&apos;</tspan>);
     <!-- line 4 -->


muziejus:
Hey, If I had a bit more leeway w/ the design here, the sizing wouldn't be a problem. The issue was trying to make the new image look as much like the old.

MelSumner:
> Hey, If I had a bit more leeway w/ the design here, the sizing wouldn't be a problem. The issue was trying to make the new image look as much like the old.

We welcome PRs that propose something different! I'd be happy to review an update. Thanks for working on this! :+1:

amyrlam:
@muziejus Thanks for submitting the original PR! Also, sorry that your original PR had been open for awhile. (For context, I was trying to clean up stale PR's.)

Thanks @MelSumner for chiming in. We can make changes to the image if desired. Or, we can close this PR and https://github.com/ember-learn/guides-source/pull/1163. 

If you're on Discord and it's easier to chat there, my handle is the same as GitHub, @ amyrlam. Let me know!

muziejus:
Artboard 1

How does this look? I can svg it up if it looks ok. I sort of guessed at the code highlighting.

chancancode:
We should replace `this.model` with `@model`, but overall it looks pretty good to me. I’ll defer to @amyrlam and others for the rest.

chancancode:
Also, not a hard requirement, but if it fits, in the same amount of physical space, it may be worth aligning the example with the tutorial. I don’t think we necessarily need a one to one direct example, but for example, all else being equal, we may as well pick component and route names that matches some part of the tutorial.

chancancode:
> Also, not a hard requirement, but if it fits, in the same amount of physical space, it may be worth aligning the example with the tutorial. I don’t think we necessarily need a one to one direct example, but for example, all else being equal, we may as well pick component and route names that matches some part of the tutorial.

Totally cool with splitting/deferring that to a future PR too.

chancancode:
Ah also it should be `@rentalUnit={{rentalUnit}}` for it to be syntactically valid

amyrlam:
@muziejus wow, thank you, looks great! Thanks for fixing the orange sidebar, looks so much better!

I would say let's get this update in as an iterative improvement. We could update the image to match the tutorial verbiage in a future PR (I need to check out the tutorial, has been awhile!). If that's cool with you @chancancode @MelSumner. 

So I think the copy edits would be:
- In Route Handler, should be `RentalRoute` instead of `RentalsRoute`
- In Template, should be `@model` instead of `this.model`
- In Template, should be `@rentalUnit={{rentalUnit}}` instead of `@rental=rentalUnit`
- In Template, should be `{{/each}}` instead of `{{aa}}`

Thanks again! 💯

muziejus:
Excellent. I'll put in the concrete changes @chancancode mentioned.

muziejus:
Artboard 1@1 5x

Trying again. Thanks, @amyrlam

amyrlam:
@muziejus Really appreciate your help and going back and forth on this! 😄

It looks great, could the closing parens in Template, after rentalUnit`}}`, be black to match the others? 

I know you said the code highlighting was a guess, but I like it!

muziejus:
Haha. Thanks. We need the text all perfect before I go through the effort of breaking it up to svg it.

hergaiety:
This is a welcome change, thanks for doing this! :) Looking good.

muziejus:
@amyrlam is it supposed to be `RentalRoute`? The file is `rentals.js`, and the route shows all the rentals, not just a specific one.

amyrlam:
@muziejus 👍 RentalsRoute sounds good. 

I was proofing with the one on https://guides.emberjs.com/release/getting-started/anatomy-of-an-ember-app/, apologies.

(I can go through the tutorial soonish to refresh my memory per chancancode's suggestion...and suggest new names that match for a future iteration also.)

muziejus:
cheers. If I do `ember g route rentals`, though, I get a plural, so perhaps the actual getting started text also needs to be corrected.

amyrlam:
@muziejus I can take a closer look at the super rentals tutorial either tonight after sending out the Ember Times, or this weekend, if it's cool to wait! 

Can check all the pluralizations etc / match the examples in the .svg to the tutorial

muziejus:
Super. I can wait 👍

amyrlam:
Just checked out some of the tutorial, it's really good! Hadn't gone through it before. 

> so perhaps the actual getting started text also needs to be corrected.

@muziejus What text do you mean, can you clarify?

---

As for other potential edits to this image: It's a bit tricky since the image is kinda a simplification of what would an app with pluralization would normally have. To me the benefit of the image is to provide a visual mental model but not to be 100% what you would necessarily have in an app e.g. 

URL: 
localhost:4200 > localhost:4200/rentals/123

Router
`this.route('rental', { path: '/rentals/:rental_id' });`

Then the route / template would be `rental.js` instead of `rentals.js` 

...but I worry this is getting too complicated, and it's better to keep the image as-is with the plural.

//

@chancancode I'll try pinging you on Discord at a more suitable hour to discuss, as I'm not sure what change you want to the image to match the super rentals tutorial.

muziejus:
@amyrlam as I understand it, the image is showing how we get the list of rentals, which is why there's the three components and the `{{#each}}`. So I don't see it as a simplification. I see it as showing precisely what happens w/ a model's plural index route. URL is the model name pluralized, the route `findAll`s the model, and then it iterates over `@model` to render each instance of the model in its own component. The tutorial, wisely imo, teaches us this index view before it teaches us to use `findRecord`, which would carry over to a route like `RentalsRentalRoute` etc. (at least that's how I always design my sites)

Anyway, that's my thinking, but I defer to the team for the text. Once we have final ok on the text, I'll turn the Illustrator file into SVG. Incidentally, might it make sense to also keep the Illustrator file as a pdf, just in case? It would still be editable that way; it wouldn't be on the site, though. Again, that's up to you all.

OK 🚀  because we're close!

amyrlam:
@muziejus makes sense, agree! 

I'm thinking the [latest image](https://github.com/ember-learn/guides-source/pull/1460#issuecomment-649928915) with `RentalsRoute` is good to go. But to be safe, I posted this in Discord and will report back / see if anyone comments here or there.

As for the format, can also upload a pdf if it also makes sense! (I am not familiar with editing pdf from Illustrator, but sounds good, I assume we would manually update the pdf to make it match any code .svg edits? Can't hurt to upload it so can test it out if that comes up!)

chancancode:
@amyrlam I agree that the examples here are abbreviated, so doesn't have to have a 1:1 with anything in the tutorial. However, for example, in the template, the tutorial have a pretty similar snippet with slightly different naming: 

https://github.com/ember-learn/super-rentals/commit/87587995cb9e7f2d3a224e1e44c08446de822025#diff-0e3732b2dec14ce25ddcbb96db42ea50R7-R13

```hbs
<div class="rentals">
  <ul class="results">
    {{#each @model as |rental|}}
      <li><Rental @rental={{rental}} /></li>
    {{/each}}
  </ul>
</div>
```

IMO, it doesn't cost anything to align with the tutorial there, so I think it makes sense to do that instead (`<Rental>` component, `app/components/rental.js`, etc).

For the route, the tutorial uses the index route. I get that this example tries to show a specific route so using a `RentalsRoute` seems like a fine/reasonable departure from the tutorial.

muziejus:
Super. Thanks, @amyrlam and @chancancode. Here's the update:

Artboard 1@1 5x

chancancode:
I think the class name in the route should be `RentalsRoute` (plural) to match the file name but other than that looks great to me!

muziejus:
Haha. Of course. I forgot to change it back! Here we go:

Artboard 1@1 5x

chancancode:
oh and it doesn’t really matter in this case, but I think the octane idiom is to make the model hook `async model() {`

muziejus:
OK. Should I also add the `<li>` or would that be too much visual clutter?

Or, here's how it would look:

Artboard 1@1 5x

chancancode:
Ah one more thing. Let’s include the <li> tag. I understand you are trying to dave space- it can go on the same line as the component invocation, wrapping the <Rental> tag, so it will still be the same number of lines.

amyrlam:
All sounds great, thank you for your help and input @muziejus and @chancancode! 🙌

muziejus:
One last q, then: should I also class the `<ul>` tag like 
Artboard 1@1 5x
in the tutorial example?

amyrlam:
I'd say without `class="results"`. In isolation like on [this page](https://guides.emberjs.com/release/getting-started/anatomy-of-an-ember-app/) which is before the tutorial, could be kind of confusing. But no strong opinion either way!

amyrlam:
Closing in favor of https://github.com/ember-learn/guides-source/pull/1469

PR #1466, opened 2020-06-29
Added preview image for ember-template-lint

hergaiety:
Part of an ongoing effort to add visuals to the guides.

This pulls from @josephdsumner's visual example of ember-template-lint in action. The motivation behind this change is showing good DX (developer experience) with the hope that it'll encourage better UX (user experience). The idea here, being, that if it's easy to see template linting notices for things like failed accessibility checks as easily as code style can be evaluated then we may see better accessibility practices in the wild.

Preview of this change:
Screenshot from 2020-06-29 12-54-18


ijlee2:
Thank you for working on taking a screenshot!

PR #1471, opened 2020-07-04
Move release instructions into script

locks:
This PR is a continuation of https://github.com/ember-learn/guides-source/pull/1432. It codifies the instruction to release a new version of the guides in a guided bash script.

While I tried to automate most of the tasks, there are still some that require human interaction.
I tried to make the script as non-destructive as possible, hence the stashing of any change the script user might have

## Notes

I was going to leave some notes about the script here but included them as comments in the script itself!

## Todo
- [x] (BLOCKING) I noticed that I am not updating versions.yml correctly. I need to add the new version to the list of the available versions.
- [ ] (IMPROVEMENT) Improve error handling.
  - If the script user had nothing to stash, the script will emit an error. This can be confusing and should be handled differently


scripts/create-new-version
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+echo "Welcome to the guided process of updating the Guides search index"
+echo "We will now be running a mix of automated (🤖) and manual (👩‍💻) steps, please read the prompts carefully."
```
ijlee2, line 4:
```suggestion
echo "We will now be running a mix of automated (🤖) and manual (👩‍💻) steps. Please read the prompts carefully."
```

MAINTAINERS.md
```diff
@@ -21,15 +21,8 @@ It is required that all maintainers use 2FA (two factor authentication). These a
...
-2. Make sure you have a clean git history with `git status`
-3. Get the latest commits on `master` using `git pull origin master`
-4. Make a branch, i.e. `git checkout -b release-new-version`
-5. Create a directory in `guides` for the version that is one less than the "lastest." For example, if the newest release of Ember is `3.9`, you will make a directory for `3.8` that is a copy of `release`. `mkdir guides/vX.Y.0`
-6. Copy the contents of `guides/release/` into the new directory, `cp -r release/* vX.Y.0/`
-7. Edit `versions.yml` - add the version number to _both_ the end of the list and the `currentVersion`. The last item and `currentVersion` should match.
-8. Double check that the new directory that you made is the latest release minus one. Commit the changes and push your branch.
-9. Create a PR, and in the comments, mention that when it is merged, the person who merges it must update the guides search ASAP, and include a link to this page for instructions. Look at the app in staging, get a review, and merge to `master`. This will trigger an auto deployment.
+1. Run `npm run release:guides` and follow the instructions.
```
ijlee2, line 24: I think describing what a contributor can expect to see at the end will help.

```suggestion
1. Run `npm run release:guides` and follow the script's instructions. At the end, the script will create a PR for releasing a new version of the Ember Guides.
```

MAINTAINERS.md
```diff
@@ -21,15 +21,8 @@ It is required that all maintainers use 2FA (two factor authentication). These a
...
+9. In the newly created PR, mention that when it is merged, the person who merges it must update the guides search ASAP, and include a link to this page for instructions. Look at the app in staging, get a review, and merge to `master`. This will trigger an auto deployment.
 10. Once it is deployed, follow the steps below to get the website search working for your new version.
```
ijlee2, line 26:
```suggestion
1. In the PR, mention that **the person who merges it must update the guides search ASAP**. Add a link to this page for instructions on updating the guides search. Look at the app in staging, get a review, and merge to `master`. This will trigger an auto deployment.
1. Once it is deployed, follow the steps below to get the website search working for your new version.
```

ijlee2: The `Updating the guides search` subsection shows using `yarn`. Can you change the command to use `npm` instead?

scripts/create-new-version
```diff
@@ -0,0 +1,59 @@
...
+
+STARTING_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+TEMP_BRANCH="create-new-guides-version-$(( $RANDOM % 1000 ))"
```
ijlee2, line 7: Instead of `$RANDOM % 1000` (which creates a chance of name collision over time), is it possible to use UNIX timestamp (current time)?

scripts/create-new-version
```diff
@@ -0,0 +1,59 @@
...
+
+echo
+echo "👩‍💻 Merge any pending PRs for the current release https://github.com/ember-learn/guides-source/pulls"
+read -n 1 -s -r -p "Press any key to continue"
+
+CURRENT_VERSION=$(grep "currentVersion" guides/versions.yml | sed "s/currentVersion.*\(v.*\)\"/\1/")
+MAJOR_VERSION=$(echo $CURRENT_VERSION | cut -d "." -f 1 | cut -c 2-)
+MINOR_VERSION=$(echo $CURRENT_VERSION | cut -d "." -f 2)
+PATCH_VERSION=$(echo $CURRENT_VERSION | cut -d "." -f 3)
+NEXT_VERSION=$(($MAJOR_VERSION.$(($MINOR_VERSION+1)).$PATCH_VERSION))
```
ijlee2, line 31: It looks like this line of code can only handle a minor version release.

What does a contributor do if there is a major version release? Do they manually create a subfolder in `guides` as they might do currently?

locks: I decided to focus on minor releases since I didn't want to handle conditionals in the bash script, and we have only had two major releases so far and they might require more coordination already as it is.
The idea would be to add the functionality of major releases in a following effort. You raise a good point, we should say the script currently does patch releases.

MAINTAINERS.md
locks: Yup, on it!.

MAINTAINERS.md
locks: On second read, this isn't completely accurate. It shows the person running the script a URL that they can use to them make the PR, it doesn't make the PR for you.

scripts/create-new-version
ijlee2: I think adding the functionality for major release at a later point is a good idea.

In a follow-up PR, we may want to update `MAINTAINERS.md` (revert some, but not necessarily all, of steps 2-8) to describe what to do when it's time for a major release.

locks:
Fixed the version listing bug which was blocking because the file would be wrong and I  think we're ready to go!

package.json
```diff
@@ -31,6 +31,7 @@
     "test:node": "mocha node-tests --exclude node-tests/local/**",
     "test:node-local": "mocha node-tests/local",
     "test:node-local-exclude-api-urls": "FOLLOW_API_URLS=false npm run test:node-local",
+    "release:guides:minor": "scripts/create-new-version",
```
ijlee2, line 34: I think, when you renamed the script file, you forgot to update the line here.
```suggestion
    "release:guides:minor": "scripts/create-new-minor-version",
```

locks: Thanks for catching it!

locks:
Thank you for your feedback and for pushing for documentation of the process and the script. Merging it for now, and I'll test it For Real for the 3.20 release.

PR #1474, opened 2020-07-07
Document the caching of autotracked properties

YoranBrondsema:
Addresses https://github.com/emberjs/ember.js/issues/19030.

### DIFF
--- a/guides/release/in-depth-topics/autotracking-in-depth.md
+++ b/guides/release/in-depth-topics/autotracking-in-depth.md
@@ -321,4 +321,90 @@ class ShoppingList {
 }
 ```
 
+## Caching of tracked properties
+
+In contrast to computed properties from pre-Octane, tracked properties are not
+cached. A tracked property can also be recomputed even though its dependencies
+haven't changed. The following example shows this behavior:
+
+```js
+import { tracked } from '@glimmer/tracking';
+
+let count = 0;
+
+class Photo {
+  @tracked width = 600;
+  @tracked height = 400;
+
+  get aspectRatio() {
+    count++;
+    return this.width / this.height;
+  }
+}
+
+let photo = new Photo();
+
+console.log(photo.aspectRatio); // 1.5
+console.log(count); // 1
+console.log(photo.aspectRatio); // 1.5
+console.log(count); // 2
+
+photo.width = 800;
+
+console.log(photo.aspectRatio); // 2
+console.log(count); // 3
+```
+
+From the value of `count`, we see that `aspectRatio` was calculated 3 times.
+
+Recomputing is fine in most cases. If the computation that happens in the
+getter is very expensive, however, you will want to cache the value and
+retrieve it when the dependencies haven't changed. You want to recompute only
+if a dependency has been updated.
+
+Ember's [cache API](https://github.com/ember-polyfills/ember-cache-primitive-polyfill) lets
+you cache a getter in 2 steps:
+
+1. Pass a function that is costly to compute to `createCache`.
+1. In the getter, call the function with `getValue` and return its value.
+
+With these steps in mind, let's introduce caching to `aspectRatio`:
+
+```js
+import { tracked } from '@glimmer/tracking';
+import { createCache, getValue } from '@glimmer/tracking/primitives/cache';
+
+let count = 0;
+
+class Photo {
+  @tracked width = 600;
+  @tracked height = 400;
+
+  // `#` indicates a private field on the class
+  #aspectRatioCache = createCache(() => {
+    count++;
+    return this.width / this.height;
+  });
+
+  get aspectRatio() {
+    return getValue(this.#aspectRatioCache);
+  }
+}
+
+let photo = new Photo();
+
+console.log(photo.aspectRatio); // 1.5
+console.log(count); // 1
+console.log(photo.aspectRatio); // 1.5
+console.log(count); // 1
+
+photo.width = 800;
+
+console.log(photo.aspectRatio); // 2
+console.log(count); // 2
+```
+
+From the value of `count`, we see that, this time, `aspectRatio` was calculated
+only twice.
+
 <!-- eof - needed for pages that end in a code block  -->
--- a/guides/v3.15.0/in-depth-topics/autotracking-in-depth.md
+++ b/guides/v3.15.0/in-depth-topics/autotracking-in-depth.md
@@ -149,7 +149,7 @@ export default class HelloComponent extends Component {
   constructor() {
     super(...arguments);
 
-    fetch('./api/preferences')
+    fetch('/api/preferences')
       .then(r => r.json()) // convert the response to a JS object
       .then(response => {
         this.language = response.preferredLanguage;
@@ -262,7 +262,7 @@ export default class ApplicationController extends Controller {
 ```handlebars {data-filename=app/templates/application.hbs}
 {{@model.fullName}}
 
-<button {{on "click" (fn this.updateName 'Krati' 'Ahuja')}}>
+<button type="button" {{on "click" (fn this.updateName 'Krati' 'Ahuja')}}>
   Update Name
 </button>
 ```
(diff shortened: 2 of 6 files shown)


ijlee2:
Hi, @YoranBrondsema. I appreciate your looking into the current state for caching getters. I didn't know about `ember-cache-primitive-polyfill` before, so I learned something new from your PR today.

I thought that the code examples were clear and illustrated the effect of caching well. I think we can shorten the surrounding texts a little (we can come back to this suggestion later).


### Potential problem

Right now, I'm unable to implement your example as a component in a new Ember 3.19 app. When I introduced the import statement for the polyfill, I got the error,
```
Can't resolve '@glimmer/tracking/primitives/cache' in '/Users/isaac/Desktop/hello-world'
```
I believe it's the problem reported in https://github.com/ember-polyfills/ember-cache-primitive-polyfill/issues/3.

I reached out to @pzuraq on Discord to see if the fix is something that we can anticipate soon. If not, I wonder if the Guides today should show the polyfill solution, or if we can present a native JS solution instead (if this is simple to do).

Please let us know what you think. Thank you! 🙂

YoranBrondsema:
@ijlee2 Thanks for your prompt reply.

I have to be honest with you, I didn't even try `ember-cache-primitive-polyfill` but went straight to using `@cached` from [tracked-toolbox](https://github.com/pzuraq/tracked-toolbox) for my own app (which runs in production). I 100% agree that we can't publish a guide for something that doesn't work.

I suggest we await @pzuraq's input and I imagine it'll work when https://github.com/ember-polyfills/ember-cache-primitive-polyfill/issues/3 is resolved.

ijlee2:
Thanks for understanding. I also think it's a good idea to wait for the issue to be resolved. I think Chris and others are more aware of the issue since yesterday so I'm hopeful that it will be resolved soon.

pzuraq:
I believe this API should be released in 3.20, so we can probably remove the text about using the polyfill and land this guide once it is available. We should be able to confirm with the current beta.

YoranBrondsema:
Cool, I removed the mention of the polyfill. Right now, it refers to the [RFC](https://github.com/emberjs/rfcs/blob/master/text/0615-autotracking-memoization.md). Once it lands, I imagine it should refer to the doc in the API docs (https://api.emberjs.com/) instead?

ijlee2:
@YoranBrondsema Apologies for a late reply.

For v3.15-3.19 Guides, I think I'd prefer that we present a working solution for caching getters, something that developers can use now. We would continue to use your full name example to demo the solution.

When the API is officially a part of Ember (say, it's v3.20 as an example), we can link to the API docs for more information, yep! (Or to the RFC if the API docs are yet to be updated.)

**Discussion items:**

What are your thoughts on my preference for presenting a working solution?

Besides using `@computed`, off the top of my head, I don't know a solution to caching getters with vanilla JavaScript. It should be doable, I believe, so I need to spend some time on research. (It's possible that @pzuraq already presented a solution in a blog post. ☺️)

If you know of a solution, can you update your full name example to show us how?

rwjblue:
> I believe this API should be released in 3.20, so we can probably remove the text about using the polyfill and land this guide once it is available. We should be able to confirm with the current beta.

Actually, the internals are available in 3.20 but the feature itself is not exposed there just yet. Needs feature flagging and whatnot (so likely 3.22 at the earliest for usage without the polyfill).

YoranBrondsema:
Thanks for all your feedback. A quick summary:

## 3.20 to 3.22? --> mention polyfill
From 3.20 to whenever the feature is available without a feature flag (possibly 3.22), we should mention that the polyfill [ember-cache-primitive-polyfill](https://github.com/ember-polyfills/ember-cache-primitive-polyfill) is required.

## 3.15 to 3.19 --> also mention polyfill?
The [doc for ember-cache-primitive-polyfill](https://github.com/ember-polyfills/ember-cache-primitive-polyfill#compatibility) mentions that it's compatible with Ember 3.13 and up. So I imagine we can use the same guide for 3.15 all the way to 3.22? Or is anything that opposes the use of the caching polyfill for 3.15 to 3.19?

## Release of feature and beyond --> remove mention of polyfill
We remove the mention of the polyfill when the caching feature is available in Ember.js without a feature flag.

## Release of [@cached RFC](https://github.com/emberjs/rfcs/pull/566) --> replace primitive caching API with `@cached` decorator
The `@cached` decorator is a lot simpler to use for developers than the primitive caching API.

What are you thoughts @ijlee2 @pzuraq @rwjblue ?

rwjblue:
FYI - The `@glimmer/tracking/primitives/cache` API has been enabled by default over in https://github.com/emberjs/ember.js/pull/19067 and is currently slated for release as part of 3.22.0 (assuming everything goes well on canary + beta).

ijlee2:
@YoranBrondsema Thanks for creating a suggested path forward.

I think your plan is sound and suggesting the polyfill for 3.15-3.19 works. (The import issue doesn't seem to have been addressed yet. However, I don't want that to be a blocker for your PR.)

If I understood Robert's last comment correctly, it seems we can enable a feature flag starting 3.20? If so, I think we can document the use of the flag instead of the polyfill (from 3.20 until release of feature).

Let me try out the feature flag after work today.

rwjblue:
> If I understood Robert's last comment correctly, it seems we can enable a feature flag starting 3.20? If so, I think we can document the use of the flag instead of the polyfill (from 3.20 until release of feature).

Sorry I wasn't very clear, the feature flag is enabled by default and will be included in ember-source >= 3.22.0-beta.1 when that is released (~ 3 or 4 weeks from now). For Ember itself, you cannot enable feature flags on any release _other_ than a canary build (this ensures that experimental / unstable features do not sneak into any tagged releases) which means you *can't* use the caching primitives in Ember until 3.22+.

So basically for 3.13 through 3.21, we should mention / use the polyfill; for 3.22+ we should remove mention of the polyfill and use the native API directly.

ijlee2:
@rwjblue Thanks for clarifying!

@YoranBrondsema May I review your current changes once more later today and make suggestions? Afterwards, you can apply the suggestions (feel free to suggest something else) and backport the text in `release` folder to `v3.15.0`-`v3.19.0` folders.

guides/release/in-depth-topics/autotracking-in-depth.md
```diff
@@ -321,4 +321,84 @@ class ShoppingList {
 }
 ```
 
+## Caching of tracked properties
+
+In contrast with computed properties from pre-Octane, autotracked properties
+are not cached. This means that an autotracked property will always be
+recomputed, regardless of whether its dependencies have changed. The following
+example shows how this works:
```
ijlee2, line 329: I think it's better that we don't use the term `autotracked property`, in case this may cause a reader to think it's different from `tracked property`.
```suggestion
In contrast to computed properties from pre-Octane, tracked properties are
not cached. A tracked property can also be recomputed even though its
dependencies haven't changed. The following example shows this behavior:
```

guides/release/in-depth-topics/autotracking-in-depth.md
```diff
@@ -321,4 +321,84 @@ class ShoppingList {
...
+
+console.log(person.fullName); // Jen Weber
+console.log(count); // 1;
+console.log(person.fullName); // Jen Weber
+console.log(count); // 2;
+
+person.firstName = 'Jennifer';
+
+console.log(person.fullName); // Jennifer Weber
+console.log(count); // 3;
```
ijlee2, line 356: To help address https://github.com/ember-learn/guides-source/issues/1480, let's not use a `Person` class. Can we use the following example instead?

```suggestion
class Photo {
  @tracked width = 600;
  @tracked height = 400;

  get aspectRatio() {
    count++;
    return this.width / this.height;
  }
}

let photo = new Photo();

console.log(photo.aspectRatio); // 1.5
console.log(count); // 1;
console.log(photo.aspectRatio); // 1.5
console.log(count); // 2;

photo.width = 800;

console.log(photo.aspectRatio); // 2
console.log(count); // 3;
```

guides/release/in-depth-topics/autotracking-in-depth.md
```diff
@@ -321,4 +321,84 @@ class ShoppingList {
...
+```
+
+This works fine in most cases. But sometimes you want a bit more control over
+the recomputation, for instance if the computation that happens in the getter
+is very expensive. In that case, you want to cache the value in between calls
+when the dependencies haven't changed. You want to recompute only if one of the
+dependencies has been modified.
```
ijlee2, line 363:
```suggestion
Recomputing is fine in most cases. If the computation that happens in the getter
is very expensive, however, you will want to cache the value and retrieve it when
the dependencies haven't changed. You want to recompute only if one of the
dependencies has been modified.
```

guides/release/in-depth-topics/autotracking-in-depth.md
```diff
@@ -321,4 +321,84 @@ class ShoppingList {
...
+
+The way to achieve this is through Ember's [cache
+API](https://github.com/emberjs/rfcs/blob/master/text/0615-autotracking-memoization.md).
+Using `createCache`, we can cache an autotracked property. The property can
+then be consumed with the `getValue` function. The way it works is as follows:
```
ijlee2, line 368: I added a link to the polyfill.

According to the [RFC](https://github.com/emberjs/rfcs/blob/master/text/0615-autotracking-memoization.md#usage), I think `createCache` receives a function. I updated your Guides text to make this clear:

```suggestion
Ember's [cache API](https://github.com/ember-polyfills/ember-cache-primitive-polyfill) lets you cache a getter in 2 steps.
First, you pass a function that is costly to compute to `createCache`.
Then, in the getter, you call the function with `getValue` and return its value.

With these 2 steps in mind, let's introduce caching to `aspectRatio`:
```

guides/release/in-depth-topics/autotracking-in-depth.md
```diff
@@ -321,4 +321,84 @@ class ShoppingList {
...
+
+console.log(person.fullName); // Jen Weber
+console.log(count); // 1;
+console.log(person.fullName); // Jen Weber
+console.log(count); // 1;
+
+person.firstName = 'Jennifer';
+
+console.log(person.fullName); // Jennifer Weber
+console.log(count); // 2;
```
ijlee2, line 401:
```suggestion
class Photo {
  @tracked width = 600;
  @tracked height = 400;

  // `#` indicates a private field on the class
  #aspectRatioCache = createCache(() => {
    count++;
    return this.width / this.height;
  })

  get aspectRatio() {
    return getValue(this.#aspectRatioCache);
  }
}

let photo = new Photo();

console.log(photo.aspectRatio); // 1.5
console.log(count); // 1;
console.log(photo.aspectRatio); // 1.5
console.log(count); // 1;

photo.width = 800;

console.log(photo.aspectRatio); // 2
console.log(count); // 2;
```

guides/release/in-depth-topics/autotracking-in-depth.md
```diff
@@ -321,4 +321,84 @@ class ShoppingList {
...
+```
+
```
ijlee2, line 403:
```suggestion

From the value of `count`, we see that, this time, `aspectRatio` was calculated only twice.
```

guides/release/in-depth-topics/autotracking-in-depth.md
YoranBrondsema: Makes sense.

guides/release/in-depth-topics/autotracking-in-depth.md
YoranBrondsema: Cool, agree.

guides/release/in-depth-topics/autotracking-in-depth.md
YoranBrondsema: Nice!

guides/release/in-depth-topics/autotracking-in-depth.md
YoranBrondsema: Good.

YoranBrondsema:
> For each paragraph, I made a suggestion that I think would be concise and easy to understand for someone who is new to caching getters or using Ember's cache API. You are welcome to apply the suggestions as they are, or leave a comment to ask questions or to suggest an alternative text.

I agree with your changes, nice.

> For code, I do need us to show an example that doesn't involve first and last names, to help address an active issue (#1480). I think the Photo class and calculating the aspect ratio (some math to stand for an expensive computation) would make a good example. Will this be okay with you?

Yes!

I updated the text following your suggestions. Once I get your OK, I'll:
- backport the file to 3.15 to 3.19
- squash this PR into a single commit

I think we'll be good to go after these two steps.

ijlee2:
@YoranBrondsema Thanks for reviewing my suggestions and updating the texts. I think they read great and we're ready to backport them to 3.15-3.19!

YoranBrondsema:
OK should be good now @ijlee2.

Issue #1237, opened 2019-12-24
Octane guides don't mention template `else if`/`else`


CodingItWrong:
Before Octane the Ember Guides referenced `{{#if}}`, `{{else if}}`, and `{{else}}` on the "Conditionals" page: https://guides.emberjs.com/v3.14.0/templates/conditionals/

As of the Octane Guides, the Conditional Content page only mentions `{{#if}}` — and `{{else if}}` and `{{else}}` are missing. I wasn't able to find the latter referenced anywhere in the guides.

I think it would be difficult for most Ember users who don't already know about `{{else if}}` and `{{else}}` to find out about them without them being on this guide page—I think they should be added back into the new form of the page.

chancancode:
Sounds good to me, I think we can add something back

ijlee2:
I'll make an update to https://guides.emberjs.com/release/components/conditional-content/ and backport the update all the way to 3.15.

Issue #1440, opened 2020-05-21
Fix API links in the Guides that go to "release"


jenweber:
When we released Octane, the API docs changed a whole bunch! Ideal Guides UX is that if someone is on an older version of Ember, like 3.12, if they follow the links to the API docs, they should see 3.12 docs (non-Octane). However, right now, a lot of links point directly to `api.emberjs.com/ember/release` and `api.emberjs.com/ember-data/release`.

To close this ticket, do a find-replace to change the "release" links to point to the appropriate version, for each version of the guides. The lastest version of the guides should still point to `release`, so don't change those.

ijlee2:
@jenweber @MelSumner Before asking for community help, I think we may need a concrete plan for how we'll update all links to the API Guides across all versions. To my understanding:

Currently, when there's a new Ember release, we copy-paste the `release` folder so that there will be a new folder for what used to be the release version. Take the upcoming 3.20 release as example. We'd copy-paste the `release` folder to create `3.19` folder.

Somehow, we'd need to find and replace all instances of "release" in the `3.19` folder to "3.19," and check that these links are all correct. Do we have a mechanism to automate this process?

As an alternative solution, would it be possible to send metadata to API Guides, so that the Ember Guides can always use `/release` in the links and the API Guides is responsible for redirecting from `/release` to the correct version?

jenweber:
Good questions! We don't have a mechanism to automate it, besides a careful find-replace. If you search the docs for `/release/` it will bring up all the places to change. We have a link checker that can be run on-demand: https://github.com/ember-learn/guides-source#internal-and-external-links If someone wants to take this on, I suggest fixing links for just one version of the guides and PR-ing that, as a warmup.

I would not want to make any changes to the API docs app, since its redirect rules are pretty complicated and the app is brittle.

There may be a way to build this into Guidemaker, but it's likely much faster to do a find-replace than code a new feature.

ijlee2:
@jenweber Sounds good. Thanks for the explanation! I can try find-and-replace on `3.18` and see how it goes.

ijlee2:
At the Learning Team meeting on July 30, we discussed updating the API links in the Ember Guides in all versions if possible. I made checklists below to divide work and track progress.

To ensure that search results in Google point to the `release` in API Guides, we'll want to look at the problem of addressing canonical links in the [API repo](https://github.com/ember-learn/ember-api-docs) soon.

## Phase 1
- [ ] `v3.19`
[...]

PR #1476, opened 2020-07-08
Explained how to write if-else statement in templates

ijlee2:
## Description

This PR addresses and can close the issue #1237. There are 3 main changes:

1\. I reviewed the current content in `/guides/release/components/conditional-content.md` and updated it so that I can introduce the `if-else` syntax in a natural manner. Afterwards, I backported (copy-pasted) the changes to v3.15-3.18.

2\. The `Block If` and `Inline If` now have subsections called `Motivation` and `Usage`. I think making subsections helps a reader take a break (we don't overload them by showing text after text). People can also link to `Usage` for a quick access to the block and inline syntax.

3\. I created Learn More section at the end. For v3.15-3.18, I ensured that the redirect link to the API docs includes the correct version number. This addresses the issue that was brought up in #1440.

## How to review

I recommend taking one of 2 approaches:

- Review commits one at a time. Each commits corresponds to updating a section in `conditional-content.md`. The end result is the same for v3.15-release, so feel free to look at only the file in the `release` directory.
- For more visual review, check the [preview deploy site](https://deploy-preview-1476--ember-guides.netlify.app/release/components/conditional-content/) or [Percy snapshot](https://percy.io/Ember/guides-app/builds/6058156/view/350018563/1280?mode=diff&browser=firefox&snapshot=350018563).

### DIFF
--- a/guides/release/components/conditional-content.md
+++ b/guides/release/components/conditional-content.md
@@ -1,6 +1,4 @@
-You can conditionally include some content in the output of a template by using
-Ember's `if` syntax in templates.
-There are two styles of `if`, block and inline:
+In a template, you can use `if` to conditionally render content. There are 2 styles of `if`: **block** and **inline**.
 
 ```handlebars
 {{#if this.thingIsTrue}}
@@ -11,6 +9,7 @@ There are two styles of `if`, block and inline:
   This div used the inline "if" to calculate the class to use.
 </div>
 ```
+
 
 ## Block `if`
 
--- a/guides/release/components/conditional-content.md
+++ b/guides/release/components/conditional-content.md
@@ -195,4 +195,6 @@ Ember also allows you to write an `if-else` statement in inline form. It looks s
 ```
 
 
-Refer the [API documentation of the `if` helper](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/if?anchor=if) for more patterns.
+## Learn More
+
+Please see the [API documentation of the `if` helper](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/if?anchor=if) for more patterns.
--- a/guides/release/components/conditional-content.md
+++ b/guides/release/components/conditional-content.md
@@ -13,7 +13,9 @@ In a template, you can use `if` to conditionally render content. There are 2 sty
 
 ## Block `if`
 
-Let's take a look at two similar components representing a user's username.
+### Motivation
+
+Let's take a look at two components that display a person's username.
 
 ```handlebars {data-filename="app/components/received-message/username.hbs"}
 <h4 class="username">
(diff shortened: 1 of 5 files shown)


guides/release/components/conditional-content.md
```diff
@@ -54,52 +54,88 @@ We can do this with an `if`:
...
+    Z
+  </div>
 </aside>
 ```
 
-The first component needs to have an `is-active` class applied to the avatar to
-show the active icon. That icon could also change over time, so it probably
-doesn't make sense to try to use `...attributes` to apply it. It's a part of the
-_public API_ of the component, so an argument makes sense.
+Again, the two components look similar. The first component has an `is-active` class, while the second a `current-user` class. How should we unify the components into one?
```
locks, line 132:
```suggestion
Again, the two components look similar.
The first component has an `is-active` class, while the second a `current-user` class.
How should we unify the components into one?
```

guides/release/components/conditional-content.md
```diff
@@ -54,52 +54,88 @@ We can do this with an `if`:
...
 
-So, we want to add the `is-active` class if an argument, like say `@isActive`,
-is passed in and is truthy.
+The `is-active` class is responsible for showing the active icon. _How_ that icon is rendered may change over time, so we won't use `...attributes` to apply the `is-active` class. Instead, we'll pass the argument `@isActive` to dictate _what_ to do (e.g. render the icon).
```
locks, line 134:
```suggestion
The `is-active` class is responsible for showing the active icon.
_How_ that icon is rendered may change over time, so we won't use `...attributes` to apply the `is-active` class.
Instead, we'll pass the argument `@isActive` to dictate _what_ to do (e.g. render the icon).
```

guides/release/components/conditional-content.md
```diff
@@ -54,52 +54,88 @@ We can do this with an `if`:
...
+
+As for the `current-user` class, it may have been just one of a few classes that can be applied to the `<aside>` element. Let's use `...attributes` to apply the `current-user` class.
```
locks, line 136:
```suggestion
As for the `current-user` class, it may have been just one of a few classes that can be applied to the `<aside>` element.
Let's use `...attributes` to apply the `current-user` class.
```

guides/release/components/conditional-content.md
```diff
@@ -54,52 +54,88 @@ We can do this with an `if`:
...
+
+We take these API designs into account and end up with a reusable component. The component uses an inline `if` to conditionally apply the `is-active` class.
```
locks, line 138:
```suggestion
We take these API designs into account and end up with a reusable component.
The component uses an inline `if` to conditionally apply the `is-active` class.
```

ijlee2:
Ah, ok. I'll work on breaking lines later today. Thanks for the feedback!

ijlee2:
@locks In addition to line breaks, I applied the suggestions that you made (e.g. remove hyphen in `if-else`). When you have time, can you have a look at the last commit?

PR #1494, opened 2020-07-28
Updated API links in guides/3.18.0

ijlee2:
## Description

The 1st part of a series of PRs for #1440. Don't let the number of files changed scare you! 👻

(Subsequent PRs will have fewer files changed. I opted to fix incorrect links now rather than wait, especially because there's an imminent 3.20 Guides release.)

## How to review

I recommend reviewing the commits one at a time.

In the first 2 commits, I used find-and-replace inside the `/guides/v3.18.0` folder to replace all instances of `release` to `3.18` in URLs for the API Guides.

I did temporarily modify `/node-tests/local/external-link-checker.js` to test external links in `/guides/v3.18.0`. I'll make a separate PR that addresses a couple of issues that I found.

However, the primary way of checking external links was a manual check. I opened every link to API Guides to ensure that we redirect a reader to the appropriate content.

In the subsequent 9 commits, I fixed URLs to be "better." (The anchor locations were technically correct, but the URLs showed a mismatch, which might cause an issue later.) Each commit looks at 1 particular file across all affected versions.

### DIFF
--- a/guides/v3.18.0/components/built-in-components.md
+++ b/guides/v3.18.0/components/built-in-components.md
@@ -1,10 +1,10 @@
 Ember provides a few helpful components out-of-the-box for common tasks,
 including:
 
-* [`<Input />`](https://api.emberjs.com/ember/3.18/classes/Ember.Templates.helpers/methods/if?anchor=input)
+* [`<Input />`](https://api.emberjs.com/ember/3.18/classes/Ember.Templates.components/methods/Input?anchor=Input)
 * [`<Textarea />`](https://api.emberjs.com/ember/3.18/classes/Ember.Templates.components/methods/Textarea?anchor=Textarea)
 
-Using these helpers, you can create form controls that are almost identical to
+Using these components, you can create form controls that are almost identical to
 the native HTML `<input>` or `<textarea>` elements, and which automatically
 update the state of their values.
 
@@ -77,8 +77,8 @@ More [events types](https://api.emberjs.com/ember/3.18/classes/Component#event-n
 ### Checkboxes
 
 You can also use the
-[`<Input />`](https://api.emberjs.com/ember/3.18/classes/Ember.Templates.helpers/methods/if?anchor=input)
-helper to create a checkbox by setting its `@type`:
+[`<Input />`](https://api.emberjs.com/ember/3.18/classes/Ember.Templates.components/methods/Input?anchor=Input)
+component to create a checkbox by setting its `@type`:
 
 ```handlebars
 <label for="admin-checkbox">Is Admin?</label>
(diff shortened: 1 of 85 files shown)


guides/v3.18.0/components/built-in-components.md
```diff
@@ -72,13 +72,13 @@ The following event types are supported (dasherized format):
 * `@key-up`
 
 
-More [events types](https://api.emberjs.com/ember/release/classes/Component#event-names) are also supported but these events need to be written in camelCase format, such `mouseEnter`. Note, there are events of the same type in both the list above and linked. Event names listed above must be dasherized. Additional work is performed on these events.
+More [events types](https://api.emberjs.com/ember/3.18/classes/Component#event-names) are also supported but these events need to be written in camelCase format, such `mouseEnter`. Note, there are events of the same type in both the list above and linked. Event names listed above must be dasherized. Additional work is performed on these events.
```
ijlee2, line 75: The `#event-names` anchor doesn't exist. I wasn't sure what'd be a good substitute.

ijlee2:
Only 9 Percy snapshots should have been marked different. (These correspond to 9 files changed in `/guides/release`.)

I think Percy may have showed many snapshots as new possibly due to the merge of https://github.com/ember-learn/guides-source/pull/1462. We can probably safely ignore new snapshots.

guides/v3.18.0/components/built-in-components.md
MelSumner: I see what you mean, although I think we should still link to /release/. 

I think the missing anchor tag is for the "Event Handler Methods" - is there a way to add the anchor in to the correct place?

ijlee2: Hmm, I think the version would be `3.18` according to what the issue asked for (point API links in 3.18 Guides to 3.18 API Guides?).

I think you're right that the anchor can point to `https://api.emberjs.com/ember/3.18/classes/Component#event-handler-methods` instead, since that section lists examples of events.

ijlee2: I found that, prior to `v3.10`, the API Guides used to have a section called "Event Names." Since `v3.10`, that section was renamed to "Event Handler Methods."

- https://api.emberjs.com/ember/3.9/classes/Component#event-names
- https://api.emberjs.com/ember/3.10/classes/Component#event-handler-methods

I pushed a commit to update the links for v3.7 - 3.19.

jenweber:
It looks like CI is doing ok this time! We can merge this in before the release.

ijlee2:
@jenweber Thanks for the review! I'll merge this then. Let me know if you need help with recreating #1496.

PR #1495, opened 2020-07-29
Added missing .jumbo class for consistency's sake

RichardOtvos:
In all the other assertions for the other pages we also use the ´.jumbo´ class. While I understand this still works correctly for the tests in the tutorial, I feel like it unecessaryily confusing for beginners (especially when it's not called out in the text after teh example either).


jenweber:
Hi @RichardOtvos, thanks for catching this! To make changes to the Tutorial sections, could you open a PR in the super-rentals-tutorial repository instead? https://github.com/ember-learn/super-rentals-tutorial

This guides-source repository contains a copy of the tutorial repository, and the copy is overwritten every time there's a change in the tutorial repo. Thanks!

RichardOtvos:
No problem! [Link for posterity](https://github.com/ember-learn/super-rentals-tutorial/pull/154)

PR #1500, opened 2020-07-30
Re-roll package-lock.json

### LINKED ISSUE #1498: Can only run search index using Node 10
When I was using Node 12, I was not able to run `ember deploy production` or the release script due to 

```
- build failed
Error: Cannot find module '@babel/compat-data/corejs3-shipped-proposals'
Require stack:
- /Users/jweber/projects/guides-source/node_modules/guidemaker/node_modules/@babel/preset-env/lib/polyfills/corejs3/usage-plugin.js
```

This resolved when I switched to Node 10.

rwjblue:
Fixes #1498


rwjblue:
`lint:md` is failing (the packages we use to do it drifted)

ijlee2:
A couple of mobile Percy snapshots showed a difference in how, perhaps, a long text in `<code>` element is wrapped around. I think it's a non-issue, though, so I approved these snapshots.

Screen Shot 2020-07-30 at 4 46 26 PM

ijlee2:
@rwjblue Can you clarify what you mean by "drifted"? I didn't quite follow the terminology.

It seemed like the Travis CI build failed because the word "syntaxes," which had been used in several places, is now marked as misspelled. Maybe we can add this word to `.local.dic` so that the build may pass?

rwjblue:
> Can you clarify what you mean by "drifted"?

Ya, I'm specifically talking about semver drift. Since we depend on `remark@^8.0.0` we technically allow any version from 8.0.0 until (some future) 9.x. This PR totally recreates `package-lock.json` so we end up changing from using that version to now using `8.0.1`. The same is true for the other `remark` related packages in our package.json:

https://github.com/ember-learn/guides-source/blob/d014e773bc871c4d1c591e3275d2f6f19a759a44/package.json#L86-L90

I'm not sure which package brings the default dictionary, but my point was that whatever package it is has changed and now we need to add that override.

rwjblue:
> Maybe we can add this word to .local.dic so that the build may pass?

Done!

ijlee2:
I tested the preview site and checked basic functionalities. (On Netlify, I did have to set Node version to `v12.18.3` so that it wouldn't default to `v8.17.0` and cause build error.)

The site was still mobile responsive. I also checked that one can navigate from one section to another, open and close the navigation menu, and use Algolia search.

PR #1502, opened 2020-07-31
Fixed a broken link

ijlee2:
## Description

When I tested the use of `/node-tests/local/external-link-checker.js` on the `/guides/v3.18.0` folder, I found that:

- The link to the article "[How to Use Application Role](https://www.a11yproject.com/posts/2013-02-09-how-to-use-application-role/)" changed over time.
- The script incorrectly flagged `http://localhost:4200/tests` as a broken link.

I corrected the link to a11yproject and updated the script to ignore all URLs that begin with `http://localhost:4200`.

### DIFF
--- a/node-tests/local/external-link-checker.js
+++ b/node-tests/local/external-link-checker.js
@@ -9,11 +9,11 @@ const { extname } = require('path');
 const { inspect } = require('util');
 const {
   findMarkdownLinks,
-	checkExternalLink,
-	mapToLocalUrl,
-	removeTrailingApostrophe,
-	randomizeFetchDelays,
-	sleep
+  checkExternalLink,
+  mapToLocalUrl,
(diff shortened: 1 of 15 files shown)


node-tests/local/external-link-checker.js
```diff
@@ -9,11 +9,11 @@ const { extname } = require('path');
 const { inspect } = require('util');
 const {
   findMarkdownLinks,
-	checkExternalLink,
-	mapToLocalUrl,
-	removeTrailingApostrophe,
-	randomizeFetchDelays,
-	sleep
+  checkExternalLink,
```
ijlee2, line 12: The diff noise that you see here is due to mixing spaces with tabs and mixing 2 and 4 spaces for an indentation.

If you'd like you can look at only the 2nd and 3rd commits for changes described in the PR.

PR #1506, opened 2020-08-04
Updated API links in guides/3.17.0

ijlee2:
## Description

This PR addresses https://github.com/ember-learn/guides-source/issues/1440 for Ember `v3.17`.

I used Find All and Replace to update the URLs. Afterwards, I ran `npm run test:node-local` to check that we can visit the updated URLs.

### DIFF
--- a/guides/v3.17.0/applications/dependency-injection.md
+++ b/guides/v3.17.0/applications/dependency-injection.md
@@ -11,11 +11,11 @@ guide covers the details of the system, and how to use it when needed.
 
 Applications and application instances each serve a role in Ember's DI implementation.
 
-An [`Application`](https://api.emberjs.com/ember/release/classes/Application) serves as a "registry" for dependency declarations.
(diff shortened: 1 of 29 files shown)


ijlee2:
Much appreciate your help with reviewing multiple PRs!

It's possible to run the `test:node-local` script in CI. I think the value would be there if we check all external links in more than 1 version, at a scheduled interval (e.g. every week) instead of for every PR.

However, the script currently takes a long time to run locally (~10 min? for 1 version). One reason is we're checking if a visit to an API Guides page was successful by rendering the whole page (I think). It'd be nice if there was a way to get the success status code and don't render the page. Another reason is we introduce artifical delays between tests (visiting a different page), I think, so that we don't hit a rate limit.

amyrlam:
@ijlee2 Aha, thanks for the context, that makes sense now. I'll have to try it out next PR

PR #1509, opened 2020-08-05
Rewrite `ref` example using custom modifier

chriskrycho:
Per discussion in the [June 2020 Learning Team meeting][minutes], remove the use of `ember-ref-modifier` in favor of a custom modifier implemented using `ember-modifier`, since a custom modifier here is more idiomatic, performs better, and is more maintainable.

[minutes]: https://github.com/emberjs/core-notes/blob/259e63c328af76afff9bea4523290a57557d905d/learning-team/2020-06/2020-06-04.md

Note that this is *not* a criticism of @lifeart's original implementation of `ember-ref-modifier`: it was a very valuable early experiment with modifiers, and filled a gap when `ember-modifier` didn't even exist yet. It was precisely in experimenting with modifiers like `ref` that we figured out the patterns we now want to suggest instead!

### DIFF
--- a/guides/release/components/template-lifecycle-dom-and-modifiers.md
+++ b/guides/release/components/template-lifecycle-dom-and-modifiers.md
@@ -310,9 +310,7 @@ Now we can use our custom `{{autofocus}}` modifier throughout our application.
 
 ## Communicating Between Elements in a Component
 
-What if you want to handle an event in one part of your component by calling a DOM method on another part? For example, let's say you're creating an audio component, and you want clicking the "Play" button to call the audio tag's `play` method.
-
-Let's start with the HTML we're working with:
+What if you want to handle an event in one part of your component by calling a DOM method on another part? For example, let's say you're creating an audio component:
 
 ```handlebars {data-filename="app/components/audio-player.hbs"}
 <audio src={{@srcURL}} />
@@ -321,17 +319,21 @@ Let's start with the HTML we're working with:
 <button type="button">Pause</button>
 ```
 
-It might be tempting to define all interactions in the component class. However, as we will see soon, the key to a better solution is to separate concerns. Let the component manage the state and the modifier manage the DOM.
+How should we make it so that clicking the "Play" and "Pause" buttons to call the audio tag's `play` and `pause` methods?
 
-In general, there are 3 reasons why we might want to introduce a modifier:
+While we *could* manage these DOM interactions in the component class (for example, by using `{{did-render}}`), we're better off using a modifier here. It lets us cleanly separate our concerns: the component manages the *state*, and the modifier manages *interactions with the DOM*.
 
-1. Components don't have access to DOM elements directly. We'd have to render the page, push an element back up into the component class, and *then* wire it up. This would come with performance costs, because we would have to render twice before anything could work.
-2. We can't really use autotracking or 1-way data flow that way. If we wanted the caller of our component to be able to pass an argument in and have the component respond appropriately, we'd have to use something like `{{did-update}}`.
-3. None of the logic we built to manage playing and pausing would be reusable! It would be specific to just this component.
+There are three reasons we should keep state management in a component and DOM element management in a modifier:
 
-Using a custom modifier, we can avoid all of these issues!
+1. A component, by itself, doesn't have direct access to DOM elements. We have to render the page, push an element back up into the component class, and only then can we safely refer to that element. This can sometimes require us to render the component's HTML twice in order for things to start working. Modifiers let us avoid this possible performance issue.
 
-First, let's add event handlers to the `Play` and `Pause` buttons:
+2. By keeping state in the component and handling DOM method calls in a modifier, we can use autotracking and practice 1-way data flow. We could change the component's own design later *without* having to change how we interact with the DOM element.
+
+3. The code for calling the audio element's `play` and `pause` can be reused. It isn't tied to this particular audio component. It can be tested independently, too!
+
+Now that we see *why* to use a modifier for our audio component, let's walk through *how* to create one. We will start with the component (to manage the state) and then implement the modifier (the manage the DOM).
+
+First, we add actions to handle the `click` events for the `Play` and `Pause` buttons:
 
 ```handlebars {data-filename="app/components/audio-player.hbs" data-diff="-3,+4,-5,+6"}
 <audio src={{@srcURL}} />
--- a/guides/release/components/template-lifecycle-dom-and-modifiers.md
+++ b/guides/release/components/template-lifecycle-dom-and-modifiers.md
@@ -392,7 +392,7 @@ ember install ember-modifier
 ember generate modifier play-when
 ```
 
-In this modifier, we'll take an argument that specifies whether we should be playing or paused.
+The modifier takes 1 argument, a Boolean that specifies if we should call the element's `play` or `pause` method.
 
 ```js {data-filename="app/modifiers/play-when.js"}
 import { modifier } from "ember-modifier";
--- a/guides/release/components/template-lifecycle-dom-and-modifiers.md
+++ b/guides/release/components/template-lifecycle-dom-and-modifiers.md
@@ -319,7 +319,7 @@ What if you want to handle an event in one part of your component by calling a D
 <button type="button">Pause</button>
 ```
 
-How should we make it so that clicking the "Play" and "Pause" buttons to call the audio tag's `play` and `pause` methods?
+How should we connect clicking the "Play" and "Pause" to calling the audio tag's `play` and `pause` methods?
 
 While we *could* manage these DOM interactions in the component class (for example, by using `{{did-render}}`), we're better off using a modifier here. It lets us cleanly separate our concerns: the component manages the *state*, and the modifier manages *interactions with the DOM*.
 
--- a/guides/release/components/template-lifecycle-dom-and-modifiers.md
+++ b/guides/release/components/template-lifecycle-dom-and-modifiers.md
@@ -418,7 +418,7 @@ With that, we can now click the buttons to play and pause the audio!
 
 In summary, when you want to allow elements in a component to communicate, see if you can separate the concerns of _managing state_ and _managing DOM interactions_. The component can manage the state, while a modifier can manage the DOM.
 
-The modifier that we made for the audio player component can be reused on _any_ element that implements `play` and `pause` methods. In particular, we can reuse the modifier on an `HTMLMediaElement`, which includes `audio` and `video` elements.
+The modifier that we made for the audio player component can be reused on _any_ element that implements `play` and `pause` methods. In particular, we can reuse the modifier on any `HTMLMediaElement`, which includes `audio` and `video` elements.
 
 ## Out-of-Component Modifications
 
--- a/guides/release/components/template-lifecycle-dom-and-modifiers.md
+++ b/guides/release/components/template-lifecycle-dom-and-modifiers.md
@@ -331,7 +331,7 @@ There are three reasons to reach for modifiers for DOM element interactions:
 
 3. The code for calling the audio element's `play` and `pause` can be reused. It isn't tied to this particular audio component. It can be tested independently, too!
 
-Now that we see *why* to use a modifier for our audio component, let's walk through *how* to create one. We will start with the component (to manage the state) and then implement the modifier (the manage the DOM).
+Now that we see *why* we want to use a modifier for our audio component, let's walk through *how* to create one. We will start with the component (to manage the state) and then implement the modifier (the manage the DOM).
 
 First, we add actions to handle the `click` events for the `Play` and `Pause` buttons:
 
--- a/guides/release/components/template-lifecycle-dom-and-modifiers.md
+++ b/guides/release/components/template-lifecycle-dom-and-modifiers.md
@@ -321,14 +321,23 @@ Let's start with the HTML we're working with:
 <button type="button">Pause</button>
 ```
 
-Next, let's add an event handler to the `Play` button:
+It might be tempting to try to handle this by managing the whole set of interactions at the component level. But there are three reasons we might want to tackle this using a modifier instead:
 
-```handlebars {data-filename="app/components/audio-player.hbs" data-diff="-3,+4"}
+1. Components don't have access to DOM elements directly. We'd have to render the page, push an element back up into the component class, and *then* wire it up. This would come with performance costs, because we would have to render twice before anything could work.
+2. We can't really use autotracking or 1-way data flow that way. If we wanted the caller of our component to be able to pass an argument in and have the component respond appropriately, we'd have to use something like `{{did-update}}`.
+3. None of the logic we built to manage playing and pausing would be reusable! It would be specific to just this component.
+
+Using a custom modifier, we can avoid all of these issues!
+
+First, let's add event handlers to the `Play` and `Pause` buttons:
+
+```handlebars {data-filename="app/components/audio-player.hbs" data-diff="-3,+4,-5,+6"}
 <audio src={{@srcURL}} />
 
 <button type="button">Play</button>
 <button type="button" {{on "click" this.play}}>Play</button>
 <button type="button">Pause</button>
+<button type="button" {{on "click" this.pause}}>Pause</button>
 ```
 
 ```js {data-filename="app/components/audio-player.js"}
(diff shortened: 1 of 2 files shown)


guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -321,14 +321,23 @@ Let's start with the HTML we're working with:
 <button type="button">Pause</button>
 ```
 
-Next, let's add an event handler to the `Play` button:
+It might be tempting to try to handle this by managing the whole set of interactions at the component level. But there are three reasons we might want to tackle this using a modifier instead:
 
-```handlebars {data-filename="app/components/audio-player.hbs" data-diff="-3,+4"}
+1. Components don't have access to DOM elements directly. We'd have to render the page, push an element back up into the component class, and *then* wire it up. This would come with performance costs, because we would have to render twice before anything could work. It would also
```
rwjblue, line 326: This trails off oddly:

> It would also

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -340,40 +349,71 @@ export default class AudioPlayerComponent extends Component {
...
 ```
 
+Now the component is managing the state of the player, but we still need to translate that into the DOM method calls. So let's create a modifier!
+
+```bash
+ember install ember-modifier
+ember generate modifier play-when
+```
+
+In this modifier, we'll simply take an argument that specifies whether we should be playing or paused.
```
rwjblue, line 391: I tend to prefer to avoid "simply" and "just" as a bit condescending, maybe its fine without?

```suggestion
In this modifier, we'll take an argument that specifies whether we should be playing or paused.
```

lifeart:
@chriskrycho @rwjblue I have an new way to access dom https://github.com/lifeart/ember-ref-bucket#link-div-to-node-property with kinda "similar" api, but without cons of original `ref-modifier`, could you take a look? (wondering, if it make sence to use, should we mention it)

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -340,40 +349,71 @@ export default class AudioPlayerComponent extends Component {
...
+});
+```
+
+We can now attach the modifier to the `audio` element:
+
+```handlebars {data-filename="app/components/audio-player.hbs" data-diff="-1,+2"}
+<audio src={{@srcURL}} />
+<audio src={{@srcURL}} {{play-when this.isPlaying}} />
+
+<button type="button" {{on "click" this.play}}>Play</button>
```
lifeart, line 411: wondering, is it real use-case? (looks like we could remove `play-when` modifier and handle play logic in actions at all)

lifeart: usecase from our project:

```js
export default modifier(function scrollTop(element, [enabled]) {
  if (enabled) {
    element.scrollIntoView({behavior: "smooth", block: "start", inline: "nearest"});
  }
});
```

rwjblue:
> but without cons of original ref-modifier, could you take a look? (wondering, if it make sence to use, should we mention it)

This definitely looks like a nice bit of invocation simplification, but I don't see how this avoids requiring two rendering passes generally speaking (the demo here in the guide didn't have that issue, but a decent amount of the usages I have seen in apps have).

Also, some other things about the new prose here:

* it teaches folks how to _think_ about the modifier case
* it begins down the (long) road towards making a more official modifier base class / function invocation etc

Ultimately, I think this approach is the right path forward.

guides/release/components/template-lifecycle-dom-and-modifiers.md
lifeart: or we could add more logic to `play-when`, to toggle `isPlaying` once audio will be finished (to get it close to real-life)

chriskrycho: A couple notes here—

1. This is in fact a real-world use case (as well as the example used originally here). The implementation I put up is *very* similar to one we designed for use in our own app.
2. My take on maintainability is a little different, I think: maintaining a very simple modifier and its tests is actually a big *win*, because it only has one thing it does and can focus on doing it well.

guides/release/components/template-lifecycle-dom-and-modifiers.md
chriskrycho: Fixed via a force push. Thank you!

lifeart:
This definitely looks like a nice bit of invocation simplification, but I don't see how this avoids requiring two rendering passes generally speaking (the demo here in the guide didn't have that issue, but a decent amount of the usages I have seen in apps have).

- it does not schedule anything, and non-tracked calculations are `lazy` if we don't reuse ref in dom.

guides/release/components/template-lifecycle-dom-and-modifiers.md
chriskrycho: Heh, I turned this into the next bullet point and forgot to remove it. Thank you!

rwjblue:
> it does not schedule anything, and non-tracked calculations are lazy if we don't reuse ref in dom.

Sure, but if you use a getter that uses the (newly stored off ref): boom, a double render is required

lifeart:
>> Sure, but if you use a getter that uses the (newly stored off ref): boom, a double render is required

nope, if it not tracked `trackedRef`

rwjblue:
> nope, if it not tracked (trackedRef)

I don't see how that is possible. If a property is accessed before it is set, then set later, it will require two render passes to work. Alternatively, if some part of the template _after_ the `{{ref` invocation attempts to use the ref (generally by way of a getter) it will also require two renders (because appending DOM happens _before_ the `installModifier` hook is called on modifiers). Happy to take that discussion elsewhere though.

lifeart:
if we reuse it in dom - yes.
if we use it in event handlers (inside component logic) it will not cause rerenders.

And @chriskrycho example may looks like:

```hbs
<audio {{ref "audio" src={{@srcURL}} />
<button type="button" {{on "click" this.play}}>Play</button>
<button type="button" {{on "click" this.pause}}>Pause</button>
```
```js
export default class AudioPlayerComponent extends Component {
  @ref('audio') audioElement = null;
  @action
  play() {
    this.audioElement.play();
  }
  @action
  pause() {
    this.audioElement.pause();
  }
}
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -321,14 +321,23 @@ Let's start with the HTML we're working with:
 <button type="button">Pause</button>
 ```
 
-Next, let's add an event handler to the `Play` button:
+It might be tempting to try to handle this by managing the whole set of interactions at the component level. But there are three reasons we might want to tackle this using a modifier instead:
```
ijlee2, line 324: Initially, I found the 3 reasons to be confusing (a bit too abstract). It wasn't until I reached the end of this subsection, where you wrote the tagline,

```
The component manages the state. The modifier manages the DOM.
```

that I understood what you wanted readers to do.

I think it'd be good to include these lines in the beginning, then reiterate them at the end.

```suggestion
It might be tempting to define all interactions in the component class. However, as we will see soon, the key to a better solution is to separate concerns. Let the component manage the state and the modifier manage the DOM.

In general, there are 3 reasons why we might want to introduce a modifier:
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -321,14 +321,23 @@ Let's start with the HTML we're working with:
...
 
-```handlebars {data-filename="app/components/audio-player.hbs" data-diff="-3,+4"}
+1. Components don't have access to DOM elements directly. We'd have to render the page, push an element back up into the component class, and *then* wire it up. This would come with performance costs, because we would have to render twice before anything could work.
+2. We can't really use autotracking or 1-way data flow that way. If we wanted the caller of our component to be able to pass an argument in and have the component respond appropriately, we'd have to use something like `{{did-update}}`.
+3. None of the logic we built to manage playing and pausing would be reusable! It would be specific to just this component.
```
ijlee2, line 328:
```suggestion
1. Components don't have direct access to DOM elements. We have to render the page, push an element into the component class, then wire it up. These steps can cause a performance issue because we have to render the page twice in order for things to start working.
2. We can't use autotracking and one-way data flow. If we want the caller of a component to be able to pass an argument and have the component respond appropriately, we will have to use a solution like `{{did-update}}`.
3. The code for managing play and pause is not reusable. It is tied to the audio component.
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -321,14 +321,23 @@ Let's start with the HTML we're working with:
...
+
+Using a custom modifier, we can avoid all of these issues!
+
+First, let's add event handlers to the `Play` and `Pause` buttons:
```
ijlee2, line 332:
```suggestion
With these reasons in mind, let's look at how to create a modifier for our audio component. We will arrive at the final solution in incremental steps.

First, we add event handlers to the `Play` and `Pause` buttons:
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -340,40 +349,71 @@ export default class AudioPlayerComponent extends Component {
...
-<button type="button">Pause</button>
+  @action
+  pause() {
+    // TODO
+  }
+}
 ```
 
-Now, the component can access the audio element in the `play` method:
+From the component's point of view, all it needs to do is track whether it should be playing or not, since we're going to manage the DOM element with the modifier. So we can just use `@tracked` and a normal property here.
```
ijlee2, line 360: Let's reiterate our point (separate concerns) for the 2nd time. 🙂

```suggestion
Recall that our modifier will manage the DOM (i.e. calling the audio element's `play` or `pause` method). All the component needs to do is to track if the audio is playing.
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -340,40 +349,71 @@ export default class AudioPlayerComponent extends Component {
...
+
+  @action
+  pause() {
+    // TODO
+    this.isPlaying = false;
   }
 }
 ```
 
+Now the component is managing the state of the player, but we still need to translate that into the DOM method calls. So let's create a modifier!
```
ijlee2, line 384:
```suggestion
Notice that the component manages the state of the audio player. We still need to translate the state into the appropriate DOM method call. Let's create the modifier now!
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -340,40 +349,71 @@ export default class AudioPlayerComponent extends Component {
...
+
+```bash
+ember install ember-modifier
+ember generate modifier play-when
+```
+
+In this modifier, we'll take an argument that specifies whether we should be playing or paused.
```
ijlee2, line 391:
```suggestion
The modifier takes 1 argument, a Boolean that specifies if we should call the element's `play` or `pause` method.
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -340,40 +349,71 @@ export default class AudioPlayerComponent extends Component {
...
+export default modifier((element, [isPlaying]) => {
+  if (isPlaying) {
+    element.play();
+  } else {
+    element.pause();
+  }
+});
+```
+
+We can now attach the modifier to the `audio` element:
```
ijlee2, line 405:
```suggestion
Last but not least, we attach the modifier to the `audio` element:
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -340,40 +349,71 @@ export default class AudioPlayerComponent extends Component {
...
+
+```handlebars {data-filename="app/components/audio-player.hbs" data-diff="-1,+2"}
+<audio src={{@srcURL}} />
+<audio src={{@srcURL}} {{play-when this.isPlaying}} />
+
+<button type="button" {{on "click" this.play}}>Play</button>
+<button type="button" {{on "click" this.pause}}>Pause</button>
+```
+
+With that, we're done! Notice that we've separated the concerns of *managing state* from *managing DOM interactions*. The component manages the state. The modifier manages the DOM. And this modifier could be reused for *any* `HTMLMediaElement`, including other `audio` or `video` element in our code.
```
ijlee2, line 415:
```suggestion
With that, we can now click the buttons to play and pause the audio!

In summary, when you want to allow elements in a component to communicate, see if you can separate the concerns of *managing state* and *managing DOM interactions*. The component can manage the state, while a modifier can manage the DOM.

The modifier that we made for the audio player component can be reused on *any* element that implements `play` and `pause` methods. In particular, we can reuse the modifier on an `HTMLMediaElement`, which includes `audio` and `video` elements.
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
chriskrycho: Good callout! Committed that suggestion directly.

chriskrycho: As I'm mulling, I'm going to push a slight tweak here, but the overall direction I'm still very much 👍 on.

guides/release/components/template-lifecycle-dom-and-modifiers.md
chriskrycho: Took this idea and tweaked it a bit; see 414d25ee.

guides/release/components/template-lifecycle-dom-and-modifiers.md
chriskrycho: This is helpful! I went with some slightly different wording and structure here, while trying to get at these goals; see 414d25ee.

guides/release/components/template-lifecycle-dom-and-modifiers.md
chriskrycho: Did a slight variant on this in b1da6ccd!

guides/release/components/template-lifecycle-dom-and-modifiers.md
chriskrycho: I *really* liked this; it makes the takeaway much clearer. Thank you!

guides/release/components/template-lifecycle-dom-and-modifiers.md
chriskrycho: Much clearer!

chriskrycho:
@ijlee2 thanks for the great feedback! I addressed it all—committed a bunch of your suggestions directly, and in a couple cases iterated yet further with the same concepts in mind. Let me know if you see anything else!

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -321,15 +319,21 @@ Let's start with the HTML we're working with:
...
 ```
 
-It might be tempting to try to handle this by managing the whole set of interactions at the component level. But there are three reasons we might want to tackle this using a modifier instead:
+How should we connect clicking the "Play" and "Pause" to calling the audio tag's `play` and `pause` methods?
+
+While we *could* manage these DOM interactions in the component class (for example, by using `{{did-render}}`), we're better off using a modifier here. It lets us cleanly separate our concerns: the component manages the *state*, and the modifier manages *interactions with the DOM*.
+
+There are three reasons to reach for modifiers for DOM element interactions:
+
+1. A component, by itself, doesn't have direct access to DOM elements. We have to render the page, push an element back up into the component class, and only then can we safely refer to that element. This can sometimes require us to render the component's HTML twice in order for things to start working. Modifiers let us avoid this possible performance issue.
```
ijlee2, line 328: Thanks for clarifying what you meant by "wire it up". 💯

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -321,15 +319,21 @@ Let's start with the HTML we're working with:
...
-2. We can't really use autotracking or 1-way data flow that way. If we wanted the caller of our component to be able to pass an argument in and have the component respond appropriately, we'd have to use something like `{{did-update}}`.
-3. None of the logic we built to manage playing and pausing would be reusable! It would be specific to just this component.
+2. By keeping state in the component and handling DOM method calls in a modifier, we can use autotracking and stick to 1-way data flow in the component. Further, we could change the component's own design later *without* having to change how we interact with the DOM element.
 
-Using a custom modifier, we can avoid all of these issues!
+3. The code for calling the audio element's `play` and `pause` can be reused. It isn't tied to this particular audio component. It can be tested independently, too!
 
-First, let's add event handlers to the `Play` and `Pause` buttons:
+Now that we see *why* to use a modifier for our audio component, let's walk through *how* to create one. We will start with the component (to manage the state) and then implement the modifier (the manage the DOM).
```
ijlee2, line 334:
```suggestion
Now that we see *why* we want to use a modifier for our audio component, let's walk through *how* to create one. We will start with the component (to manage the state) and then implement the modifier (the manage the DOM).
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -381,14 +385,14 @@ export default class AudioPlayerComponent extends Component {
 }
 ```
 
-Now the component is managing the state of the player, but we still need to translate that into the DOM method calls. So let's create a modifier!
+That's it for the component: we're translating the user's interactions into *state*. Now we need to build a modifier to translate the state into the appropriate DOM method calls!
```
ijlee2, line 388: I like these sentences. They reminded me of composing functions in math.

chriskrycho: Don’t tell anyone, but I’m kind of obsessed with composition in the mathematical sense as a key ingredient of software design. 🤓

chriskrycho:
@ijlee2 great catch—that *was* a weird sentence. I think we’re ready to go now!

guides/release/components/template-lifecycle-dom-and-modifiers.md
chriskrycho: Thanks for catching it as a point of confusion!

ijlee2:
@chriskrycho My bad, I need you to make 1 more change. Can you replace all instances of `*` with `_` for italicizing text? The linter for Markdown is failing.

chriskrycho:
Doh! Yep, will get that squared away in a bit!

chriskrycho:
Okay, I *think* I got them all. 🤞

chriskrycho:
Whoops! Caught a visual regression (probably something we should fix in the underlying styles):

visual-bug-lists

Pushing a change now.

chriskrycho:
Let me know next steps for back-porting; I'm happy to open that PR as well!

ijlee2:
@chriskrycho Yes, please. Can you check that you have the latest code in your default branch? Afterwards, will you backport the update to:

- `/guides/v3.15.0`
- `/guides/v3.16.0`
- `/guides/v3.17.0`
- `/guides/v3.18.0`
- `/guides/v3.19.0`

Thank you!

chriskrycho:
Will have that up by the end of the day!

PR #1530, opened 2020-09-02
Replace (some) of First & Last Name variable example references

hergaiety:
## Issue
https://github.com/ember-learn/guides-source/issues/1480

TL;DR Replacing `firstName` and `lastName` example code with `givenName` or `familyName`, or `title` and `name`, or just simply `name`, or something else entirely for i18n/W3C/respect concerns.

## Pages touched in this batch of changes

* [X] Configuration - Optional Features - default-async-observers
* [X] Ember Data - Creating, Updating and Deleting - Updating Records
* [X] Ember Data - Customizing Serializers - Attribute Names
* [X] Testing - Testing Basics - Stubbing an Object
* [X] Accessibility - Component Considerations - Accessible name - Adding a label to an input element
* [X] Components - Helper Functions - Built-in Helpers - The let helper
* [X] Components - Built-in Components - <Input/> - Setting attributes on Input
* [X] Components - Built-in Components - <Textarea />
* [X] Models - Introduction - Models

## Notes

I don't feel this needs to go back prior to `/release` but I'm open to a suggested amount of backporting.

Open to suggested alternatives for these changes as well.

PR #1535, opened 2020-09-18
Removed incorrect assertion

hergaiety:
As pointed out by @ijlee2 in #1530 this assertion is invalid because...

> ...`someThing.employees[0]` and `person` are both class instances...

```js
assert.equal(someThing.employees[0], person);
```


ijlee2 (approved):
✨

Issue #1543, opened 2020-09-27
Show how to reference input event names correctly


jenweber:
In earlier versions of Ember, inputs had a bug where certain event names could not be referenced as they were documented in the Guides, such as `<Textarea @focus-in={{this.didFocusIn}} />`. This has been fixed now! We should update the Guides to make sure they give the correct guidance. This issue is recommended for writers with some past technical writing experience or past contributions to Ember's documentation.

To-do:
- Try out the event passing like `@focus-in` in the latest stable version of Ember and confirm that it works
- Read over https://guides.emberjs.com/release/components/built-in-components/ and see which examples/explanations should be updated. It is possible that no changes are needed, since the source of the confusion was that the Guides said something should work that didn't, and now there's a bugfix. If that's the case, please comment on this issue.
- Feel free to open Issues for anything else you notice along the way that needs fixing
- Reach out to the learning team on #dev-ember-learning if you have questions.

For reference:
- Read past discussion here https://github.com/emberjs/ember.js/issues/18654
- Bugfix 1 https://github.com/emberjs/ember.js/pull/18997
- Bugfix 2 https://github.com/emberjs/ember.js/pull/19001

Note that there are still some complexities that are _not bugs_ which remain, such as understanding how action, fn, and modifiers fit into this story. Those are out of scope for this issue, since our first step should always be to make sure that the Guides are correct. If you are on a roll and want to tackle that too, go for it! For ease of review, please do separate PRs if you choose to keep going.

jenweber:
@ijlee2  is working on this issue.

PR #1546, opened 2020-09-29
Updated Built-In Components page to clarify how to use <Input> component

ijlee2:
## Description

This PR addresses and can close issue #1543 . In particular, in the Ember Learning Team's `Input component plans` document, this PR completes 3 action items:

- [x] Take out the parts that talk about what doesn't work (for example, "Checkboxes are a special input type..." section
- [x] Present modifiers as the main way to do events
- [x] Show example of `{{on "input}}` or `{{on "change"}}` handling ~w/o addons in the first pass. Maybe later show keydown~ (For simplicity in code, I showed how to migrate `@enter` and `@escape-press` using `ember-keyboard`)

I read everything in [Built-In Components](https://guides.emberjs.com/release/components/built-in-components/) to figure out how to update each subsection and its examples, so that the updated page will be coherent as a whole.

I also created a new Ember v3.21.1 app. I copy-pasted the existing code samples to figure out what worked and what didn't. The updated code samples should work when a developer copy-pastes them to a new app (assuming v3.21).

## How to Review

I recommend reviewing commits one by one. Each commit looks at one subsection in `Built-In Components`.

## Notes

In meeting, @chancancode, @jenweber, @locks, and I weren't sure whether, with `<Input>` component, a developer needed to use `@id` (argument) or `id` (attribute) to correctly bind to the input element's `id` attribute.

Based on my `v3.21.1` app, I think we have to use `id`. (This is good for us, as it requires fewer changes.)

Consider the following code. It changes the input ID every 1 second.

```handlebars
{{!-- app/templates/application.hbs --}}
<label for={{this.myUniqueId}}>
  Name:
</label>

<Input
  id={{this.myUniqueId}}
  @value={{this.name}}
/>

<div {{did-insert this.changeUniqueId}}></div>
```

```javascript
// app/controllers/application.js

import Controller from '@ember/controller';
import { later } from '@ember/runloop';

  @tracked myUniqueId = 'happy-little-trees';
  @tracked name = 'Bob Ross';

  @action changeUniqueId() {
    later(() => {
      this.myUniqueId = Math.random().toString();
      this.changeUniqueId();
    }, 1000);
  }
}
```

When we use `<Input id=... />`, in Chrome Inspector, we will see that the ID changes for the `for` and `id` attributes every second. Clicking on the label will focus on the input as well.

However, when we use `<Input @id=... />`, we will see that the ID changes only for the `for` attribute in the label. Clicking on the label will not focus on the input anymore.

### DIFF
--- a/guides/release/components/built-in-components.md
+++ b/guides/release/components/built-in-components.md
@@ -14,7 +14,11 @@ Consider the following example in a template file.
 
 ```handlebars
 <label for="user-question">Ask a question about Ember:</label>
-<Input id="user-question" @type="text" @value="How do text fields work?" />
+<Input
+  @id="user-question"
+  @type="text"
+  @value="How do text fields work?"
+/>
 ```
 
 When Ember renders this template, you will see the following HTML code:
@@ -42,7 +46,7 @@ Every input should be associated with a label. In HTML, there are a few ways to 
    </label>
    ```
 
-2. You can create an ID (globally unique within the webpage), then associate the label to the input with `for` and `id` attributes.
+2. You can create an ID (globally unique within the webpage), then associate the label to the input with `for` attribute and `@id` argument.
 
    ```handlebars
    <label for={{this.myUniqueId}}>
@@ -50,7 +54,7 @@ Every input should be associated with a label. In HTML, there are a few ways to 
    </label>
 
    <Input
-     id={{this.myUniqueId}}
+     @id={{this.myUniqueId}}
      @type="text"
      @value={{this.userQuestion}}
    />
@@ -88,7 +92,7 @@ In the next example, the `disabled` attribute is bound to the value of `isReadOn
 ```handlebars
 <label for="input-name">Name:</label>
 <Input
-  id="input-name"
+  @id="input-name"
   @value={{this.name}}
   disabled={{this.isReadOnly}}
   maxlength="50"
@@ -98,6 +102,7 @@ In the next example, the `disabled` attribute is bound to the value of `isReadOn
 Recall that there were a few exceptions. The following input attributes must be passed as arguments (i.e. do prepend `@`) to the `<Input>` component:
 
 - `@checked`
+- `@id`
 - `@type`
 - `@value`
 
@@ -109,7 +114,7 @@ Starting with Ember Octane, we recommend using the `{{on}}` modifier to call an 
 ```handlebars
 <label for="input-name">Name:</label>
 <Input
-  id="input-name"
+  @id="input-name"
   @value={{this.name}}
   {{on "input" this.validateName}}
 />
@@ -122,7 +127,7 @@ Due to legacy code, it is possible to call an action by passing an event argumen
 ```handlebars
 <label for="input-name">Name:</label>
 <Input
-  id="input-name"
+  @id="input-name"
   @value={{this.name}}
   @input={{this.validateName}}
 />
@@ -160,7 +165,7 @@ component to create a checkbox. Please set `@type` to the string `"checkbox"`, a
 ```handlebars
 <label for="admin-checkbox">Is Admin?</label>
 <Input
-  id="admin-checkbox"
+  @id="admin-checkbox"
   @type="checkbox"
   @checked={{this.isAdmin}}
 />
@@ -171,7 +176,7 @@ To call an action on specific events, please use the `{{on}}` modifier:
 ```handlebars
 <label for="admin-checkbox">Is Admin?</label>
 <Input
-  id="admin-checkbox"
+  @id="admin-checkbox"
   @type="checkbox"
   @checked={{this.isAdmin}}
   {{on "input" this.validateRole}}
@@ -186,7 +191,7 @@ The following example shows how to bind `this.userComment` to a text area's valu
 ```handlebars
 <label for="user-comment">Comment:</label>
 <Textarea
-  id="user-comment"
+  @id="user-comment"
   @value={{this.userComment}}
   rows="6"
   cols="80"
@@ -214,7 +219,10 @@ in conjunction like shown in the following example:
 
 ```handlebars
 <label for="input-name">Name:</label>
-<Input @value={{mut (get this.person this.field)}} id="name" name="input-name" />
+<Input
+  @id="input-name"
+  @value={{mut (get this.person this.field)}}
+/>
 ```
 
 The `{{get}}` helper allows you to dynamically specify which property to bind,
--- a/guides/release/components/built-in-components.md
+++ b/guides/release/components/built-in-components.md
@@ -153,46 +153,31 @@ Note, the `keydown` event was used for `Escape` because `keypress` is deprecated
 
 ### Checkboxes
 
-You can also use the
-[`<Input />`](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input?anchor=Input)
-component to create a checkbox by setting its `@type`:
+You can use the
+[`<Input>`](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input?anchor=Input)
+component to create a checkbox. Please set `@type` to the string `"checkbox"`, and use `@checked` instead of `@value`.
 
 ```handlebars
 <label for="admin-checkbox">Is Admin?</label>
-<Input id="admin-checkbox" @type="checkbox" name="isAdmin" @checked={{this.isAdmin}} />
+<Input
+  id="admin-checkbox"
+  @type="checkbox"
+  @checked={{this.isAdmin}}
+/>
 ```
 
-Checkboxes support the following properties:
-
-* `@checked`
-* `disabled`
-* `tabindex`
-* `indeterminate`
-* `name`
-* `autofocus`
-* `form`
-
-
-Which can be bound or set as described in the previous section.
-
-
-Checkboxes are a special input type.  If you want to dispatch an action on a certain [event](https://api.emberjs.com/ember/release/classes/Component#event-handler-methods),
-you will always need to either define the event name in camelCase format (e.g. `@keyDown`), or
-use an `on` helper with the [Web-API event name](https://developer.mozilla.org/en-US/docs/Web/API/Document/keydown_event) (e.g. `on 'keydown'`):
+To call an action on specific events, please use the `{{on}}` modifier:
 
 ```handlebars
-<label for="input-name">Name:</label>
-{{!-- This works: uses camelCase event name --}}
-<Input @type="checkbox" @keyDown={{this.updateName}} id="input-name" />
-{{!-- This works: uses 'on' with actual event name --}}
-<Input @type="checkbox" {{on "keydown" this.updateName}} id="input-name" />
-{{!-- This does not work: uses dasherized event name --}}
-<Input @type="checkbox" @key-down={{this.updateName}} id="input-name" />
-{{!-- This does not work: uses actual event name --}}
-<Input @type="checkbox" @keydown={{this.updateName}} id="input-name" />
+<label for="admin-checkbox">Is Admin?</label>
+<Input
+  id="admin-checkbox"
+  @type="checkbox"
+  @checked={{this.isAdmin}}
+  {{on "input" this.validateRole}}
+/>
 ```
 
-Internally, `<Input @type="checkbox" />` creates an instance of Checkbox. Do *not* use `Checkbox` directly.
 
 ## `<Textarea />`
 
--- a/guides/release/components/built-in-components.md
+++ b/guides/release/components/built-in-components.md
@@ -160,7 +160,7 @@ Note, the `keydown` event was used for `Escape` because `keypress` is deprecated
 
 You can use the
 [`<Input>`](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input?anchor=Input)
-component to create a checkbox. Please set `@type` to the string `"checkbox"`, and use `@checked` instead of `@value`.
+component to create a checkbox. Set `@type` to the string `"checkbox"`, and use `@checked` instead of `@value`.
 
 ```handlebars
 <label for="admin-checkbox">Is Admin?</label>
--- a/guides/release/components/built-in-components.md
+++ b/guides/release/components/built-in-components.md
@@ -104,26 +104,52 @@ Recall that there were a few exceptions. The following input attributes must be 
 
 ### Actions
 
-To dispatch an action on specific events such as `key-down`, use the following
+Starting with Ember Octane, we recommend using the `{{on}}` modifier to call an action on specific events such as the [input event](https://developer.mozilla.org/docs/Web/API/HTMLElement/input_event).
 
 ```handlebars
 <label for="input-name">Name:</label>
-<Input id="input-name" name="name" @type="text" @value={{this.name}} @key-down={{this.updateName}} />
+<Input
+  id="input-name"
+  @value={{this.name}}
+  {{on "input" this.validateName}}
+/>
 ```
 
-The following event types are supported (dasherized format):
+The event name (e.g. `"focusout"`, `"input"`, `"keydown"`) always follows the casing that the HTML standard uses.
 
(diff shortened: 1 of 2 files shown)


chancancode:
I think we probably need to do more testing. It is in fact illegal to change the ID of an ember component after its instantiated (and probably a bit sketchy though valid on HTML elements in general), which is why it doesn’t update using your method (I think you are supposed to get an assertion).

I think this isn’t that common in practice and instead we should focus on whether things like event delegation still works if you use the ID attribute. ie does the focusIn(), etc hooks defined on the component still fire correctly.

ijlee2:
@chancancode Sounds good, thanks for the quick feedback. I'll look at event delegation more closely and let you know how it goes.

ijlee2:
@chancancode Here is my updated example. I hope I understood what you were looking for:

<details>
<summary>Template</summary>

```handlebars
{{!-- app/templates/application.hbs --}}

<label for="version-1">Name:</label>
<Input
  id="version-1"
  @value={{this.name1}}

  @mouse-down={{this.mouseDown}}
  @mouse-up={{this.mouseUp}}
  @context-menu={{this.contextMenu}}
  @click={{this.click}}
  @double-click={{this.doubleClick}}

  @change={{this.change}}
  @focus-in={{this.focusIn}}
  @focus-out={{this.focusOut}}
  @input={{this.input}}
/>


<label for="version-2">Name:</label>
<Input
  id="version-2"
  @value={{this.name2}}

  @mouseDown={{this.mouseDown}}
  @mouseUp={{this.mouseUp}}
  @contextMenu={{this.contextMenu}}
  @click={{this.click}}
  @doubleClick={{this.doubleClick}}

  @change={{this.change}}
  @focusIn={{this.focusIn}}
  @focusOut={{this.focusOut}}
  @input={{this.input}}
/>


<label for="version-3">Name:</label>
<Input
  id="version-3"
  @value={{this.name3}}

  {{on "mousedown" this.mouseDown}}
  {{on "mouseup" this.mouseUp}}
  {{on "contextmenu" this.contextMenu}}
  {{on "click" this.click}}
  {{on "dblclick" this.doubleClick}}

  {{on "change" this.change}}
  {{on "focusin" this.focusIn}}
  {{on "focusout" this.focusOut}}
  {{on "input" this.input}}
/>
```

</details>

<details>
<summary>Controller</summary>

```javascript
// app/controllers/application.js

import Controller from '@ember/controller';

  @tracked name1 = 'Bob Ross';
  @tracked name2 = 'Bob Ross';
  @tracked name3 = 'Bob Ross';


  // https://api.emberjs.com/ember/release/classes/Component#event-handler-methods
  @action mouseDown(event) {
    console.log('mousedown');
    console.log(event);
  }

  @action mouseUp(event) {
    console.log('mouseup');
    console.log(event);
  }

  @action contextMenu(event) {
    console.log('contextmenu');
    console.log(event);
  }

  @action click(event) {
    console.log('click');
    console.log(event);
  }

  @action doubleClick(event) {
    console.log('dblclick');
    console.log(event);
  }


  @action change(event) {
    console.log('change');
    console.log(event.target.value);
  }

  @action focusIn(event) {
    console.log('focusin');
    console.log(event);
  }

  @action focusOut(event) {
    console.log('focusout');
    console.log(event);
  }

  @action input(event) {
    console.log('input');
    console.log(event.target.value);
  }
}
```

</details>

I created 3 sets of `label-input`:

1. Uses event arguments with dasherized names (we want these to work)
2. Uses event arguments with camel-cased names (we want these to not work)
3. Uses `{{on}}` modifiers (we want these to work)

In `v3.21.1`, the 1st set doesn't work while the 2nd set does work. The 3rd set does work, of course. The results for all 3 sets seem to be the same whether I use `id` or `@id`.

Perhaps we need to test in `v3.16 LTS` instead (bugfix https://github.com/emberjs/ember.js/pull/19001)? From the bugfix https://github.com/emberjs/ember.js/pull/18997, I wasn't sure in which recent Ember version the fix landed.

chancancode:
Yeah that seems like a good test to me. The fix should be available in v3.21.1 so I would expect it to work. Are you able to:

1. Set a breakpoint around here https://github.com/emberjs/ember.js/blob/bd15e2d1d76278712dc3051b766149e31eefe26a/packages/%40ember/-internals/views/lib/mixins/text_support.js#L302 and see if it's called for the first input, and if so, then
2. Follow it into here https://github.com/emberjs/ember.js/blob/bd15e2d1d76278712dc3051b766149e31eefe26a/packages/%40ember/-internals/views/lib/mixins/text_support.js#L310 and follow it and see why it didn't end up calling the callback?

Otherwise you can push your app somewhere and I can take a look too

ijlee2:
I uploaded my demo app and recorded what I tried in `README`: https://github.com/ijlee2/investigate-input-events-in-3.21

chancancode:
@ijlee2 you can do it from the chrome devtools.

1. Go to the "Sources" pane in the Chrome dev tools.
2. Press "Command-P" to open the "Open File" dialog (you can find it in the UI buried in some menu too).
3. If you have source maps enabled, you can probably type in text_support and find the source file I linked. Otherwise, you'll probably have to open vendor.js and search the source text for the AMD module ("@ember/-internals/views/lib/mixins/text_support") using Command-F (full text search).
4. Either way, once you found the line, you can click on the line number to set a breakpoint.

chancancode:
(Instead of `keyDown`, you probably want to pick one that you actually bound in the template in your test. The focus events are okay, but they tend to be quite finicky because clicking on the devtools will cause the input to unfocus.)

chancancode:
I think I see the problem in your test: you tried to do `@mouse-down` etc but those aren't actually arguments supported by the component. `@foo-bar` isn't a general purpose way of registering callbacks. For example, `@key-down` works, but only because the component [is explicitly documented to support it](https://github.com/emberjs/ember.js/blob/bd15e2d1d76278712dc3051b766149e31eefe26a/packages/%40ember/-internals/views/lib/mixins/text_support.js#L106) and [has code that dispatches the callback](https://github.com/emberjs/ember.js/blob/bd15e2d1d76278712dc3051b766149e31eefe26a/packages/%40ember/-internals/views/lib/mixins/text_support.js#L302). Otherwise, it won't do anything.

Then, things like `@click`, `@change`, etc basically "works" by "overriding" the [methods on the component with the same name](https://github.com/emberjs/ember.js/blob/bd15e2d1d76278712dc3051b766149e31eefe26a/packages/%40ember/-internals/views/lib/mixins/text_support.js#L178-L180) which as we discussed is problematic.

ijlee2:
@chancancode Thanks very much for the instructions. I was able to set a breakpoint on `sendAction` and trigger the breakpoint for `keydown` event when I typed a character into the input field for Versions 1 and 3, but not Version 2.

[Changes I made to the sample app](https://github.com/ijlee2/investigate-input-events-in-3.21/commit/b6c938df9c7ebe8630b397f4a521ded03af2090c)

I think you're right that we have to use `@id` to bind to the `<Input>` component's `elementId`. If we use `id` instead, the component's `elementId` is the default, unique string `ember###`.

It's also worth noting that version 3 (`{{on}}` modifier approach) does not create `key-down`, `key-press`, and `key-up` properties in the `attrs` object. I imagine this supports the idea that `{{on}}` modifier is a better approach.

## Screenshots

### Version 1, `@id`

version 1 with @id

### Version 3, `@id`

version 3 with @id

### Version 1, `id`

version 1 with id

### Version 3, `id`

version 3 with id

chancancode:
> I was able to set a breakpoint on sendAction and trigger the breakpoint for keydown event when I typed a character into the input field for Versions 1 and 3, but not Version 2.

👍 I think that all went as expected. Version 2 doesn’t trip the breakpoint because the passed argument overwrote and replaced the handler on the component, which is The Problem™️.

> I think you're right that we have to use @id to bind to the <Input> component's elementId. If we use id instead, the component's elementId is the default, unique string ember###.

Yeah so that’s the problem/quirk of classic components in general. However, the fact that you can get version 1 working as expected either way means that Ember internally no longer uses `elementId` for the purpose of event dispatching, which was what I was originally worried that this would break.

If this is your own component or something from an addon that you are invoking, you may care about `elementId` being correct for other reasons, but it _seems_ like the built-in Input component doesn’t actually use it for anything or otherwise care about its value? In that case, even though it is more Correct™️ to use `@id` we may be able to get away without it in this case and still not cause any real issues.

> It's also worth noting that version 3 ({{on}} modifier approach) does not create key-down, key-press, and key-up properties in the attrs object. I imagine this supports the idea that {{on}} modifier is a better approach.

Yep that is also as expected. The `{{on}}` modifier doesn’t interact with the component at all and instead operate on the underlying DOM element (the `<input>` in this case) via `...attributes` (which is implicit on the wrapper element for classic components). So the component doesn’t “see” these event handlers at all.

ijlee2:
Awesome. Other than a review of the writing, this PR should be good to go then?

guides/release/components/built-in-components.md
```diff
@@ -1,190 +1,215 @@
...
 
-* `@enter`
-* `@insert-newline`
-* `@escape-press`
-* `@focus-in`
-* `@focus-out`
-* `@key-down`
-* `@key-press` ([Deprecated Web API](https://developer.mozilla.org/en-US/docs/Web/API/Document/keypress_event))
-* `@key-up`
+Due to legacy code, it is possible to call an action by passing an event argument.
```
jenweber, line 125:
```suggestion
For backwards compatibility with earlier versions of Ember, it is possible to call an action by passing an event argument.
```

guides/release/components/built-in-components.md
```diff
@@ -1,190 +1,215 @@
...
+  @value={{this.name}}
+  @input={{this.validateName}}
+/>
+```
 
-More [event types](https://api.emberjs.com/ember/release/classes/Component#event-handler-methods) are also supported but these events need to be written in camelCase format, such `mouseEnter`. Note, there are events of the same type in both the list above and linked. Event names listed above must be dasherized. Additional work is performed on these events.
+The argument name is always dasherized (e.g. `@focus-out`, `@input`, `@key-down`). To minimize confusion, we recommend that you use the `{{on}}` modifier. ([Learn more about the `{{on}}` modifier.](../../upgrading/current-edition/action-on-and-fn/#toc_the-on-modifier))
 
-### Checkboxes
+Lastly, Ember Classic provided custom input events `@enter` and `@escape-press`. These are not a part of the HTML standard, but may facilitate accessibility.
```
jenweber, line 138:
```suggestion
Lastly, Ember also provides custom input events `@enter` and `@escape-press`. These events do not exist on native input elements, but you may find them to be useful for handling keyboard interactions.
```

guides/release/components/built-in-components.md
```diff
@@ -1,190 +1,215 @@
...
 
-You can also use the
-[`<Input />`](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input?anchor=Input)
-component to create a checkbox by setting its `@type`:
+Starting with Ember Octane, please consider [writing a modifier](../../upgrading/current-edition/glimmer-components/#toc_writing-your-own-modifiers) to separate concerns: The component manages the state, while the modifier manages interactions with the DOM. If you prefer installing an addon, we recommend the `{{on-key}}` modifier from [ember-keyboard](https://github.com/adopted-ember-addons/ember-keyboard). Your action will receive an actual `event` object.
```
jenweber, line 140:
```suggestion
The modern, Octane-style way to handle keyboard events is to [write a modifier](../../upgrading/current-edition/glimmer-components/#toc_writing-your-own-modifiers) to separate concerns: The component manages the state, while the modifier manages interactions with the DOM. Your action will receive an actual `event` object. There are also [community-made addons](https://emberobserver.com/?query=keyboard) to help manage keyboard events.

```

jenweber: Saying that an addon is recommended is sticky business. It's preferred to link to Ember Observer unless there is only one option to talk about (i.e. there is no equivalent to ember-concurrency)

guides/release/components/built-in-components.md
```diff
@@ -1,190 +1,215 @@
...
+
+{{!-- After --}}
+<Input
+  {{on-key "Enter" this.doSomething}}
+  {{on-key "Escape" this.doSomethingElse event="keydown"}}
+/>
 ```
 
-Checkboxes support the following properties:
+Note, the `keydown` event was used for `Escape` because `keypress` is deprecated.
```
jenweber, line 156: Optionally - I think we could probably cut this now. It's been deprecated for a while.

guides/release/components/built-in-components.md
```diff
@@ -1,190 +1,215 @@
...
-* `name`
-* `autofocus`
-* `form`
 
+### Checkboxes
 
-Which can be bound or set as described in the previous section.
+You can use the
+[`<Input>`](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input?anchor=Input)
+component to create a checkbox. Please set `@type` to the string `"checkbox"`, and use `@checked` instead of `@value`.
```
jenweber, line 163:
```suggestion
component to create a checkbox. Set `@type` to the string `"checkbox"`, and use `@checked` instead of `@value`.
```

Just edits to match the tone of the rest of the guides.

guides/release/components/built-in-components.md
```diff
@@ -1,190 +1,215 @@
...
+  @id="admin-checkbox"
+  @type="checkbox"
+  @checked={{this.isAdmin}}
+/>
+```
 
-Checkboxes are a special input type.  If you want to dispatch an action on a certain [event](https://api.emberjs.com/ember/release/classes/Component#event-handler-methods),
-you will always need to either define the event name in camelCase format (e.g. `@keyDown`), or
-use an `on` helper with the [Web-API event name](https://developer.mozilla.org/en-US/docs/Web/API/Document/keydown_event) (e.g. `on 'keydown'`):
+To call an action on specific events, please use the `{{on}}` modifier:
```
jenweber, line 174:
```suggestion
To call an action on specific events, use the `{{on}}` modifier:
```

jenweber (changes requested):
Hello! Thanks so much for writing this! It is a big improvement for our users.

I added a few comments and wording changes. The ones that I would say are required are, we should not say that any particular addon is recommended, if linking to a list of them will suffice. The others are around cutting down references to Ember Classic and "legacy." We have some of that language sprinkled through the Guides to help give people a glide path, but Octane has been out for a while now and I think we should walk towards removing the references. These may not be described in a styleguide anywhere - I will try to fix that.

guides/release/components/built-in-components.md
ijlee2: To my understanding, `ember-keyboard` is the only addon that handles complexities (e.g. `key` vs `code`) in listening to keyboard events and provides a modifier as solution. I don't think it'd be a good idea to ask a developer to create their own modifier and risk not addressing the complexities, then breaking the web.

Since the repo is maintained by `adopted-ember-addons` organization, I was thinking it should be okay for us to recommend this addon? The code example of migrating `@enter` and `@escape-press` can be simple because it assumed that the addon has been installed.

guides/release/components/built-in-components.md
ijlee2: Just in case a developer wasn't aware, I think we can keep it.

guides/release/components/built-in-components.md
ijlee2: I think we could also consider the option of linking to both Ember Observer (in case there will be more keyboard-related addons in the future) and `ember-keyboard`.

```suggestion
The modern, Octane-style way to handle keyboard events is to [write a modifier](../../upgrading/current-edition/glimmer-components/#toc_writing-your-own-modifiers) to separate concerns: The component manages the state, while the modifier manages interactions with the DOM. Your action will receive an actual `event` object.

There are also [community-made addons](https://emberobserver.com/?query=keyboard) to help manage keyboard events. For example, with [ember-keyboard](https://github.com/adopted-ember-addons/ember-keyboard), you can write,
```

jenweber (approved):
Looks great, thank you!

PR #1588, opened 2020-10-30
Update built-in-components.md

aguynamedryan:
Maybe it's just my brain, but I had to reread "do prepend" several times over as I kept thinking it said "do not prepend".  "do prepend" seems a bit awkward, so I changed it to a more idiomatic phrasing


ijlee2:
Let me decline the suggested change (I appreciate your feedback). I think the two phrases may boil down to a personal preference.

PR #1526, opened 2020-08-23
Updated API links in guides/3.7.0

ijlee2:
## Description

This PR addresses #1440 for Ember `v3.7.0`. After performing Find All and Replace, I manually visited each link to ensure that we redirect readers to the correct content in the API Guides.

Similarly to #1512, this PR involves many file changes because it backports fixing incorrect links to `v3.0.0`-`v3.6.0`. (These are the versions grouped under [Phase 3](https://github.com/ember-learn/guides-source/issues/1440#issuecomment-666800602).)

I hope the number of files doesn't deter you from reviewing this PR. Subsequent PRs for Phase 3 (`v3.0.0`-`v3.6.0`) will have fewer file changes.

## How to review

I recommend reviewing commits one by one.

Every commit, starting with commit 2 ("Fixed incorrect links in..."), looks at just 1 Markdown file across different versions. The pattern in URL change often tends to be the same. (The only exception is if the API Guides changed URL from one version to another.)

## TODO

- [x] Use Find All and Replace to update API links.
- [x] Manually check each URL to ensure that we redirect a user to the correct API section.
- [x] Run `npm run test:node-local` to check external links in `v3.7.0`.

## Notes

```
# Use regexes for Find All and Replace
Find:    https://api.emberjs.com/ember/release/
Replace: https://api.emberjs.com/ember/3.7/

Find:    https://api.emberjs.com/ember/\d{1}.\d{1,2}/
Replace: https://api.emberjs.com/ember/3.7/

Find:    https://api.emberjs.com/ember-data/release/
Replace: https://api.emberjs.com/ember-data/3.7/

Find:    https://api.emberjs.com/ember-data/\d{1}.\d{1,2}/
Replace: https://api.emberjs.com/ember-data/3.7/
```

PR #1512, opened 2020-08-13
Updated API links in guides/3.14.0

ijlee2:
## Description

This PR addresses https://github.com/ember-learn/guides-source/issues/1440 for Ember `v3.14.0`. After performing Find All and Replace, I manually visited each link to ensure that we redirect readers to the correct content in the API Guides.

Similarly to https://github.com/ember-learn/guides-source/pull/1494, this PR involves many file changes because it backports fixing incorrect links to `v3.8.0`-`v3.13.0`. (These are the versions grouped under [Phase 2](https://github.com/ember-learn/guides-source/issues/1440#issuecomment-666800602).)

I hope the number of files doesn't deter you from reviewing this PR. Subsequent PRs for Phase 2 (`v3.8.0`-`v3.13.0`) will have fewer file changes.

## How to Review

I recommend reviewing the commits one at a time.

Every commit, starting with commit 4 ("Fixed incorrect links in..."), looks at just 1 Markdown file across different versions. The pattern in URL change often tends to be the same. (The only exception is if the API Guides changed URL from one version to another.)

## TODO

- [x] Use Find All and Replace to update API links.
- [x] Manually check each URL to ensure that we redirect a user to the correct API section.
- [x] Run `npm run test:node-local` to check external links in `v3.14.0`.

## Notes

In `/guides/v3.14.0`, I realized that API links don't always use `release`, but can use some version prior. I will need to open a separate PR to address this issue for `v3.15.0` to `v3.19.0`. 

```
# Use regexes for Find All and Replace
Find:    https://api.emberjs.com/ember/release/
Replace: https://api.emberjs.com/ember/3.14/

Find:    https://api.emberjs.com/ember/\d{1}.\d{1,2}/
Replace: https://api.emberjs.com/ember/3.14/

Find:    https://api.emberjs.com/ember-data/release/
Replace: https://api.emberjs.com/ember-data/3.14/

Find:    https://api.emberjs.com/ember-data/\d{1}.\d{1,2}/
Replace: https://api.emberjs.com/ember-data/3.14/
```

PR #1592, opened 2020-11-01
New script to update version links using the correct version number instead of release

neojp:
Currently using `npm run release:guides:minor` to create a new release copies `guides/release` to a its version number (eg. `guides/v3.22.0`), but there is no automated way to update the version numbers in all links pointing to its guides or API links.

See issue #1440 and its https://github.com/ember-learn/guides-source/issues/1440#issuecomment-666800602 to see how much of a chore this convoluted process is.

This PR attempts to:

* Provide a CLI script that can be used on old releases and update its version numbers
image

* Hook into the current `release:guides:minor` script and automate the process for future releases
Code_yRoYCR7jrp

---

The new npm script `release:guides:links` syntax looks like this:

```
npm run release:guides:links <guides-path> <release-number> -- <options...>
```

A good example for this use case is version `v3.21.0` which hasn't gone through the process to be updated.

```
npm run release:guides:links guides/v3.21.0 3.21
```

Would return:

```
Welcome to the automated process of updating the Guides version links

Found: 168 files
Updated: 34 files
```

And would update all `.md` files that contain links like:

```
https://api.emberjs.com/ember/release/ -> https://api.emberjs.com/ember/3.21/
https://api.emberjs.com/ember/3.15/ -> https://api.emberjs.com/ember/3.21/
https://api.emberjs.com/ember-data/release/ -> https://api.emberjs.com/ember-data/3.21/
https://api.emberjs.com/ember-data/3.15/ -> https://api.emberjs.com/ember-data/3.21/
https://guides.emberjs.com/release/getting-started/ -> https://guides.emberjs.com/v3.21/getting-started/
```

---

There are a set of options built into the script:

* `--verbose` to display the full list of updated files
* `--silent` to prevent messages to be sent to the console; used in `release:guides:minor`.
* `--dry-run` to see how many files are to be updated. Use it in combination with `--verbose` to see the full list. 
* `--tests` to see a simple list of unit tests used to match the URLs

---

Hopefully this will make things easier moving forward, and will be significantly easier to update older releases as well.

**Note:** this script can work independently from npm and doesn't have to be an npm-script to run. We can remove it from the npm-scripts if preferred.

---

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "test:node-local": "mocha node-tests/local",
     "test:node-local-exclude-api-urls": "FOLLOW_API_URLS=false npm run test:node-local",
     "release:guides:minor": "scripts/create-new-minor-version",
+    "release:guides:links": "node scripts/update-version-links",
     "release:search": "scripts/update-search-index"
   },
   "devDependencies": {
--- a/scripts/create-new-minor-version
+++ b/scripts/create-new-minor-version
@@ -66,6 +66,11 @@ sed -i .bak -e "/currentVersion/i \\
 rm guides/versions.yml.bak
 echo "  DONE"
 
+# Update version numbers in links
+echo "🤖 Updating version number for links in /guides/$CURRENT_VERSION/**/*.md"
+node scripts/update-version-links guides/$CURRENT_VERSION $(echo $CURRENT_VERSION | cut -d "v" -f2) --silent
+echo "  DONE"
+
 echo
 echo "🤖 Committing changes and publishing branch to remote"
 git add .
--- a/scripts/update-version-links
+++ b/scripts/update-version-links
@@ -1,0 +1,140 @@
+#!/usr/bin/node
+
+/**
+ * Search & Replace previous release numbers
+ * such as `3.15`, `3.15.0`, `v3.15.0` or `release`
+ * to the provided release number for guides and api URLs
+ *
+ * Running tests:
+ * ./update-version-links --tests
+ *
+ * How to use:
+ * ./update-version-links <guides-path> <release-number> <options...>
+ *
+ * Examples:
+ * ./update-version-links ../guides/v3.21.0 3.21
+ * ./update-version-links ../guides/v3.21.0 3.21 --dry-run
+ *
+ * Will replace the following:
+ * https://api.emberjs.com/ember/release/ -> https://api.emberjs.com/ember/3.21/
+ * https://api.emberjs.com/ember/3.15/ -> https://api.emberjs.com/ember/3.21/
+ * https://api.emberjs.com/ember-data/release/ -> https://api.emberjs.com/ember-data/3.21/
+ * https://api.emberjs.com/ember-data/3.15/ -> https://api.emberjs.com/ember-data/3.21/
+ * https://guides.emberjs.com/release/getting-started/ -> https://guides.emberjs.com/v3.21/getting-started/
+ */
+
+const fs = require('fs');
+
+const replaceURLsVersions = (str, version) => {
+  let output;
+
+  // guides use `v3.20.0` version format
+  const patchVersion = /\d+\.\d+.\d+/.test(version) ? version : version + '.0';
+  output = str.replace(
+    /https:\/\/guides.emberjs.com\/(release|v\d+\.\d+.\d+)(\/?)/g,
+    `https://guides.emberjs.com/v${patchVersion}$2`
+  );
+
+  // apis use `3.20` version format
+  output = output.replace(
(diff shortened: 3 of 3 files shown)


jenweber:
Hi, thanks for your work on this! I tested it locally, and the commands such as `npm run release:guides:links guides/v3.21.0 3.21` seems to work as described. This will save a lot of manual effort. However, if I run the script as documented inside the script itself, I get the following issue:

```
cd scripts
bash ./update-version-links ../guides/v3.21.0 3.21

./update-version-links: line 3: /Applications: is a directory
./update-version-links: line 4: create-new-minor-version: command not found
./update-version-links: line 4: Replace: command not found
./update-version-links: line 5: 3.15: command not found
./update-version-links: line 5: 3.15.0: command not found
./update-version-links: line 5: v3.15.0: command not found
./update-version-links: line 5: release: command not found
./update-version-links: line 5: create-new-minor-version: command not found
./update-version-links: line 6: create-new-minor-version: command not found
./update-version-links: line 7: create-new-minor-version: command not found
./update-version-links: line 8: create-new-minor-version: command not found
./update-version-links: line 9: create-new-minor-version: command not found
./update-version-links: line 10: create-new-minor-version: command not found
./update-version-links: line 11: create-new-minor-version: command not found
./update-version-links: line 12: syntax error near unexpected token `<'
./update-version-links: line 12: ` * ./update-version-links <guides-path> <release-number> <options...>'
```

Is there something I should do differently?

neojp:
This is technically a nodejs script. My guess is that using bash to run it
isn't correctly using /usr/bin/node.

Although your errors seem to be very specific and somehow mention the other
script create-new-minor-version.

Could you try this instead?

```
node ./update-version-links ../guides/v3.21.0 3.21
```

>

ijlee2:
Hi, @neojp. Thanks for coming up with a script that may automate fixing the API links. Due to lack of time, I haven't reviewed the code carefully. Can you give us a brief overview of what running the script will do on various versions?

I'm concerned that running the script on versions 3.5 and below will break some of the redirects. When I worked on updating links for 3.6 and above, there wasn't a uniform pattern to find-and-replace. In addition, some links pointed to an incorrect page in the API Guides.

As a result, I had to manually check all links in the most recent version in a phase, then backport the fix to previous versions in that phase. I was wondering how your script may address these problems.

ijlee2:
For 3.21, we could do a find-and-replace all in VS Code. Since the links in recent versions have been audited already, the find-and-replace all technique would work.

neojp:
@jenweber I can see your result in my machine as well. I thought that by having the shebang comment at the top it would execute itself with node even by using `bash` or `sh`, but I guess I was mistaken.

The correct way to run the script is:

```
cd scripts
./update-version-links ../guides/v3.21.0 3.21
```

or

```
cd scripts
node update-version-links ../guides/v3.21.0 3.21
```

We could rename the script to `update-version-links.js` so it's obvious that it's meant to be run with node.

---

@ijlee2 I have been testing it in a few versions and it was working so far. I manually reviewed the links to see if they were working and if they existed. Like you said, it's possible that older versions of the guides might fail, so I will review those first.

However, if the URLs follow the same schema (eg. `v1.0.0` for guides and `1.0` for apis) then the script should work.

I was planning on running them and submitting PRs anyway, so maybe we can use those as the test cases to confirm that the script works correctly.

ijlee2:
@neojp Yep, I think we can create a pull request to test your script on v3.5. (Version 3.0 may be the one that will trip us.)

neojp:
@ijlee2 I think I see I finally understand what you mean, by running the script in `v3.15.0` I noticed there are 3 links pointing to older versions of the guides, `v2.15.0` to be precise.

I thought this was a human mistake, but after reviewing them manually I noticed they don't exist in `v3.15.0` anymore and that's why they are linked there.

I'm guessing we should keep them as is? I'm attaching a Gist with the 3 link changes.

https://gist.github.com/neojp/1f3ad791152bcea6645c40a5a84be39f

ijlee2:
@neojp Yes, I believe I left that version `2.15` alone. You can assume that all API links in the Ember Guides starting `v3.6` are correct, as I already audited them.

For more examples of exceptions, I'd check the most recent version in each "phase."

- https://github.com/ember-learn/guides-source/pull/1512/files
- https://github.com/ember-learn/guides-source/pull/1526/files

neojp:
@ijlee2 Got it. I will keep looking at these links manually and leave the exceptions alone.

ijlee2 (approved):
Let's merge this PR so that the script's temporarily available to help with updating API links for `v3.5.0` and below. Thanks again for looking into automating the process!

PR #1613, opened 2020-12-10
Update native-classes-in-depth.md

nruth:
I found alert confusing, it's not defined & console.log in the caller would expect a string return value?

### DIFF
--- a/guides/release/in-depth-topics/native-classes-in-depth.md
+++ b/guides/release/in-depth-topics/native-classes-in-depth.md
@@ -426,7 +426,7 @@ directly on the class itself.
 ```js
 class Alert {
   static helloWorld() {
-    alert('Hello, world!');
+    return 'Hello, world!';
   }
 }
 


locks (approved):
Thanks!

guides/release/in-depth-topics/native-classes-in-depth.md
```diff
@@ -426,7 +426,7 @@ directly on the class itself.
 ```js
 class Alert {
   static helloWorld() {
-    alert('Hello, world!');
```
ijlee2, line 429: I believe the `alert` refers to the built-in `Window.alert()` method. https://developer.mozilla.org/docs/Web/API/Window/alert

Based on lines 433 and 437, I think you are correct that `helloWorld` should be returning a string.

ijlee2 (approved):

ijlee2:
@nruth Would you like to make another pull request, to help out with backporting (copy-pasting) the fix to older versions of Ember Guides (versions 3.15 - 3.21)?

You can find the corresponding `native-classes-in-depth.md` file in the following folders:

- `guides/v3.15.0`
- `guides/v3.16.0`
...
- `guides/v3.21.0`

guides/release/in-depth-topics/native-classes-in-depth.md
nruth: That sounds right. I wasn't sure if that returns a string though, and there's no return statement so I thought maybe someone started off the example with an alert pop-up then decided to use console.log instead, and forgot to switch to returning the string.

nruth:
Sure, I'll give that a go now. If I see #1614 there too can I do both in one commit or is it best to keep them split up?

ijlee2:
> Sure, I'll give that a go now. If I see #1614 there too can I do both in one commit or is it best to keep them split up?

Thanks in advance for your help! Since it's the same file with not many changes, I think 1 pull request is perfect. If you'd like, you can make 2 commits, one for each different type of changes made.

PR #1620, opened 2020-12-16
Update how we access the model from the template

taltcher:
According to the [RFC](https://github.com/emberjs/rfcs/blob/master/text/0523-model-argument-for-route-templates.md#drawbacks) we should access the model using `this.model` - also this is the behavior which will be enforced by `ember-template-lint` as well: https://github.com/ember-template-lint/ember-template-lint/pull/1596

### DIFF
--- a/guides/v3.22.0/tutorial/part-1/working-with-data.md
+++ b/guides/v3.22.0/tutorial/part-1/working-with-data.md
@@ -63,7 +63,7 @@ We'll get to that bit later on. At the moment, we are just returning the same ha
 
 ## Accessing Route Models from Templates
 
-So, now that we've prepared some model data for our route, let's use it in our template. In route templates, we can access the model for the route as `@model`. In our case, that would contain the POJO returned from our model hook.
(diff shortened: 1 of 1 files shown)


ijlee2:
@taltcher Thanks for informing us of the issue with `@model`!

The changes to the Markdown file in `v3.22.0` looks good, but let me ask Robert Jackson, the creator of the linter rule, [on Discord](https://discord.com/channels/480462759797063690/480777444203429888/788877373516152863) to make sure that updating all `@model`s to `this.model` is what we need to do.

ijlee2:
@taltcher Hello again! Robert responded to my question on Discord.

What we'd like to do is to fix the reported issue with `@model` (Robert and Godfrey were unable to reproduce the issue yet), rather than to update the Ember Guides by replacing `@model` with `this.model`.

Is it okay with you if I close this PR for now, and we wait on making changes regarding `@model`?

taltcher:
@ijlee2 - sure so I'll turn off the lint rule in my project and wait for the issue to be resolved :-) Thanks!

PR #1623, opened 2020-12-30
Add empty alternate text and role presentation to blog logo example

srsgores:
This adds the missing `alt` attribute (needed for screen readers and accessibility).

### DIFF
--- a/guides/release/components/block-content.md
+++ b/guides/release/components/block-content.md
@@ -271,7 +271,7 @@ to them.
 ```handlebars
 <!-- usage -->
 <BlogPost @post={{@blogPost}} as |postBody|>
-  <img src="./blog-logo.png" alt="Blog Logo">
+  <img alt="" role="presentation" src="./blog-logo.png">
 
   {{postBody}}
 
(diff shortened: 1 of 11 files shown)


ijlee2 (changes requested):
Hi, @srsgores. Thanks for noticing that the `` element in the example is missing its `alt` attribute.

In this example, it looked to me like the `blog-logo` image is [decorative](https://www.w3.org/WAI/tutorials/images/decorative/). I believe, in such case, we will want to set an empty `alt` and a `role` of `presentation` instead:

```handlebars
<img src="./blog-logo.png" alt="" role="presentation">
```

What do you think about setting the `alt` to empty string? If this looks right to you, can we change line 274 as well as line 291 in the `block-content.md` file?

ijlee2:
@srsgores Hello. I hope that this message finds you well. 👋

I was wondering if you can help complete the pull request. Once we add the empty alternate text to lines 274 and 291, I think we're good to go!

srsgores (commented):
Should look better now

ijlee2 (approved):

PR #1628, opened 2021-01-27
Bugfix: Update the code snippet with correct name

abadri:
Update the code snippet with the correct variable name

### DIFF
--- a/guides/release/applications/run-loop.md
+++ b/guides/release/applications/run-loop.md
@@ -90,10 +90,10 @@ If we execute the following code without the run loop:
 
 ```javascript
 let profilePhoto = Image.create({ width: 250, height: 500 });
-user.set('width', 300);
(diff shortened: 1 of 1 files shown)


ijlee2 (approved):
@abadri Thank you for catching the typos and making a pull request to fix them. 💯 A couple of months ago, we changed the example from a person to a photo, and it looks like we forgot to update the rest of the code.

I noticed that, for versions 3.15 to 3.23 of the Ember Guides, we still show `user.set(...)` instead of `profilePhoto.set(...)`. Could you open another pull request to address the issue? Specifically, you will want to update the following files:

- `guides/v3.15.0/applications/run-loop.md`
- `guides/v3.16.0/applications/run-loop.md`
...
- `guides/v3.23.0/applications/run-loop.md`

PR #1630, opened 2021-01-29
[Docs] Correct some closing HTML tags

syncerax:
- Fixed some incorrectly closed `<h2>` tags in block-content.md.
- Also made the same corrections in the previous versions of the
  docs (v3.15.0-v3.23.0).


locks (approved):
Thanks for the help!

PR #1632, opened 2021-02-14
Fix missing syntax highlighting for JS snippet

cnocon:
In [this section of the Service Injection page](https://guides.emberjs.com/release/tutorial/part-2/service-injection/#toc_why-we-cant-test-windowlocationhref), the code snippet for the `super-rentals-test.js` file lacks syntax highlighting because the snippet language was set to Handlebars when it should be JavaScript:

Screen Shot 2021-02-14 at 3 29 28 AM


cnocon:
Closing due to the fact I edited an automatically generated file. I will reopen a new PR with the fix in the source.

PR #1635, opened 2021-02-18
Added a note about stubbing the router service in non-application tests

williamhaley:
https://github.com/ember-cli/ember-cli-qunit/issues/203#issuecomment-781547400

### DIFF
--- a/guides/release/testing/testing-components.md
+++ b/guides/release/testing/testing-components.md
@@ -283,7 +283,7 @@ You stub non-Ember services by using the built-in `register()` function to regis
     <div class="cta-note-body">
       <div class="cta-note-heading">Zoey says...</div>
       <div class="cta-note-message">
-        Not all services can or should be stubbed. Certain services like router should never be stubbed. If an integration or unit test relies on routing then that behavior would be better tested as an acceptance test. The behavior of internal services like router may change over time and you should not attempt to override their behavior.
+If you are thinking of stubbing the router service in a unit or integration test, consider writing an acceptance test instead. Acceptance tests let you navigate through many routes in your app, and the router does not need to be stubbed in them. If you choose to stub the router, you will need to stub multiple methods.
       </div>
     </div>
     <img src="/images/mascots/zoey.png" role="presentation" alt="">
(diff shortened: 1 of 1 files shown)


guides/release/testing/testing-components.md
```diff
@@ -276,7 +276,19 @@ module('Integration | Component | comment-form', function(hooks) {
...
 it is possible to stub these dependencies for rendering tests.
-You stub Ember services by using the built-in `register()` function to register your stub service in place of the default.
+You stub non-Ember services by using the built-in `register()` function to register your stub service in place of the default.
+
+<div class="cta">
+  <div class="cta-note">
+    <div class="cta-note-body">
+      <div class="cta-note-heading">Zoey says...</div>
+      <div class="cta-note-message">
+        Not all services can or should be stubbed. Certain services like router should never be stubbed. If an integration or unit test relies on routing then that behavior would be better tested as an acceptance test. The behavior of internal services like router may change over time and you should not attempt to override their behavior.
```
NullVoxPopuli, line 286: 🤔 

I don't know if I agree with the statement about using an acceptance test. 

You can setup the router in unit and integration tests. This probs also related to the recent feature of lazily setting up the router upon access.
The main gotchya is that the router needs a map and a transition before it'll have useful behavior

jenweber:
Thanks for your comments and work! I think there is a middle ground where we can still caution devs against mocking the router, while still making it clear that it's ok if they do. I'll add a suggestion to the prose. It's nice to help people avoid this pitfall, especially now that we think of the router as a service and not just something on a route/controller.

guides/release/testing/testing-components.md
jenweber, line 286:
```suggestion
If you are thinking of stubbing the router service in a unit or integration test, consider writing an acceptance test instead. Acceptance tests let you navigate through many routes in your app, and the router does not need to be stubbed in them. If you choose to stub the router, you will need to stub multiple methods.
```

How is this? It's not as strong as "never do this" but hopefully we can steer people in the right direction - either with a clue about multiple methods or the suggestion to write acceptance tests.

williamhaley: Much better 👍

jenweber (approved):
This looks good to me! Thanks for the edits.

@NullVoxPopuli what do you think? Good to go?

ijlee2 (approved):
Thank you for improving the testing section in the guides!

PR #1639, opened 2021-03-06
Mark v3.24.0 as LTS version

bertdeblock:


ijlee2 (approved):
Thank you!

bertdeblock:
@ijlee2 I was wondering, should 3.16 be removed again or should that list also contain all past LTS versions?

ijlee2:
Not sure, but it may be that `ltsVersions` lists all currently supported versions according to https://emberjs.com/releases/lts. That may be why we don't see 3.12, 3.8, etc. If this is indeed the case, we can remove 3.16 from the list on March 17.

Issue #1621, opened 2020-12-21
Add back meta image


amyrlam:
We've lost the meta image for guides links, e.g. https://twitter.com/Steady5063/status/1340297223733735426

<img width="595" alt="screenshot" src="https://user-images.githubusercontent.com/16869656/111861683-79df5300-8950-11eb-8364-871b10d8485b.png">

Relates to / probably same fix as: https://github.com/ember-learn/ember-blog/issues/854

ijlee2:
@nickschot @mansona Can you have a look at this issue? I checked that `e` logo does not show up still.

amyrlam:
Ah thanks for the reminder. I added it to ember-website with https://github.com/ember-learn/ember-website/pull/804, by hardcoding the path...because I noticed the emberjs and embertimes Twitter accounts sending out links with a blank image. 

I made a similar change here: https://github.com/ember-learn/guides-source/pull/1647 and will tag you 3 on that PR.

PR #1647, opened 2021-03-20
Fix meta image

amyrlam:
Resolves: https://github.com/ember-learn/guides-source/issues/1621

Related to: https://github.com/ember-learn/ember-website/pull/804

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="description" content="Ember.js helps developers be more productive out of the box. Designed with developer ergonomics in mind, its friendly APIs help you get your job done—fast.">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta property="og:description" content="Ember.js helps developers be more productive out of the box. Designed with developer ergonomics in mind, its friendly APIs help you get your job done—fast." />
     <meta property="og:image" content="https://blog.emberjs.com/images/logos/e-icon.png" />
--- a/app/index.html
+++ b/app/index.html
@@ -3,7 +3,13 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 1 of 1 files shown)


amyrlam:
I copied the contents from https://github.com/ember-learn/ember-website/pull/804/files except for ` <meta name="google-site-verification"` as wasn't sure what that was yet.

ijlee2 (changes requested):
@amyrlam Thanks for working on a solution. I checked that the image shows up when I post a Guides link in a tweet.

Before we merge the pull request, I'd like to see application tests added so that there's no regression in the future. Can we update the file `https://github.com/ember-learn/guides-source/blob/master/tests/acceptance/meta-data-test.js` in the following manner:

```javascript
test('We set title and description correctly', async function(assert) {
  // Navigate to Routing, Defining Your Routes

  // Check title and description

  // Navigate to Services, Overview

  // Check title and description
});

test('We set metadata for Open Graph correctly', async function(assert) {
  // Navigate to Routing, Defining Your Routes

  // Check title, description, and image

  // Navigate to Services, Overview

  // Check title, description, and image
});

test('We set metadata for Twitter correctly', async function(assert) {
  // Navigate to Routing, Defining Your Routes

  // Check title, description, and image

  // Navigate to Services, Overview

  // Check title, description, and image
});
```

app/index.html
```diff
@@ -3,7 +3,13 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="description" content="Ember.js helps developers be more productive out of the box. Designed with developer ergonomics in mind, its friendly APIs help you get your job done—fast.">
```
locks, line 6: This is applied by guidemaker so it shouldn't be hardcoded in here.

amyrlam:
From the Learning Team meeting: This tests should be made in guidemaker, I just made a separate issue: https://github.com/empress/guidemaker/issues/51

locks:
We reviewed this at the learning meeting and decided to move the PR forward after removing the description since it is applied by guidemaker.
There are tests in guidemaker and empress for meta tags so we felt like upstreaming to those projects made more sense than adding the tests here.
We will follow with further upstream work.

locks (approved):

PR #1654, opened 2021-03-27
Improved passage in Customizing Serializers

esquith:


ijlee2 (approved):
Good catch. I think the sentence flows better now (it could still be better, but I think the small change you suggested works well). Thank you for your help!

PR #1661, opened 2021-03-29
Remove sentence

esquith:
This should probably be removed as it will only cause confusion, in my opinion.

What do other people think?


ijlee2 (approved):
Yeah, I think it's okay to remove the sentence, since the following sentence mentions that it's included in a new Ember app by default. Thanks!

PR #1671, opened 2021-04-02
Removed 3.16 from currently supported LTS versions

ijlee2:
## Description

Since [March 17, 2021](https://emberjs.com/releases/lts), 3.16 LTS is no longer supported.

### DIFF
--- a/guides/versions.yml
+++ b/guides/versions.yml
@@ -50,6 +50,5 @@ allVersions:
   - "v3.25.0"
 currentVersion: "v3.25.0"
 ltsVersions:
-  - "v3.16.0"
   - "v3.20.0"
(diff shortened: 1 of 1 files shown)


jaredgalanis (approved):
Thank you! 👍

jaredgalanis:
> Thank you! 👍

@ijlee2 do you know, is there a new LTS being added when 3.26 is released?

ijlee2:
@jaredgalanis 3.24 is the latest LTS (this happened when 3.25 was released). The next LTS candidate will likely be 3.28, both because it's 4 minor version after 3.24 and possibly the last minor version before 4.0.

PR #1673, opened 2021-04-03
Fixed typos in Customizing Adapters

esquith:


guides/release/models/customizing-adapters.md
```diff
@@ -39,8 +39,8 @@ following file:
 
 ```javascript {data-filename=app/adapters/post.js}
 import JSONAPIAdapter from '@ember-data/adapter/json-api';
-export default class ApplicationAdapter extends JSONAPIAdapter {
-  namespace = 'api/v1';
+
+export default class PostAdapter extends JSONAPIAdapter {
```
ijlee2, line 43: **suggestion (blocking)**

The preceding paragraph, although not entirely clear, mentions overriding the default `application` adapter. As an example, I think the code showed overriding `namespace` with `api/v1`, which refers to an older version of someone's API.

```suggestion
export default class PostAdapter extends JSONAPIAdapter {
  namespace = 'api/v1';
```

ijlee2: Can you update the preceding lines, 34-38, to:

```markdown
If there is a model that does not follow the backend's typical conventions, you
can create an adapter that is specific to that model. The model-specific adapter
will override the rules in the `application` adapter.

To create a model-specific adapter, run the command `ember generate adapter <adapter-name>`.
For example, suppose there is a `post` model that needs to talk to the `v1` API
in the backend. We can run `ember generate adapter post` to create the adapter,
then specify the `post` adapter's namespace:
```

The update doesn't have to be exactly this, but I think it's a good start. If you like to suggest something else, please feel free to add a comment.

ijlee2 (changes requested):
I think we want to keep the line `namespace='api/v1';` but make the preceding paragraph more clear about why this line is shown in the code example.

Can you have a look at my suggestion?

After we merge the pull request, I think it'd be good to backport the changes to earlier Ember versions (3.15 - 3.24).

esquith:
@ijlee2 That paragraph looks good to me. Would you like me to make the PR for Ember versions 3.15 - 3.24?

ijlee2 (approved):
Thank you for having a look at Customizing Adapters!

PR #1674, opened 2021-04-05
Octanified app and added tests

ijlee2:
## Description

In preparation for updating `ember-source` to `v3.25` or higher, I added missing tests and modernized syntax.

I used the Netlify preview deploy to check that we can still navigate between pages and across different Ember versions.

## How to Review

I recommend reviewing commits one by one. Each commit makes a specific, meaningful change.


app/router.js
```diff
@@ -1,44 +1,37 @@
...
-
-import { get } from '@ember/object';
 import { inject as service } from '@ember/service';
-import { scheduleOnce } from '@ember/runloop';
+import config from 'ember-guides/config/environment';
 
-const Router = EmberRouter.extend({
-  location: config.locationType,
-  rootURL: config.rootURL,
+export default class Router extends EmberRouter {
```
ijlee2, line 5: TODO:

After updating `ember-source` to `v3.26`, try injecting the `router` service when calling `routeDidChange`.

PR #1675, opened 2021-04-05
Updated development dependencies to their latest version

ijlee2:
## Description

I wanted to update `ember-source` to `v3.25.4`, but ran into a build issue (could not run `ember s`) due to old versions of packages (e.g. `ember-basic-dropdown` and `ember-popper`) being pulled in. Chris also mentioned that `guidemaker-ember-template` is in progress of an update.

Due to these 2 factors, I reduced the scope of the pull request by updating development dependencies to their latest version. I will look into updating remark-related packages separately and open another pull request if the update can be done.


package.json
```diff
@@ -80,21 +80,21 @@
...
     "prember": "^1.0.3",
     "qunit-dom": "^1.0.0",
-    "remark-cli": "^8.0.0",
-    "remark-lint": "^7.0.0",
-    "remark-preset-lint-consistent": "^3.0.0",
-    "remark-preset-lint-recommended": "^4.0.0",
+    "remark-cli": "^9.0.0",
+    "remark-lint": "^8.0.0",
+    "remark-preset-lint-consistent": "^4.0.0",
+    "remark-preset-lint-recommended": "^5.0.0",
```
ijlee2, line 92: TODO:

Updating `remark`-related packages slowed down `test:node` script to the point that CI timed out. Investigate why in a separate issue.

ijlee2: Update: The problem seems to occur when we update `markdown-link-extractor`.

mansona:
Hey @ijlee2 👋  I'm wondering if it's worth doing much work to upgrade the Guides like this before we deploy the redesign? 🤔  this component that needs to be upgraded is going to be deleted as it will be provided by the template once the redesign goes live (and it is already fully octane-ified) https://github.com/ember-learn/guidemaker-ember-template/blob/website-redesign/addon/components/table-of-contents.hbs

ijlee2:
@mansona Other than using the `class` attribute, the `<TableOfContents>` component hasn't been changed in this pull request.

I'm okay with updating `ember-source` until later, but would like to update other dependencies to their latest version.

---

Over the weekend, I found that the currently installed version of `guidemaker-ember-template` does not work with `ember-source 3.25` (3.24, either). We'll get error messages such as,

```
Template Compiler Error (broccoli-persistent-filter:TemplateCompiler) in ember-basic-dropdown/templates/components/basic-dropdown/content.hbs

Cannot read property 'type' of undefined
```

```
Template Compiler Error (broccoli-persistent-filter:TemplateCompiler) in ember-popper/templates/components/ember-popper-targeting-parent.hbs

loc$$1.sliceStartChars is not a function
```

I think it's cause older versions of `ember-bootstrap` and `ember-power-select` pull in dependencies that don't work with 3.24 & 3.25. I'm guessing that this isn't a problem in the `website-redesign` branch and the next release of `guidemaker-ember-template`?

mansona:
Yes that's correct 👍  bootstrap is a thing of the past in the new redesign and we are already developing things using ember 3.25 with no issues: https://github.com/ember-learn/guidemaker-ember-template/blob/website-redesign/package.json#L68

mansona (approved):
Looks good 👍

locks:
I'm doing the 3.26 release, please hold until it's out 😁

ijlee2:
@locks Ok.

locks (approved):
Release done 👍

PR #1689, opened 2021-05-25
Make backburner docs more explicit

MichalBryxi:
I'm not entirely sure that this is the right place to put it, but current wording is pretty vague.

Do I need to add it to my application code? To my tests? In `beforeEach` of my tests?

Also current version makes an assumption that the user knows that they need to import `Ember`.

### DIFF
--- a/guides/release/configuring-ember/debugging.md
+++ b/guides/release/configuring-ember/debugging.md
@@ -106,8 +106,10 @@ this is quite slow and is not appropriate for production or even normal developm
 To enable full stacktrace mode in Backburner, and thus determine the stack of the task
 when it was scheduled onto the run loop, you can set:
 
-```javascript
-Ember.run.backburner.DEBUG = true;
+```javascript {data-filename=app/app.js}
+import { run } from '@ember/runloop';
+
+run.backburner.DEBUG = true;
 ```
 
 Once the `DEBUG` value is set to `true`, when you are at a breakpoint you can navigate


MichalBryxi:
Based my PR on [this issue](https://github.com/emberjs/ember.js/issues/13074)

jenweber:
Hi @MichalBryxi, we should not show `import Ember`, and instead show importing from `@ember/runloop`. Can you also make sure that works as you expect in your app code? Here are the docs and syntax: https://api.emberjs.com/ember/3.26/functions/@ember%2Frunloop/run

Thank you!

MichalBryxi:
Thanks for correction. Force pushed as you suggested @jenweber and tried to verify that the code works:

```js
// /app/app.js

import { run } from '@ember/runloop';

run.backburner.DEBUG = true;
```

```js
// /app/controllers/application.js

import { run } from '@ember/runloop';

  @action
  foo() {
    run.later(() => {
      debugger;
    });
  }
}
```

Screenshot 2021-05-31 at 21 06 17

Seems ok to me?

mansona:
So we looked at this in the ember learning team meeting today and we wondered if this method still worked in Ember 3.27 (since there were so many module changes)

I tested it and it looks like it's got the same behaviour that you describe @MichalBryxi 🎉 so I'm happy for us to merge this one 💪 

Screenshot 2021-07-22 at 17 37 45

mansona (approved):
Looks great! 🎉  thanks for doing this

PR #1691, opened 2021-05-27
Fix Object property name

j000shDotCom:
Minor edit correcting a property name on an address Object, `this.address.state` to `this.address.region`


ijlee2 (approved):
Thank you for the corrections!

PR #1696, opened 2021-06-17
Remove redundant import

saracope:
Merging this PR will:
- Remove line 204 in the _Overriding HTML Attributes in ...attributes_ example for `map-test.js`
`import { render } from '@ember/test-helpers';`

Reasoning:
- The following import statement includes the render and find which makes line 204 unnecessary
`import { render, find } from '@ember/test-helpers';`


saracope:
It looks like this just wasn't displaying for me properly, so closing out!

PR #1697, opened 2021-06-17
add SPA netlify redirect

mansona:
I noticed when working on the design that the 404 page doesn't seem to be working on the guides for some reason 🙈  turns out we didn't setup the correct netlify redirects for SPAs

### DIFF
--- a/.netlifyredirects
+++ b/.netlifyredirects
@@ -1,2 +1,3 @@
 /                 /release/
 /current/*        /release/:splat
+/*                /index.html       200


mansona:
Ah... maybe there was a reason we didn't include the redirects 🤔  it looks like putting this at the top of the file actually cases most of the other redirects to stop working 😞 

We should figure out a way to make this work effectively

jenweber:
I'm doing some work to clean up loose ends and stale PRs. I think we should close this, and reopen if we resume the work.

PR #1703, opened 2021-07-21
Fix nested code block indentation

hellomattio:
Some code blocks on the `native-classes` page are nested inside an unordered list.

For the code block indentation to work correctly, an indentation of `4` spaces is needed for every level of the list - there are currently `2`, which is causing the first degree of indentation in each code block to be lost (example shown below):

Screen Shot 2021-07-21 at 12 45 38 pm

Live example: [as-is](https://guides.emberjs.com/release/upgrading/current-edition/native-classes/#toc_instantiation) vs [proposed](https://deploy-preview-1703--ember-guides.netlify.app/release/upgrading/current-edition/native-classes/#toc_instantiation).

### DIFF
--- a/guides/release/upgrading/current-edition/native-classes.md
+++ b/guides/release/upgrading/current-edition/native-classes.md
@@ -450,11 +450,11 @@ syntax and _not_ extending from `EmberObject` at all in your apps.
 - Use `class` when defining a class, and `class ... extends` when extending a
   class.
 
-  ```js
-  class Person {}
+    ```js
+    class Person {}
 
-  class Actress extends Person {}
-  ```
+    class Actress extends Person {}
+    ```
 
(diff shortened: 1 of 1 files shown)


ijlee2 (approved):
Hello, thank you very much for fixing the code indentations in the Native Classes page!

If you have time, in a separate pull request, could you backport the changes to previous Ember versions (3.15 to 3.26)? Since 156 lines changed, I think we could try the following:

- Copy-paste the Markdown file from `release` to `v3.15`. Save and commit.
- Check that only 156 lines changed in the commit, and that the changes are indentation-related.
- If the check above passes, we can copy-paste the Markdown file to `v3.16`, ..., `v3.26` (in a 2nd commit).

What are your thoughts?

hellomattio:
> If you have time, in a separate pull request, could you backport the changes to previous Ember versions (3.15 to 3.26)? 

Sounds good @ijlee2, I'll do so shortly.

PR #1704, opened 2021-07-21
Backport code block indentation

hellomattio:
A follow-up to https://github.com/ember-learn/guides-source/pull/1703, backporting the change made in `current-edition` to `v3.15`, ..., `v3.26`.

> Some code blocks on the `native-classes` page are nested inside an unordered list.

> For the code block indentation to work correctly, an indentation of `4` spaces is needed for every level of the list - there are currently `2`, which is causing the first degree of indentation in each code block to be lost (example shown below):


hellomattio:
@ijlee2 as discussed.

ijlee2 (approved):
Awesome, much thanks!

PR #1714, opened 2021-08-13
Remove Yarn references and update pinned Node.js version

bertdeblock:


package.json
```diff
@@ -110,7 +110,6 @@
     "edition": "octane"
   },
   "volta": {
-    "node": "10.16.3",
-    "yarn": "1.22.4"
+    "node": "12.19.1"
```
NullVoxPopuli, line 113: Should at npm version be pinned here?

bertdeblock: We can do that yes! Which version should it be?

NullVoxPopuli: I usually start with latest, `volta pin npm`

bertdeblock: Pinned an mentioned in the README.

PR #1715, opened 2021-08-13
Add information on how to use injection in native classes

LeopoldHock:
The article on native classes should include information on how to access injection (or possibly other framework features). I provided an example and wrote about what's going on to the best of my abilities, but take it with a grain of salt. :)
@NullVoxPopuli


NullVoxPopuli:
I'm a big fan of this! 🎉

Miexil (approved):

phndiaye (approved):

guides/release/in-depth-topics/native-classes-in-depth.md
```diff
@@ -494,6 +494,49 @@ will be described in greater detail later on in the guides.
...
+  }
+}
+
+// On any framework object...
+let item = new Item();
+setOwner(item, getOwner(this));
+item.addToCart();
+```
+
+Alternatively, you can call `setOwner` in the class constructor and simply supply the caller as an argument to the constructor.
```
chancancode, line 518: I think setting it first thing in the constructor (before any of the injections were accessed) is the only form that is guaranteed to work. If the owner relationship is not established at initialization time things may behave unexpectedly and the wrong results may get cached.

NullVoxPopuli:
anything keeping this from being merged? I'd like to see a list of blockers, otherwise let's move forward / merge?

jenweber:
The blocker on this PR was that I would like to hear from someone on the Framework Core Team to confirm that this content should be included. @chancancode does this seem ok to you?

(sorry everyone else for the ping, I'm doing some work on old PRs/loose threads this week!)

chancancode:
I think it's totally good to include it in the in-depth guide. I do still think we should either remove the setting-owner-after-construction pattern or at the very least deprioritize it. The general expectation is that the owner will be set by the time the class has been "constructed completely".

For framework classes this is generally enforced by making the owner a required constructor argument on the base class (and the owner is set for you there). Since JavaScript enforces that you _must_ call `super()` very early in your constructor, this is usually enforced automatically.

For non-framework classes, I think it's still generally good to follow this pattern/expectation. I cannot personally guarantee (i.e. I haven't asked others who were more involved in designing this and re-reading the relevant RFCs) that it is considered "supported" to set the owner outside of the constructor before "touching" `this`. I believe the intention is to consider setting the owner conceptually part of the object allocation/initialization, we just don't really have tools to enforce it unlike JavaScript.

So if we just want to add the section on setting it in the constructor, which covers a lot of the use cases, I think it's a no-brainer. If it is for some reason very important to discuss/prioritize the other part of this PR then I am probably not the best person to review it.

guides/release/in-depth-topics/native-classes-in-depth.md
```diff
@@ -494,6 +494,49 @@ will be described in greater detail later on in the guides.
 > you exercise caution if using any external decorator libraries which may not
 > have the same stability guarantees.
 
+### Using injection
```
jenweber, line 497:
```suggestion
### Using injection

```
I am only making this change in order to force CI to re-run.

jenweber:
Hello @LeopoldHock, thanks again for writing this! And thanks to our reviewers. I appreciate everyone's patience. This is a little outside my knowledge of Ember and I wanted to be sure before signing off on it. Merged.

jenweber (approved):

Issue #1719, opened 2021-08-26
Outdated rootElement management example


lifeart:
![image](https://user-images.githubusercontent.com/1360552/130924746-cfb0f945-5efe-4e2f-bf28-22fb144bb471.png)

https://guides.emberjs.com/release/configuring-ember/embedding-applications/

Latest blueprint:

![image](https://user-images.githubusercontent.com/1360552/130924812-73fa25e2-72bd-48b8-9f5c-7d72fc762214.png)

locks:
This seems to be fixed! Let me know if I missed anything.

PR #1724, opened 2021-08-27
update code examples to use classes

lifeart:
related to: https://github.com/ember-learn/guides-source/issues/1719


locks (approved):
Thanks for the update!

PR #1737, opened 2021-09-09
Remove tmp  during search index creation

jenweber:
The script was creating too many indices for algolia.
We discovered that a tmp dir was holding all the old
versions. You can reproduce this bug by deleting dist,
deleting some versions, and running ember build.
The resulting dist had all the versions, including
those that were deleted.

Co-authored by @mansona

### DIFF
--- a/scripts/update-search-index
+++ b/scripts/update-search-index
@@ -62,6 +62,11 @@ rm -r dist
 echo "  DONE"
 
 echo
+echo "🤖 Deleting tmp folder"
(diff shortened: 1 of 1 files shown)


jenweber:
@locks good catch that we were making too many indices for search. I think this should fix it.

locks (approved):
A-ha! Good catch!

ijlee2:
Can we merge this pull request?

PR #1748, opened 2021-11-05
Update conditional-content.md

MelSumner:
If merged, this PR adds an example for concatenation in an inline conditional, common in CSS classes.

### DIFF
--- a/guides/release/components/conditional-content.md
+++ b/guides/release/components/conditional-content.md
@@ -11,7 +11,7 @@ There are 2 styles of `if`: **block** and **inline**.
 </div>
 ```
 
-Additionally, you concatenate a conditional like this: 
+Additionally, you can use template helpers like `concat` within a conditional. For the example below, if `@color` has a truthy value, such as `'navy'`, the div classes will include `badge-navy`:
 
 ```handlebars
 <div class="badge {{if @color (concat 'badge-' @color)}}">
--- a/guides/release/components/conditional-content.md
+++ b/guides/release/components/conditional-content.md
@@ -11,6 +11,13 @@ There are 2 styles of `if`: **block** and **inline**.
 </div>
 ```
 
+Additionally, you concatenate a conditional like this: 
+
+```handlebars
+<div class="badge {{if @color (concat 'badge-' @color)}}">
+  Badge Text
+</div>
+```
 
 ## Block `if`
 


guides/release/components/conditional-content.md
```diff
@@ -11,6 +11,13 @@ There are 2 styles of `if`: **block** and **inline**.
 </div>
 ```
 
+Additionally, you concatenate a conditional like this:
```
jenweber, line 14:
```suggestion
Additionally, you can use template helpers like `concat` within a conditional. For the example below, if `@color` has a truthy value, such as `'navy'`, the div classes will include `badge-navy`:
```

Thanks for the PR! What do you think about these suggestions? I think it's good to tie this into the general approach of helpers. Helpers have not been introduced yet at this point in the guides, but there's a section dedicated to them within the same subheading. Possibly the "template helpers" text should be a link.

jenweber: One more thought, maybe this should go at the bottom? It could be a new section, "Combining if with other helpers." In looking at the full article, it seems like we should stick to the basics right at the very top of the article. I'm open to whatever you think is best!

MelSumner:
Jen that makes sense too, I’m fine with changing it. I just ran into
needing the syntax today & saw that it wasn’t here. I’ll make the edits
in the morning. 👍

On November 5, 2021, Howie ***@***.***> wrote:
> @jenweber commented on this pull request.
>

> In guides/release/components/conditional-content.md
> <https://github.com/ember-learn/guides-
> source/pull/1748#discussion_r744014294>:
>

> @@ -11,6 +11,13 @@ There are 2 styles of `if`: **block** and
**inline**. </div> ``` +Additionally, you concatenate a conditional like
this:
> One more thought, maybe this should go at the bottom? It could be a
> new section, "Combining if with other helpers." In looking at the full
> article, it seems like we should stick to the basics right at the very
> top of the article. I'm open to whatever you think is best!
>

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ember-learn/guides-
source/pull/1748#discussion_r744014294>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/ABC77O6GPC2ZZGEUJN6AKMDUKRUSXANCNFSM5HOF6N2A>.

locks (approved):

PR #1750, opened 2021-11-30
Update documentation for get helper

MahemaS27:
This PR is motivated by a question that was surfaced by [this issue](https://github.com/emberjs/ember.js/issues/19519) in the [emberjs repo ](https://github.com/emberjs/ember.js).  

The `get` helper use can be extended to objects with string or numeric keys, so arrays can also make use of the helper. This PR updates the docs to have examples of all of these use cases and have a more generic definition of what the `get` helper can do.

### DIFF
--- a/guides/release/components/helper-functions.md
+++ b/guides/release/components/helper-functions.md
@@ -208,9 +208,21 @@ For the full list of available helpers, you can check the [template helpers API 
 ### The `get` helper
 
 The [`{{get}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/get?anchor=get)
-helper makes it easy to dynamically send the value of a variable to another
-helper or component. This can be useful if you want to output one of several
-values based on the result of a getter.
+helper makes it easy to dynamically look up a property on an object. The second argument to `{{get}}` can be a string or an integer depending on the object being accessed.
+
+
+accessing properties on an object with string keys:
+```handlebars
+{{get this.someObject "object_key"}}
+```
+accessing via indices:
+
+```handlebars
+{{get this.someObjectWithNumericKeys 0}}
+
+{{get this.someArray 0}}
+```
+or outputting one of several values based on result of a getter:
 
 ```handlebars
 {{get this.address this.part}}
@@ -218,6 +230,7 @@ values based on the result of a getter.
 
 If the `part` getter returns "zip", this will display the result of `this.address.zip`.
 If it returns "city", you get `this.address.city`.
+
 
 ### The `concat` helper
 
--- a/guides/release/components/helper-functions.md
+++ b/guides/release/components/helper-functions.md
@@ -208,20 +208,24 @@ For the full list of available helpers, you can check the [template helpers API 
 ### The `get` helper
 
 The [`{{get}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/get?anchor=get)
-helper makes it easy to dynamically look up a property on an object. The second argument to `{{get}}` can be a string or an integer depending on the object being accessed.
+helper makes it easy to dynamically look up a property on an object or an element in an array. The second argument to `{{get}}` can be a string or an integer depending on the object being accessed.
 
 
 accessing properties on an object with string keys:
 ```handlebars
 {{get this.someObject "object_key"}}
 ```
-accessing via indices:
+accessing properties on an object with numeric keys:
 
 ```handlebars
 {{get this.someObjectWithNumericKeys 0}}
+` ` `
 
+accessing elements in an array:
+
+```handlebars
 {{get this.someArray 0}}
-```
+` ` `
 or outputting one of several values based on result of a getter:
 
 ```handlebars
--- a/guides/release/components/helper-functions.md
+++ b/guides/release/components/helper-functions.md
@@ -208,25 +208,22 @@ For the full list of available helpers, you can check the [template helpers API 
 ### The `get` helper
 
 The [`{{get}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/get?anchor=get)
-helper makes it easy to dynamically look up a property on an object or an element in an array. The second argument to `{{get}}` can be a string or an integer depending on the object being accessed.
+helper makes it easy to dynamically look up a property on an object or an element in an array. The second argument to `{{get}}` can be a string or a number, depending on the object being accessed.
 
 
-accessing properties on an object with string keys:
+To access a property on an object with a string key:
+
 ```handlebars
 {{get this.someObject "object_key"}}
 ```
-accessing properties on an object with numeric keys:
 
-```handlebars
-{{get this.someObjectWithNumericKeys 0}}
-` ` `
-
-accessing elements in an array:
+To access the first element in an array:
 
 ```handlebars
 {{get this.someArray 0}}
-` ` `
-or outputting one of several values based on result of a getter:
+```
+
+To access a property on an object with a dynamic key:
 
 ```handlebars
 {{get this.address this.part}}


ijlee2 (changes requested):
Hi, @MahemaS27. Thanks for opening a pull request.

I understand that a lack of documentation for `{{get this.someArray 0}}` may cause a developer to be stumped (maybe when working on a legacy codebase).

However, I don't think the examples to be added are good practices that we should advocate in the Ember Guides. The API Guides specify that the 2nd argument must be a string, so `{{get this.someArray "0"}}` may be the correct usage.

The `{{get}}` helper is most useful when the key is dynamic. When the key is static, there are better, more readable ways to access the value:

```handlebars
{{!-- get helper is unnecessary --}}
{{get this.someObject "someProperty"}}

{{!-- My recommendation --}}
{{this.someObject.someProperty}}
```

```handlebars
{{!-- Proposed --}}
{{get this.someArray "0"}}

{{!-- Recommendation (1): Create a getter in the backing class and provide a meaningful name --}}
{{this.firstArrayElement}}

{{!-- Recommendation (2): Use {{object-at}} helper from ember-composable-helpers, or create a custom helper --}}
{{object-at 0 this.someArray}}
{{first-element this.someArray}}
```

As a result, I'd like to propose that we close the pull request. What do you think of the approaches that I provided above?

jamescdavis:
IMHO the API docs should be updated to indicate that the 2nd argument can be a number. This has been supported since Ember 2.14 and was added in https://github.com/emberjs/ember.js/pull/15366 (along with a test to make sure it continues to work). This test [still exists in Ember 4](https://github.com/emberjs/ember.js/blob/cf1f34e2c5d94ea0797e8f27b472372a07443b95/packages/%40ember/-internals/glimmer/tests/integration/helpers/get-test.js#L62-L85). Furthermore https://github.com/emberjs/ember.js/pull/16218 added a test specifically for array index access, which is [also still present](https://github.com/emberjs/ember.js/blob/cf1f34e2c5d94ea0797e8f27b472372a07443b95/packages/%40ember/-internals/glimmer/tests/integration/helpers/get-test.js#L87-L105). There are also tests for accessing objects and arrays using a path that _evaluates_ to a number. It's clear that `get` is intended to work with a number for the second argument.

jamescdavis:
API Docs source PR documenting numeric second arg and array element access: https://github.com/emberjs/ember.js/pull/19871

guides/release/components/helper-functions.md
```diff
@@ -208,9 +208,21 @@ For the full list of available helpers, you can check the [template helpers API
 ### The `get` helper
 
 The [`{{get}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/get?anchor=get)
-helper makes it easy to dynamically send the value of a variable to another
-helper or component. This can be useful if you want to output one of several
-values based on the result of a getter.
+helper makes it easy to dynamically look up a property on an object. The second argument to `{{get}}` can be a string or an integer depending on the object being accessed.
```
jamescdavis, line 211:
```suggestion
helper makes it easy to dynamically look up a property on an object an element in an array. The second argument to `{{get}}` can be a string or an integer depending on the object being accessed.
```

guides/release/components/helper-functions.md
```diff
@@ -208,9 +208,21 @@ For the full list of available helpers, you can check the [template helpers API
...
+```handlebars
+{{get this.someObject "object_key"}}
+```
+accessing via indices:
+
+```handlebars
+{{get this.someObjectWithNumericKeys 0}}
+
+{{get this.someArray 0}}
+```
```
jamescdavis, line 224:
```suggestion
accessing properties on an object with numeric keys:

```handlebars
{{get this.someObjectWithNumericKeys 0}}
` ` `

accessing elements in an array:

```handlebars
{{get this.someArray 0}}
` ` `
```
(Note: I had to put spaces between the closing  ```s to keep it from closing the suggestion block)

jamescdavis (commented):
A couple of suggestions.

MahemaS27:
Hi @ijlee2 , thank you for taking a look at my PR!

As @jamescdavis mentioned, this seems to be a supported behavior for `get` and has tests as well. Personally, I found the use of `"0"` to be confusing, as the integer 0 is an index of an array rather than a traditional key.

In regards to the examples you gave for alternative use, I think those are helpful and provide different ways to use the `get` helper. I would say that in the case of a simple array, its nice to rely on template logic rather than have to add more logic to a js or ts file. Also, array access via numeric index is very recognizable to most developers (for that specific case). I am fairly new to ember, so I was wondering why this is not considered best template practice? Thank you!

guides/release/components/helper-functions.md
```diff
@@ -208,9 +208,25 @@ For the full list of available helpers, you can check the [template helpers API
 ### The `get` helper
 
 The [`{{get}}`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/get?anchor=get)
-helper makes it easy to dynamically send the value of a variable to another
-helper or component. This can be useful if you want to output one of several
-values based on the result of a getter.
+helper makes it easy to dynamically look up a property on an object or an element in an array. The second argument to `{{get}}` can be a string or an integer depending on the object being accessed.
```
ijlee2, line 211: Although integer is technically correct, I think number may be easier to understand. I added a comma, to help break the sentence.

```suggestion
helper makes it easy to dynamically look up a property on an object or an element in an array. The second argument to `{{get}}` can be a string or a number, depending on the object being accessed.
```

guides/release/components/helper-functions.md
```diff
@@ -208,9 +208,25 @@ For the full list of available helpers, you can check the [template helpers API
...
+```handlebars
+{{get this.someObjectWithNumericKeys 0}}
+` ` `
+
+accessing elements in an array:
+
+```handlebars
+{{get this.someArray 0}}
+` ` `
+or outputting one of several values based on result of a getter:
```
ijlee2, line 229: I think @jamescdavis' suggestion was to add a new line around the code blocks. Can we begin each example with "To ..."?

``````suggestion
To access a property on an object with a string key:

```handlebars
{{get this.someObject "object_key"}}
```

To access the first element in an array:

```handlebars
{{get this.someArray 0}}
```

To access a property on an object with a dynamic key:
``````

Note, I removed the example of object with numeric key. I don't think [numeric key is possible](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors#property_names).

**Comment by ijlee2** [comment]
> In regards to the examples you gave for alternative use, I think those are helpful and provide different ways to use the `get` helper. I would say that in the case of a simple array, its nice to rely on template logic rather than have to add more logic to a js or ts file. Also, array access via numeric index is very recognizable to most developers (for that specific case). I am fairly new to ember, so I was wondering why this is not considered best template practice? Thank you!

@MahemaS27 Yep, accessing an array element with an index is common. I guess I've associated the `{{get}}` helper with accessing a POJO. As a result, `{{get this.someArray 0}}` felt like writing in JavaScript,

```javascript
const { 0: elementOfInterest } = this.someArray;

return elementOfInterest;
```

But, of course, I'd prefer that we write,

```javascript
const elementOfInterest = this.someArray[0];

return elementOfInterest;
```

I think a reason for using a getter or a helper such as `{{first-element}}` is readability. Code such as `{{get this.someArray 0}}` may not immediately make sense to everyone (i.e. what does 0 mean, in terms of syntax and maybe business logic? is an array zero- or one-indexed?). It is my perception that, with Octane, we prefer code that more people (designers, developers, and others) can understand.

ijlee2:
@MahemaS27 @jamescdavis Could one of you help with updating the API documentation for `{{get}}` as a follow-up?

jamescdavis:
@ijlee2

> Could one of you help with updating the API documentation for `{{get}}` (that a number is also supported) as a follow-up?

https://github.com/ember-learn/guides-source/pull/1750#issuecomment-986503273

guides/release/components/helper-functions.md
MahemaS27: that sounds good to me!

ijlee2 (approved):
✨

PR #1766, opened 2022-01-09
Use service named import instead of inject

locks:
Updated to be inline with [RFC 752](https://emberjs.github.io/rfcs/0752-inject-service.html).


jaredgalanis (approved):
looks like there's one CI issue, otherwise seems good. 👍

locks:
Ah shoot, I updated pages that belong to the tutorial.

jaredgalanis (approved):

PR #1770, opened 2022-01-23
add couple of remarks regarding inline/block usage of -if-

ciur:
Add couple of remarks regarding usage of block and inline version of `if`.


locks:
Thank you for taking the time to contribute!
I noticed a couple of things with your suggestions. Generally, we try to focus content on the happy path (the working code examples) because people tend to miss the "won't work" wording and copy paste the code and get confused.
Another thing I noticed is that in the first example you could (and should) do `<button disabled={{@disabled}} class="btn">Sign In</button>` which is much more direct and does not require the repetition.

ciur:
> What do you think of improving the API documentation for {{if}} and putting a note in the Guides to check https://api.emberjs.com/ember/4.1/classes/Ember.Templates.helpers/methods/if?anchor=if for the full documentation?

@locks do I understand correctly that you want me move added remarks to API documentation ? or do you want me to delete remarks entirely and just to point user to the already exiting API docs ?

jenweber:
@locks can you take another look at this, and ciur's comments above? Thanks!

guides/release/components/conditional-content.md
```diff
@@ -107,6 +107,24 @@ Like many programming languages, Ember also allows you to write `if else` and
...
+
+Correct usage:
+
+```handlebars {data-filename="app/components/sign-in.hbs"}
+  {{#if @disabled}}
+    <button disabled class="btn">Sign In</button>
+  {{else}}
+    <button class="btn">Sign In</button>
+  {{/if}}
+```
```
jenweber, line 127:
```suggestion
The block form of the `if` statement is typically used to wrap
HTML elements or another block. If you want to use `if` inside of an HTML element, keep reading to learn about how to use inline `if` instead.

Here's an example of a block `if`, wrapping some HTML elements:

```handlebars {data-filename="app/components/sign-in.hbs"}
  {{#if @disabled}}
    <button disabled class="btn">Sign In</button>
  {{else}}
    <button class="btn">Sign In</button>
  {{/if}}
```
```

The reason for this suggested change is that we have found that people will copy-paste non-working examples and get confused. Let's try showing only working examples here.

File: guides/release/components/conditional-content.md [hunk 2]
```diff
@@ -218,6 +236,30 @@ It looks similar to a ternary operator.
...
+
+```handlebars {data-filename="app/components/spinner.hbs"}
+  {{!-- Won't work (!) --}}
+  <span {{if @disabled 'disabled' }}">
+  </span>
+```
+
+Ember will get confused and will complain with an error. In example above
+Ember will see `if` as modifier instead and complain about incorrect usage of
+modifier.
```
jenweber, line 261:
```suggestion
Similarly to block `if`, you need to pay attention where inline `if` is placed.
Inline `if` can be used only inside attribute values
of HTML elements. For example:

```handlebars {data-filename="app/components/spinner.hbs"}
  <span class="spinner {{if @inProgress 'visible' 'invisible'}}">
  </span>
```

In the example above, inline ``if`` is correctly used inside the `class` attribute value.
```

Let's show the correct examples only, and remove incorrect examples, to prevent copy-paste confusion.

**Comment by jenweber** [comment]
Hi @ciur thanks for bringing this up! The learning team re-reviewed this PR and we have some suggested changes for you to consider.

One thing that jumped out to me is that with your code example additions, we now give some more real-life examples. We've heard before that people had trouble finding a full example of if-else if - else, so this will be a nice improvement!

File: guides/release/components/conditional-content.md [hunk 3]
```diff
@@ -218,6 +229,14 @@ It looks similar to a ternary operator.
 {{if condition value1 value2}}
 ```
 
+Similarly to block `if`, you need to pay attention where inline `if` is placed.
+Inline `if` can be used only inside attribute values
```
bertdeblock, line 233: > Inline `if` can be used only inside attribute values of HTML elements.

That's not true.
Literally the example above shows valid usage that's unrelated to HTML attributes.
These are valid examples as well:

Conditional argument value:

```hbs
<Foo @foo={{if this.bar "bar" "baz"}} />
```

Applying a modifier conditionally:

```hbs
<button type="button" {{(if @onClick (modifier this.on "click" @onClick))}} />
```

meanking (approved):

PR #1772, opened 2022-01-27
Remove references to Bower

bertdeblock:
Since the [Bower RFC implementation](https://github.com/ember-cli/ember-cli/pull/9707) has been merged (I assume it will be part of the upcoming `v4.2` release), I think we can start removing references to Bower from the guides?


locks:
Hi! I understand the motivation but I think the team tries to avoid "deprecation by obscurity". Given that Bower will still be around for the entire 4.x cycle, I think we should instead suggest not to use it, and point to the deprecation guide or some resource that explains how to replace it?

locks:
Following the discussion in the other Bower issue, let's get this merged!

locks (approved):

PR #1774, opened 2022-02-02
Remove useless `computed` import

ctjhoa:


locks (approved):
Thanks for the eagle eye!

Issue #1777, opened 2022-02-04
Dev tools - index deploy script needs to skip pre-built guides


jenweber:
When we build Algolia search indexes, it looks at what is in `public`, including our pre-built files, and tries to make indices for those too. We don't want that.

Options:
- The best way to handle this is in Guidemaker - would be good to have a feature where you can just say what version to deploy.
- Add a step to the search index script that deletes the contents of `public/v*`

locks:
Fixed by https://github.com/ember-learn/guides-source/pull/1778.

PR #1778, opened 2022-02-04
Bugfix - search index script omits prebuilt guides

jenweber:
This fixes a bug in search index deploy scripts.
We pre-build some older versions of the guides to make our
builds faster. Our search index scripts were picking up on
these, making everything slow and overwriting a lot of our
indexes every time we deployed. Now, we delete
those pre-built versions before running the index.

#1777


locks:
Wow, nice fix!

PR #1779, opened 2022-02-04
Update disabling prototype extensions guide for v4

chriskrycho:
Ember v4 removed the `String` and `Function` prototype extensions, so the docs about both are now confusing and misleading. Remove those, andupdate the guidance for `Array` prototype extensions to describe the native `Array` and `TrackedArray` alternatives.


locks (approved):
Thank yyooouu!

PR #1780, opened 2022-02-09
Add a mention on how to do end to end tests

artgo:
Introduce end-to-end testing based on [testRigor](https://www.testrigor.com/)


artgo:
@ijlee2 what do you think?

jenweber:
Hello! We reviewed this PR during a recent learning team meeting, and it would be best if the topic of end to end testing was introduced more generally, with testRigor mentioned as an option alongside other similar services. What do you think?

Also, does testRigor and the company that makes it use Ember at all internally? If so, it would be great to highlight that. Thanks!

bertdeblock:
This just seems like self-promotion. Most (probably all?) upvotes are from people working at testRigor. I would close.

meanking (approved):

PR #1793, opened 2022-03-03
revise instructions for runloop debugging

nruth:
I haven't used the feature so don't know if it's working, or whether accessing an _variable is a good idea, but the previous steps didn't work for me during/after an Ember/cli 4.1 upgrade.


nruth:
I'm not sure this is actually working, it just gives me another backburner trace.

PR #1794, opened 2022-03-03
Swap ordering of named args and angle brackets cod mods

jaredgalanis:
- the previously mentioned order (angle brackets and then named args) in the refactoring checklist would potentially cause the angle brackets code mods to miss some modifications if named argument syntax wasn't corrected first
- by running the named args code mod first this issue can be avoided

@locks one hypo we just discussed in the learning meeting and should answer before merging this PR is: as to why the code mod didn't work is that they may have not been running the code mods with telemetry/apps running. Do you know if they were or weren't?
cc: @jenweber


locks:
> Do you know if they were or weren't?

I don't think https://github.com/ember-codemods/ember-native-class-codemod has a way to run without telemetry?

jenweber (approved):
We've heard that at least one large app took this approach and it worked great. Merging.

PR #1796, opened 2022-03-08
Fix typo

Mithrilhall:
from: The end result looks exactly the same before
to: The end result looks exactly the same as before:


locks (approved):
Thanks for the keen eye!

locks:
@Mithrilhall oops, I only noticed after approving that this one of the auto-generated Tutorial files. Could you do the equivalent change in https://github.com/ember-learn/super-rentals-tutorial/blob/main/src%2Fmarkdown%2Ftutorial%2Fpart-2%2F11-ember-data.md ?

Mithrilhall:
Will do.

PR #1797, opened 2022-03-15
Update patterns-for-actions.md

MelSumner:
noticed that there's an editor's note in the live guides. 

(also curious about what content would have gone here...)


jenweber (approved):
Thank you!

PR #1798, opened 2022-03-17
Remove Ember Language Server from the list of Visual Studio Code's extensions

kaermorchen:
`Ember Language Server` is not under active maintaining, but `Unstable Ember Language Server` is. 
`UELS` includes all `ELS` features and has a lot [more](https://github.com/lifeart/ember-language-server#features).
Right now `UELS` is the main `VSCode` extension for Ember developers.

If `UELS` will be in the first place, new Ember developers will use it and have more comfortable experience.


locks:
Can you remove Ember Language Server instead?

jenweber:
Thanks for this PR! Just left one wording suggestion, then I think this is good to go.

jenweber (approved):

PR #1802, opened 2022-04-08
Fix Ember Data call out regarding store

locks:
The implicit injection has been deprecated, so the callout is factually wrong.

After we land on the final wording we want, this should be backported to at least 4.0 (taking care to make sure `import { service }` is supported) and maybe even down to whenever the implicit injection deprecation was released?


jenweber (approved):

jenweber:
Thank you! I believe the new service import is from 4.1 onwards.

PR #1809, opened 2022-04-19
guides: fix "tools for deprecations" links

geneukum:
Throughout all of the guides since v3.15.0, we have a broken link to the
"/ember-inspector/deprecations" page. Let's go through and fix this
retroactively so consumers of the guides can find the content that they're
looking for a little bit easier.


jenweber (approved):
Thank you for your work to fix this and for including earlier versions! We do not often go back and edit earlier versions, but we definitely want to do that for broken links.

PR #1816, opened 2022-05-05
guides: Fix diff lines for adding NewMessageInput

mduggan:
A very minor issue I noticed while reading https://guides.emberjs.com/release/components/introducing-components/

Previously the + appeared on the /form line - it should appear on the NewMessageInput.

### DIFF
--- a/guides/release/components/introducing-components.md
+++ b/guides/release/components/introducing-components.md
@@ -282,7 +282,7 @@ We have one last component to extract. Let's pull out the new message input.
 
 And include it in our `application.hbs` file.
 
-```handlebars {data-filename="app/templates/application.hbs" data-diff="-6,-7,-8,-9,-10,-11,+12"}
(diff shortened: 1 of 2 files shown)


bertdeblock (approved):

MinThaMie:
@mduggan when I look at your code I think it's correct but the netlify preview does not show that it's fixed. Are you able to run it locally and have a screenshot that it works?
Screenshot 2022-05-19 at 10 06 14

jenweber:
I see why. This change needs to be made in the `release` folder and not just 4.2. The path to the file to change is: `release/components/introducing-components.md`. After that, this can be merged.

You can keep the changes to 4.2 too if you want.

Thanks for this fix!

mduggan:
Thanks for catching that - for whatever reason I was only looking at the 4.2 docs when I made the update.  Updated the `release` version too.

MinThaMie (approved):
Thanks for fixing it in both places :)

PR #1818, opened 2022-05-07
Fix for _redirect file location in quick start docs

huntlyc:
Needs to be in /public, instead of the root folder to work.


bertdeblock (approved):

acorncom (approved):
Agreed, that's accurate. Thanks!

PR #1821, opened 2022-05-28
Backport code sample diff

locks:
A code sample diff was fixed in https://github.com/ember-learn/guides-source/pull/1816 for a couple of versions. This PR backports the fix to the remaining applicable versions.


jenweber (approved):
Thank you!

PR #1822, opened 2022-06-03
Remove some outdated route method references

### LINKED ISSUE #1754: Update "Redirection" guide for 4.0
The guide currently mentions deprecated routing methods `Route#transitionTo`, `Controller#transitionToRoute`, as well as `Route#replaceWith` and `Controller#replaceWithRoute`. The content should be updated according to https://deprecations.emberjs.com/v3.x/#toc_routing-transition-methods.
This might require some learning team discussion to make sure the resulting updated content still feels coesive and instructive.

jenweber:
Closes #1754

We mostly replaced the outdated uses of transitionToRoute/replaceWithRoute, but there were a couple links hanging around. The code samples already use the service, like they are supposed to.

Only 1 reviewer is needed, from any team member.

### DIFF
--- a/guides/release/models/creating-updating-and-deleting-records.md
+++ b/guides/release/models/creating-updating-and-deleting-records.md
@@ -110,28 +110,18 @@ a promise, which makes it easy to asynchronously handle success and failure
 scenarios. Here's a common pattern:
 
 ```javascript
-let post = store.createRecord('post', {
+// Assumed to have already injected the router and store services
+const newPost = this.store.createRecord('post', {
   title: 'Rails is Omakase',
   body: 'Lorem ipsum'
 });
 
-let self = this;
-
-function transitionToPost(post) {
-  this.router.transitionTo('posts.show', post);
+try {
+  await newPost.save();
+  this.router.transitionTo('posts.show', newPost.id);
+} catch (error) {
+  // Handle error
 }
-
-function failure(reason) {
-  // handle the error
-}
-
-post
-  .save()
-  .then(transitionToPost)
-  .catch(failure);
-
-// => POST to '/posts'
-// => transitioning to posts.show route
 ```
 
 ## Deleting Records
--- a/guides/release/models/creating-updating-and-deleting-records.md
+++ b/guides/release/models/creating-updating-and-deleting-records.md
@@ -118,7 +118,7 @@ let post = store.createRecord('post', {
 let self = this;
 
 function transitionToPost(post) {
-  self.transitionToRoute('posts.show', post);
+  this.router.transitionTo('posts.show', post);
 }
 
 function failure(reason) {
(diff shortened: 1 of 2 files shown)


guides/release/models/creating-updating-and-deleting-records.md
```diff
@@ -118,7 +118,7 @@ let post = store.createRecord('post', {
 let self = this;
 
 function transitionToPost(post) {
-  self.transitionToRoute('posts.show', post);
+  this.router.transitionTo('posts.show', post);
```
ijlee2, line 121: The surrounding code initially didn't make sense to me because of a false indentation. I think it'd be nice to use newer JS syntax as well, to make the code more readable.

Could we update the code block to:

```js
// Ember Data assumed
const newPost = this.store.createRecord('post', {
  title: 'Rails is Omakase',
  body: 'Lorem ipsum'
});

try {
  await newPost.save();
  this.router.transitionTo('posts.show', newPost.id);
} catch (error) {
  // Handle the error
}
```

guides/release/routing/redirection.md
```diff
@@ -7,12 +7,10 @@ Usually you want to redirect them to the login page, and after they have success
...
 
-One of the methods is [`transitionTo()`](https://api.emberjs.com/ember/release/classes/Route/methods/transitionTo?anchor=transitionTo).
-Calling `transitionTo()` from a route or
-[`transitionToRoute()`](https://api.emberjs.com/ember/release/classes/Controller/methods/transitionToRoute?anchor=transitionToRoute) from a controller will stop any transitions currently in progress and start a new one, functioning as a redirect.
-`transitionTo()` behaves exactly like the [`LinkTo`](../../templates/links/) helper.
+One of the methods is [`transitionTo()`](https://api.emberjs.com/ember/release/classes/RouterService/methods/transitionTo?anchor=transitionTo).
+Calling `transitionTo()` on the router service will stop any transitions currently in progress and start a new one, functioning as a redirect.
 
-The other one is [`replaceWith()`](https://api.emberjs.com/ember/release/classes/Route/methods/replaceWith?anchor=replaceWith) which works the same way as `transitionTo()`.
+The other one is [`replaceWith()`](https://api.emberjs.com/ember/release/classes/RouterService/methods/replaceWith?anchor=replaceWith) which works the same way as `transitionTo()`.
```
ijlee2, line 13: We may be able to update the surrounding paragraphs so that the sentences flow better. To limit the scope of the work, I'd suggest that we keep the changes as is.

ijlee2 (changes requested):
I think, in the code block, the variable `self` had been introduced because the function `transitionToPost` is not an arrow function. Converting `self.transitionToRoute` to `this.route.transitionTo` would result in a runtime error, I think.

Can we update that entire code block to the code that I suggested?

jenweber:
I didn't even think of that. Thanks for the help! Code sample has been replaced.

ijlee2 (approved):
Thanks! ✨

PR #1824, opened 2022-06-18
Rename `ENABLE_ALL_FEATURES` flag to `ENABLE_OPTIONAL_FEATURES` for versions v2.3.0 and up

bertdeblock:
The `ENABLE_ALL_FEATURES` flag was removed in [v2.3.0](https://github.com/emberjs/ember.js/blob/5a77a4312accd672f29b48b73db2b78745335ae7/CHANGELOG.md#230-january-17-2016) in favour of `ENABLE_OPTIONAL_FEATURES`:
https://github.com/emberjs/ember.js/pull/12627

### DIFF
--- a/guides/release/configuring-ember/feature-flags.md
+++ b/guides/release/configuring-ember/feature-flags.md
@@ -58,5 +58,5 @@ let ENV = {
 };
 ```
 
-For the truly ambitious developer, setting `ENV.EmberENV.ENABLE_ALL_FEATURES` to `true` will enable all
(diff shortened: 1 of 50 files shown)


locks (changes requested):
Since it touches quite old versions, we will have to re-compile them. I think we should do this, but I'm marking as changes requested so someone doesn't accidentally merge this and we forget to precompile the old versions.
@mansona can you provide some guidance on how to do it?

bertdeblock:
@locks Is re-compiling the older versions something we still want to do?

bertdeblock:
Can we merge this and create a separate issue for re-compiling the older versions?

locks:
sure

PR #1827, opened 2022-06-28
Update orientation.md

kuldipem:


bertdeblock:
I don't think this is correct, at least, this doesn't work for me on macOS. I've always used `ctrl + c`.

kuldipem:
Sorry, In a hurry, just created the PR, without thinking it twice.

PR #1829, opened 2022-07-13
Update modifiers related sections to be accurate to what we've been able to do since 3.25

NullVoxPopuli:
This will

- [x] Update existing modifier language in the `release` docs
- [ ] once approved, I'll back port it all the way to 3.25
- [ ] removes mentions of `@ember/render-modifiers` as those modifiers were meant for migration from the old paradigms, and not exactly reflective of how the Octane mental model works.

Prior to 3.25, all modifiers had to be globally available in `app/modifiers/*` -- this is very awkward, and kind of unfortunate from a teaching perspective that we had such a big gap between modifiers being "the thing" and when they can be freely used -- so I'll be ignoring everything prior to 3.25 for these updates.

### DIFF
--- a/guides/release/components/template-lifecycle-dom-and-modifiers.md
+++ b/guides/release/components/template-lifecycle-dom-and-modifiers.md
@@ -243,7 +243,7 @@ On the other hand, if you're looking at JavaScript documentation that tells you 
 
 If you want to set a property, you can use the `prop` element modifier.
 
-## Calling Methods On First Render
+## Calling Methods On Render
 
 So far, we've talked about web APIs that work by setting attributes as well as web APIs that work by setting properties. But what about web APIs that work by calling methods, like setting focus on an element?
 
@@ -253,35 +253,30 @@ For example, let's say we want to focus an `<input>` in a form as soon as the fo
 inputElement.focus();
 ```
 
-This code needs to run after the element is rendered.
-
-The simplest way to accomplish this is by using the `did-insert` modifier from [@ember/render-modifiers][render-modifiers].
-
-[render-modifiers]: https://github.com/emberjs/ember-render-modifiers
+This code needs to run after the element is rendered, and modifiers provided that capability.
 
 ```handlebars {app/components/edit-form.hbs}
 <form>
-  <input {{did-insert this.focus}}>
+  <input {{this.autofocus}}>
 </form>
 ```
 
+Here we define a local modifier right on the component class.
+
 ```js {app/components/edit-form.js}
 import Component from '@glimmer/component';
-import { action } from '@ember/object';
+import { modifier } from 'ember-modifier';
 
 export default class EditFormComponent extends Component {
-  @action
-  focus(element) {
-    element.focus();
-  }
+  autofocus = modifier((element) => element.focus());
 }
 ```
 
-The `did-insert` modifier will call a function after its element is added to the DOM. That function receives the element as a parameter.
+More information about `ember-modifier` can be found [on the README](https://github.com/ember-modifier/ember-modifier) as well as philosophy, how to think about modifiers, etc.
 
-### Abstracting the Logic Into a Custom Modifier
+### Abstracting the Logic Into a Shareable Modifier
 
-Using the `did-insert` modifier works well for one-off cases, but if you want to pull this logic into reusable functionality that you can use throughout your app, you can make your _own_ modifier.
+Using the local modifier sworks well for one-off cases, but if you want to pull this logic into reusable functionality that you can use throughout your app, you can move the modifier to a globally accessible place
 
 The modifier that we're going to build will allow us to say:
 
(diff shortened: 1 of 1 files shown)


guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -253,35 +253,30 @@ For example, let's say we want to focus an `<input>` in a form as soon as the fo
 inputElement.focus();
 ```
 
-This code needs to run after the element is rendered.
-
-The simplest way to accomplish this is by using the `did-insert` modifier from [@ember/render-modifiers][render-modifiers].
-
-[render-modifiers]: https://github.com/emberjs/ember-render-modifiers
+This code needs to run after the element is rendered, and modifiers provided that capability.
```
jenweber, line 256:
```suggestion
We can use a modifier to run a method immediately after an element has rendered.
In this example, we add an `autofocus` method to the component class, and use it in a template:
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -253,35 +253,30 @@ For example, let's say we want to focus an `<input>` in a form as soon as the fo
...
+import { modifier } from 'ember-modifier';
 
 export default class EditFormComponent extends Component {
-  @action
-  focus(element) {
-    element.focus();
-  }
+  autofocus = modifier((element) => element.focus());
 }
 ```
```
jenweber, line 273: I think we should swap the order, putting the modifier declaration first. The reason is that sometimes devs new to Ember copy-paste the first half of examples and expect it to work, then get confused when they see a ton of errors.

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -253,35 +253,30 @@ For example, let's say we want to focus an `<input>` in a form as soon as the fo
...
 
-The `did-insert` modifier will call a function after its element is added to the DOM. That function receives the element as a parameter.
+More information about `ember-modifier` can be found [on the README](https://github.com/ember-modifier/ember-modifier) as well as philosophy, how to think about modifiers, etc.
```
jenweber, line 275:
```suggestion
Modifiers can be used in many other ways! For API documentation and more examples, visit the [the `ember-modifier` README](https://github.com/ember-modifier/ember-modifier).
```

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -253,35 +253,30 @@ For example, let's say we want to focus an `<input>` in a form as soon as the fo
...
 
-### Abstracting the Logic Into a Custom Modifier
+### Abstracting the Logic Into a Shareable Modifier
 
-Using the `did-insert` modifier works well for one-off cases, but if you want to pull this logic into reusable functionality that you can use throughout your app, you can make your _own_ modifier.
+Using the local modifier sworks well for one-off cases, but if you want to pull this logic into reusable functionality that you can use throughout your app, you can move the modifier to a globally accessible place
```
jenweber, line 279:
```suggestion
Writing a modifier in a component class works well for one-time use cases, but if you want to reuse the modifier in other places within your app, you can move the modifier to a globally accessible place.
```

jenweber (commented):
I chatted with nullvoxpopuli, and I can help finish up this PR. I have some suggestions that I will apply directly and then request review.

jenweber:
I made some small wording changes, added a label to the pre-existing input example, and added `eager: false` to address a deprecation warning. I tested this example and I think it's good to go. Would anyone else like to review before we merge? @ember-learn/learning-core-team

NullVoxPopuli:
hmm `eager: false` "feels temporary". :thinking: @chriskrycho -- any feels as to when ember-modifier v4 is happening? <3

guides/release/components/template-lifecycle-dom-and-modifiers.md
```diff
@@ -243,45 +243,51 @@ On the other hand, if you're looking at JavaScript documentation that tells you
...
-The simplest way to accomplish this is by using the `did-insert` modifier from [@ember/render-modifiers][render-modifiers].
-
-[render-modifiers]: https://github.com/emberjs/ember-render-modifiers
-
-```handlebars {app/components/edit-form.hbs}
-<form>
-  <input {{did-insert this.focus}}>
-</form>
+```
+npx ember install ember-modifier
```
NullVoxPopuli, line 260: whoops, this should be ember-cli

this package: https://www.npmjs.com/package/ember
only prints this message: https://github.com/emberjs/ember/blob/master/index.js

and doesn't actually do anything.
Would be great if `ember` were configured as a bin that delegated to `npx ember-cli`

NullVoxPopuli, line 260:
```suggestion
npx ember-cli install ember-modifier
```

NullVoxPopuli:
I just found -- `guides/release/upgrading/current-edition/glimmer-components.md` / https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/ 

I'm gonna address in a separate PR once we resolve this one :partying_face:

guides/release/components/template-lifecycle-dom-and-modifiers.md
jenweber: Oh right, thank you. I will change it to use global ember for now.

NullVoxPopuli:
this is no longer needed, because someone else rewrite this same doc with similar information

PR #1847, opened 2022-09-23
Update block-content.md

Timbuktu1982:
Incorrect structure and component call corrected...

### DIFF
--- a/guides/release/components/block-content.md
+++ b/guides/release/components/block-content.md
@@ -28,7 +28,7 @@ behavior by using HTML, modifiers, and other components within the block.
 To make that more concrete, let's take a look at two similar components
 representing different user's messages.
 
-```handlebars {data-filename="app/components/received-message.hbs"}
+```handlebars {data-filename="app/components/received-message/index.hbs"}
 <aside>
   <div class="avatar is-active" title="Tomster's avatar">T</div>
 </aside>
@@ -45,7 +45,7 @@ representing different user's messages.
 </section>
 ```
 
-```handlebars {data-filename="app/components/sent-message.hbs"}
+```handlebars {data-filename="app/components/sent-message/index.hbs"}
 <aside class="current-user">
   <div class="avatar" title="Zoey's avatar">Z</div>
 </aside>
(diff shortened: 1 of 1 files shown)


bertdeblock:
I don't think anything needed correcting here. The component paths are valid (flat colocation) and the components are just sub components of the message component I assume.

Timbuktu1982:
> I don't think anything needed correcting here? The component paths are valid (flat colocation) and the avatar and username components are just sub components of the message component I assume?

Hello Bert,
yes, if you follow the structure of the previous topics or the example topic, you cannot continue working in the topic at this point.
Follow the structure in the parts before (e.g. here: https://guides.emberjs.com/release/components/conditional-content/)...

bertdeblock:
AFAICT, the guides only make use of flat colocation, I don't see an example that uses nested colocation. Am I overlooking such an example? Also, these docs are not part of the tutorial so I'm not sure what you mean with "you cannot continue working in the topic at this point.". `Message::Avatar` and `Message::Username` or used in later sections as well, so I'm not sure if this is a mistake. But someone from learning can probably confirm.

bertdeblock:
Yeah, I read the previous and next parts, but didn't feel like anything was wrong with those examples. But again, the learning team will confirm. The change in component structure is still wrong I think. All code examples heeft flat colocated file paths.

Timbuktu1982:
https://guides.emberjs.com/release/components/introducing-components/#toc_nesting-components-in-folders

This are not nested and flat colocated file??

bertdeblock:
I can only find one section that refers to nested colocation:

> If you have a component named index.hbs, you can refer to it without the ::Index. So we can refactor app/components/received-message.hbs to app/components/received-message/index.hbs and continue to use it as <ReceivedMessage> without changing all the tags that refer to it:

But all the example file paths that I see use flat colocation (no `index.hbs` filename):

- app/components/received-message/avatar.hbs
- app/components/sent-message/avatar.hbs
- app/components/avatar.hbs
- app/components/received-message/avatar.hbs
- app/components/received-message/username.hbs
- app/components/sent-message/username.hbs
- app/components/my-component.hbs
- app/components/received-message.hbs
- app/components/error-dialog.hbs
- app/components/blog-post.hbs
- app/components/some-component.hbs
- app/components/counter.hbs
- app/components/double-it.hbs

bertdeblock:
Yes, I'm following the confusion about the avatar and username components, but I wasn't talking about that anymore. You are changing from flat colocation to nested colocation here:
https://github.com/ember-learn/guides-source/pull/1847/files#diff-643982adf4ba091b2816a675e207c79ff9a54f289778e48af4e924b9edc1bc95L31-R31.
I don't think this is needed, because alle example file paths in the docs use flat colocation.

Timbuktu1982:
Ah ok. Now i understand you... we discuss about different parts of my PR... sorry for the misstake

bertdeblock:
No worries, happens! 😄

PR #1848, opened 2022-09-23
Update index.md

Timbuktu1982:
remove "Message" because not needed (see example picture)

### DIFF
--- a/guides/release/components/index.md
+++ b/guides/release/components/index.md
@@ -64,7 +64,6 @@ instance, you can copy the following example HTML for a simple messaging app:
   </section>
 
   <form>
-    <label for="message">Message</label>
     <input id="message" />
     <button type="submit">
       Send


MelSumner (changes requested):
The label is required and should not be removed. While an accessible name for the input is required for users with assistive technology (see https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html), all users benefit from having a clear label letting them know what information the input field requires.

MinThaMie:
@MelSumner Should we update the image? Because that does not show the label.
Screenshot 2023-02-07 at 14 39 23

MelSumner:
@MinThaMie yes, definitely.

MinThaMie:
I've created a new issue #1898 that you or I could pick up, but to make sure we're not merging this I'll close this PR.

PR #1849, opened 2022-09-23
Update introducing-components.md

Timbuktu1982:
remove "Message" because not needed (see example picture)


locks:
I don't know what image you are referring to. The input should definitely have a label though, so this PR seems incorrect to me.

PR #1850, opened 2022-09-26
Update orientation.md with Ember v4.7.0

tniezurawski:
I'm not sure how much effort it would be to automate updating this version in guides to the latest version but here's a manual update.

I'm also updating the node version to the LTS so it won't look "so old".

### DIFF
--- a/guides/release/tutorial/part-1/orientation.md
+++ b/guides/release/tutorial/part-1/orientation.md
@@ -24,8 +24,8 @@ To verify that your installation was successful, run:
 
 ```shell
 $ ember --version
-ember-cli: 4.5.0
-node: 14.19.1
+ember-cli: 4.7.0
+node: 16.13.1
 os: linux x64
 ```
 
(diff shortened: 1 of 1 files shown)


MinThaMie:
Thanks for the heads-up. This part is automatically generated as you can see in #1832. The old node version is the one we're using in https://github.com/ember-learn/super-rentals-tutorial. I'll try to merge #1832 this week when I've removed all the images from the PR

tniezurawski:
> This part is automatically generated as you can see in https://github.com/ember-learn/guides-source/pull/1832

Ah, that makes sense @MinThaMie. I somehow wanted to believe this is automated or semi-automated 😅 

> The old node version is the one we're using in https://github.com/ember-learn/super-rentals-tutorial

Oki, it makes sense then to keep the node version the same as in the tutorial app. Maybe I'll contribute there then.

Closing the PR :v: Thanks for the clarification!

PR #1851, opened 2022-09-29
Update linking-between-routes.md

2hu12:
Remove the legacy `@id` argument note of `<LinkTo />` component according to https://deprecations.emberjs.com/v3.x/#toc_ember-built-in-components-legacy-arguments, using the `id` HTML attribute should be fine now.


MinThaMie (approved):
Thanks, this looks correct :)

PR #1854, opened 2022-10-18
Added storing MabBox token with dotenv.

jsantos42:
Added a Zoey box suggesting an alternative and safer way of storing the MapBox token, using dotenv.


MinThaMie:
Will close this in favour of https://github.com/ember-learn/super-rentals-tutorial/pull/216 because that is where this change should happen. Thanks for the contribution and please check the feedback on the PR.

PR #1861, opened 2022-11-09
Remove `ember-gestures` addon recommendation

davidtaylorhq:
`ember-gestures` and its dependency, `ember-hammertime`, are no longer maintained and do not function at all on recent versions of Ember.

Refs:
- https://github.com/html-next/ember-gestures/issues/139
- https://github.com/html-next/ember-hammertime/issues/103


jenweber (approved):
Thanks for this update!

PR #1864, opened 2022-11-10
Upgrade guidemaker-ember-template

wagenet:
Previous versions ended up needing Python2 during install, which I couldn't install on my M1 Mac.


MinThaMie:
Closed because we updated this with the launch of the redesign :)

PR #1865, opened 2022-11-10
Updating Fragment Syntax from handlebars

BootsStribling:
Seems that the guide is still calling for handlebars syntax but the completed app in the guide has HTML fragment syntax. I just updated the syntax to match the completed app in the guide.


kategengler:
First, thanks for attempting to make the docs more consistent. 

In this case, though, this change is made to the version of the guides for Ember v2.14.0. At that time, angle bracket invocation of components nor did component versions of `link-to` (it was previously a helper).

So, this syntax would not work in v2.14.0 of Ember. If you find an instance of this in guides newer than v3.13 (Octane), please submit another PR.

Issue #1881, opened 2022-12-05
WIP Document new and improved helper functions experience


jenweber:
Excellent news - helpers are now much more intuitive to use in Ember! The feature shipped quietly, and now that it's had some additional testing, we should document it and invite all our developers to start using it!

```
{{this.someFunctionOnComponentClass 'hello'}}
```

SkoebaSteve:
As discussed I'll take this and will create a PR by next week, thanks for the introduction @jenweber!

PR #1888, opened 2022-12-30
restructure global and local helpers

SkoebaSteve:
This should address [this issue](https://github.com/ember-learn/guides-source/issues/1881) and document the usage of plain helper functions next to global helper functions.

### DIFF
--- a/guides/v4.5.0/components/helper-functions.md
+++ b/guides/v4.5.0/components/helper-functions.md
@@ -1,8 +1,8 @@
 Helper functions are JavaScript functions that you can call from your template.
 
-Ember's template syntax limits what you can express to keep the structure of your application clear at a glance. When you need to compute something using JavaScript, you can use helper functions. It's possible to create your own helpers or just [use the built-in ones](./#toc_built-in-helpers).
+Ember's template syntax limits what you can express to keep the structure of your application clear at a glance. When you need to compute something using JavaScript, you can use helper functions. It's possible to create your own helpers, locally or just [use the built-in ones](./#toc_built-in-helpers).
 
-For instance, let's take a look at a generic message component from a messaging app.
+Let's take a look at a generic message component from a messaging app.
 
 ```handlebars {data-filename="app/components/message.hbs"}
 <Message::Avatar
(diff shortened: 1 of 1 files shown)


guides/v4.5.0/components/helper-functions.md
```diff
@@ -62,11 +62,43 @@ Since the title is just the `@username` plus some extra stuff, we can replace `@
...
-## Writing a Helper Function
+In this case we want a helper function that takes three arguments: a string, a starting position, and a length. The function will return a substring of the original string.
 
-We define helper functions in the `app/helpers` folder.
+## Local Helper Functions
 
-In this case we want a helper function that takes three arguments: a string, a starting position, and a length. The function will return a substring of the original string.
+It's possible to use plain functions for helpers and modifiers.
+
+```js {data-filename="app/components/message.js"}
```
SkoebaSteve, line 71: I am personally not the biggest fan of using the same example already used for global helper functions but I left it there to discuss. The `substring` helper example that was there initially feels a little bit abstract, especially for local helpers.

Happy to have some input on this before continuing :)

guides/v4.5.0/components/helper-functions.md
```diff
@@ -118,7 +148,7 @@ We can then use this helper in the component's template to get the first letter
 </section>
 ```
 
-### Named Arguments
+## Named Arguments
```
SkoebaSteve, line 151: To my understanding, they should work roughly similarly with local helpers:

```handlebars
{{this.calculate 1 2 op="add"}}
```

Would it be better to keep this as a separate section and insert here the difference between the using it as a local vs global helper? Or do we move all this into the separate sections created for 'Local Helper Functions' and 'Global Helper Functions'?

guides/v4.5.0/components/helper-functions.md
jenweber: I agree that substring feels abstract for local helpers. If you can think of an example that would be similar to a real world use and understandable by junior developers, feel free to replace it. It is also ok to keep using this same example for now, in the interest of keeping things moving with the rest of the refactor. All the real world examples I could think of were complex - filtering over data, displaying dates, and too much math.

guides/v4.5.0/components/helper-functions.md
jenweber: I asked myself, which option is the most clear to beginners? And I think it’s to keep the explanation separate for each kind of helper. The “global helpers” section has the potential to be really short, where we can build on previous learning in this way: “Just like local helpers, you can use positional arguments.”

guides/v4.5.0/components/helper-functions.md
SkoebaSteve: Sounds good, let me focus on the refactor first before changing the example

jenweber:
This looks great to me! I like how you structured this, and I didn’t spot any typos. Is there anything else you want to add to this PR? If not, please promote it from draft PR to ready for review. I am running the tests in CI. If anything fails and it’s confusing, I’ll give you some pointers, so don’t spend too much time debugging. Thanks again for your help!

jenweber (approved):

PR #1890, opened 2023-01-09
Add helper function changes to other guides versions

jenweber:
@SkoebaSteve updated the helpers content for v4.5 to cover how to use local helpers in https://github.com/ember-learn/guides-source/pull/1888. This PR copies those changes into more recent guides versions as well. The prose/changes were previously reviewed - this is copy-paste.

@SkoebaSteve if there's ever any other changes you want to make to the guides that do not need to go on older versions, you can make your changes in the `release` folder. I think the helper functions made sense to include on older versions as well, though!


SkoebaSteve (approved):
Awesome, thank @jenweber for explaining this :)

MinThaMie (approved):
Lekker gewerkt @SkoebaSteve  :) And thanks @jenweber for fixing it across versions

PR #1899, opened 2023-02-07
Update screenshot to match HTML and make CSS work

### LINKED ISSUE #1898: Update the messaging app screenshot
`/images/ember-core-concepts/messaging-app-1.png` is not showing the label that should be there according to accessibility. #1848 suggested to remove the label, but discussed with @MelSumner this should not be de solution.

Screenshot 2023-02-07 at 14 51 55

Screenshot 2023-02-07 at 14 52 27

MinThaMie:
Fixes #1898

Somebody noticed that the HTML did not match the screenshot [here](https://guides.emberjs.com/release/components/) so I've updated the screenshot + made sure the CSS worked (because that was broken too. Now the label shows up as per a11y standards.


MelSumner (approved):
Thank you for doing this work! Looks good! 👍

PR #1904, opened 2023-03-03
Remove use of @ember/string

jenweber:
Replace it with plain old regex `replace`.

`replaceAll` would be easier to read, but may not be available in people's projects or node environments.

Similarly I decided not to escape the _ or -, since it's all by itself, but I could add an escape if preferred.

This addresses a blocker for https://github.com/emberjs/rfcs/pull/898


locks:
From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility it seems most platforms have it since 2020 (Samsung Internet since 2021), so I think it's safe to use it!

locks (approved):

wagenet (approved):

PR #1907, opened 2023-03-07
[RFC-774] Remove documentation for implicit route model

wagenet:
[RFC-774](https://github.com/emberjs/rfcs/pull/774)

I believe it is safe to merge this now even before the deprecation has been merged, but I'm open to alternative ways to handle this.

https://github.com/emberjs/ember.js/pull/20376

### DIFF
--- a/guides/release/routing/specifying-a-routes-model.md
+++ b/guides/release/routing/specifying-a-routes-model.md
@@ -206,7 +206,7 @@ export default class PostRoute extends Route {
 
 Note that currently, if `model` is not specified, Ember will attempt
 to automatically find a store and use it for lookup. This behavior
-is unreliable and will be removed in future Ember versions.
+is a common source of confusion and will be removed in future Ember versions.
 
 ### Linking to a dynamic segment
 
--- a/guides/release/routing/specifying-a-routes-model.md
+++ b/guides/release/routing/specifying-a-routes-model.md
@@ -184,19 +184,29 @@ export default class PostRoute extends Route {
 }
 ```
 
-If you do not define a model hook for a route, it will default to using EmberData to look up the record, as shown below:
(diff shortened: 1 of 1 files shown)


jenweber:
Thank you! I like the prose you added. However, the soon to be deprecated behavior is such a big gotcha, so I think it may still merit a small, concise warning about the behavior when someone does nothing with the model hook. I have a hypothesis that this behavior is one of Ember’s main stumbling blocks for beginners. For best developer experience, I think we would want to keep the warning until support is dropped in a major. What do you think about that?

Someone would see a deprecation warning to help guide them too, but my experience is that those accumulate, and devs may ignore them due to alert fatigue.

Regardless, I don’t think we need to wait for the deprecation to land in order to ship this. The guides should represent the general way we want to encourage people to write ember apps, and we don’t need to cover all approaches in the guides.

guides/release/routing/specifying-a-routes-model.md
```diff
@@ -184,19 +184,29 @@ export default class PostRoute extends Route {
...
 ```
 
-In the `model` hook for routes with dynamic segments, it's your job to
-turn the ID (something like `47` or `post-slug`) into a model that can
-be rendered by the route's template. In the above example, we use the
-post's ID (`params.post_id`) as an argument to EmberData's `findRecord`
-method.
+Note that currently, if `model` is not specified, Ember will attempt
+to automatically find a store and use it for lookup. This behavior
+is unreliable and will be removed in future Ember versions.
```
jenweber, line 209:
```suggestion
is a common source of confusion and will be removed in future Ember versions.
```

jenweber (approved):
Left you a small wording change. We can emphasize our product priorities of nicer dev experience.

PR #1915, opened 2023-04-24
Use avatarInitial instead of avatarTitle in diff

talltorp:
The diff mentioned that the old initial was the @avatarTitle which confused me for a while.

I changed it to @avatarInitial which I assume is what's intended


MinThaMie (approved):
Thanks :) you are correct!

Issue #1921, opened 2023-05-26
Remove deprecated EmberData references for 5.0


jenweber:
- [ ] `artists.firstObject;` - replace with awaiting the array
- [ ] Can we still use `.get('firstObject')`? I don't think so
- [ ] pushObject. What to replace this with? See code sample

PushObject code sample

```js
let blogPost = this.store.peekRecord('blog-post', 1);
let comment = this.store.peekRecord('comment', 1);
let comments = await blogPost.comments;
comments.pushObject(comment);
blogPost.save();
```

PR #1928, opened 2023-06-17
Remove use of pushObject on RecordArray and ManyArray

jaredgalanis:
partially addresses: https://github.com/ember-learn/guides-source/issues/1921


jenweber (approved):

mansona:
Should I redo #1927 to make sure this gets into 4.12?

jenweber:
Good thinking! Yes, we should change this in 4.12 ideally. But low priority compared to other release tasks, if there’s anything else you are working on.

PR #1930, opened 2023-07-04
Fix broken handlebars highlighting

MinThaMie:


NullVoxPopuli (changes requested):
What highligher do we use?

As far as i know, 'hbs' is supported by prism and highlight.js.
Could be a configuration issue?

In any case, for highest fidelity highlighting, and compatibility we'd probably want to use the renpective glimmer highlighting plugin, as we don't use handlebars, and use features it doesn't have.

Brb getting plugin links

NullVoxPopuli:
Except you're mr empress :p

NullVoxPopuli:
Anywho, i posted here: https://github.com/empress/ember-showdown-prism/issues/50

PR #1931, opened 2023-07-04
update pre-built guides

mansona:
after #1717 we actually should have update the pre-built guides 🙈 

I used https://github.com/ember-learn/guides-source/pull/1430 to do this PR btw. In the description from that PR I said: 

> I'm pretty sure there is a blocker that will prevent us from shipping this 🤔

but I have no idea what that is 😂 😭  if it's working now I guess we can just ship it? 🤷 

Edit: i figured out the blocker! I was doing some experimentation on my local copy of the guides and there was no attempt to clean your local version before running the build 🙈 I have added that to the script


MinThaMie (approved):
My click-through looks good! just a bit weirded out by the `less` and `scss` files being added

PR #1932, opened 2023-07-04
Update prebuilt versions

mansona:
This is just some small prep before running the script again to add more prebuilt versions. I'm separating the PRs to make these real changes easier to review

Note: this knocks almost 3 minutes off our netlify build times and gets us just under the 5 mins mark 😍 

I ran this against the repo and opened #1933 so I've you're happy with this PR then you should merge that one too 👍

PR #1933, opened 2023-07-04
Add more versions to prebuilt

mansona:
This is running the `npm run build:prebuilt` that was updated in #1932 so I've you're happy with this then you can merge that script update too 👍


MinThaMie (approved):
Awesome :)

PR #1934, opened 2023-07-05
Fix `<LinkTo />` in title

MinThaMie:
@locks brought the issue below to our attention and this fixes it. I've also checked is this happend in other places but that was not the case, except for one weird `<a>` tag in v1.10, so I removed that one too :)

Screenshot 2023-07-05 at 09 48 13

**After**
Screenshot 2023-07-05 at 09 55 10


mansona (approved):
Looks good 👍

Issue #1937, opened 2023-07-06
Differences not shown in manual code


karavaev-evgeniy:
https://guides.emberjs.com/release/tutorial/part-1/interactive-components/. 

In this code example, the user should understand that the constructor part needs to be removed.

This problem occurs in all code snippets of the documentation.

<img width="834" alt="Screenshot 2023-07-06 at 21 20 23" src="https://github.com/ember-learn/guides-source/assets/6982951/3f559a5d-b8bd-469b-95ad-a67c899f1af0">

mdtro:
Came here to report this same issue. I'm seeing the same thing in Firefox, Chrome, and Safari.
![image](https://github.com/ember-learn/guides-source/assets/20070360/1cb7689c-9af2-4dc0-81d3-554e01b0d653)

PR #1941, opened 2023-07-11
update guidemaker-ember-template

mansona:
This should give us back diffs 👍 🎉 

fixes https://github.com/ember-learn/guides-source/issues/1937


locks (approved):
thank yyooouuu

PR #1946, opened 2023-07-19
Update the guides for element modifiers

mixonic:
At https://github.com/emberjs/rfcs/pull/934 we're tracking an effort to implement the RFC https://github.com/emberjs/rfcs/pull/811. In that RFC (at
https://github.com/emberjs/rfcs/blob/master/text/0811-element-modifiers.md#how-we-teach-this) there are two things which are suggested for documentation in the guides:

* Update references that mention `ember-modifier` must be installed, since it ships in the default blueprint (confirmed at https://github.com/ember-cli/ember-new-output/blob/master/package.json#L50)
* Introduce the class-based modifier API. In this PR I've linked to the upstream documentation at https://github.com/ember-modifier/ember-modifier#usage to execute on that suggestion. I don't believe we need to laborously explain the class-based APIs to guides readers, it just distracts from some otherwise straight-forward use cases and examples.

### DIFF
--- a/guides/release/upgrading/current-edition/glimmer-components.md
+++ b/guides/release/upgrading/current-edition/glimmer-components.md
@@ -784,15 +784,17 @@ functionality that lifecycle hooks contained.
 
 #### Writing your own modifiers
 
-There are also community APIs available for writing your own modifiers, such as
-[ember-modifier](https://github.com/ember-modifier/ember-modifier).
-Ember itself has low level APIs known as _modifier managers_ which can be used
-to write these higher level APIs. In general, it's recommended to use a
-community addon to write modifiers, and _not_ to write your own modifier
-manager.
+New Ember apps ship with a dependency on
+[ember-modifier](https://github.com/ember-modifier/ember-modifier), which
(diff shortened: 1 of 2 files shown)


MinThaMie (approved):
Looks good :)

guides/release/upgrading/current-edition/glimmer-components.md
```diff
@@ -825,12 +827,16 @@ export default class ScrollComponent extends Component {
...
 
-This modifier generalizes the functionality that the component implemented using
-lifecycle hooks before, so we can use this modifier whenever we need to in _any_
-component. This is a much better solution than manually managing event listeners
-every time we need one! At this point, the modifier is effectively the same as
-the `{{on}}` modifier as well, so we could get rid of it altogether and replace
-it with `on`:
+The new `add-event-listener` modifier presents a more expressive interface to
+the `hbs` template: There is only a single modifier to apply instead of two, the
+implemention always tears down after itself upon teardown of the target element,
```
mixonic, line 832:
```suggestion
implementation always tears down after itself upon teardown of the target element,
```

PR #1894, opened 2023-01-19
Update block-content.md

arjenblokzijl:
Hi,

A co-worker of mine was looking for some examples regarding named blocks. Since it only had one (minor - default) example I've gone ahead and added another one. Feedback appreciated! 

🙏

PR #1958, opened 2023-08-24
Update block-content.md [Named blocks]

Shishouille:
Aaah, I just saw #1894, I hope it's still ok! Some people requested more documentation about named blocks, so here's another proposition 😃

### DIFF
--- a/guides/release/components/block-content.md
+++ b/guides/release/components/block-content.md
@@ -345,9 +345,9 @@ You could also want to pass some values. This is the same process as the default
   </div>
 ```
 
-If we hadn’t named blocks, we would certainly have to pass components as `args` to the popover. But this is much more practical!
+Without named blocks, we would certainly have to pass components as `args` to the popover. But this is much more practical!
 
-Here’s how we would call it:
+Here’s how we would call our named blocks as a consumer:
 
 ```handlebars
 <Popover>
@@ -355,7 +355,7 @@ Here’s how we would call it:
     <button type="button">Click to {{if open "close" "open"}}  the popover!</button>
   </:trigger>
   <:content>
-      This is what is showed when I'm opened!
+      This is what is shown when I'm opened!
   </:content>
 </Popover>
 ```
@@ -412,7 +412,7 @@ The title being optional when you create a card, you can use the `(has-block)` h
 </Card>
 ```
 
-As you are not using named blocks, you can simply yield the content you would like to add.
+As you are not using named blocks, you can simply yield the content you would like to add, which becomes the default yield block.
 
 
 
--- a/guides/release/components/block-content.md
+++ b/guides/release/components/block-content.md
@@ -322,4 +322,96 @@ We can yield back multiple values as well, separated by spaces.
 </BlogPost>
 ```
 
+## Named Blocks
+
+If you want to yield content to different spots in the same component, you can use named blocks. You just need to specify a name for the yielded block, like this:
+
+```handlebars
+{{yield to="somePlace"}}
+```
+
+You could also want to pass some values. This is the same process as the default `yield`, but you just have to pass `to` as the last argument. An example would be the popover:
+
+```handlebars {data-filename=app/components/popover.hbs}
+<div class="popover">
+  <div class="popover__trigger">
+    {{yield this.isOpen to="trigger"}}
+  </div>
+  <div class="popover__content">
+    {{yield this.isOpen to="content"}}
+  </div>
+</div>
+```
+
+If we hadn’t named blocks, we would certainly have to pass components as `args` to the popover. But this is much more practical!
+
+Here’s how we would call it:
+
+```handlebars
+<Popover>
+  <:trigger as |open|>
+    <button>Click to {{if open "close" "open"}}  the popover!</button>
+  </:trigger>
+  <:content>
+      This is what is showed when I'm opened!
+  </:content>
+</Popover>
+```
+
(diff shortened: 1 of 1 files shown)


guides/release/components/block-content.md
```diff
@@ -322,4 +322,98 @@ We can yield back multiple values as well, separated by spaces.
...
+    </div>
+    {{#if this.isOpen}}
+      <div class="popover__content">
+        {{yield to="content"}}
+      </div>
+    {{/if}}
+  </div>
+```
+
+If we hadn’t named blocks, we would certainly have to pass components as `args` to the popover. But this is much more practical!
```
elwayman02, line 348:
```suggestion
Without named blocks, we would certainly have to pass components as `args` to the popover. But this is much more practical!
```

guides/release/components/block-content.md
```diff
@@ -322,4 +322,98 @@ We can yield back multiple values as well, separated by spaces.
...
+
+Here’s how we would call it:
```
elwayman02, line 350:
```suggestion
Here’s how we would call our named blocks as a consumer:
```

guides/release/components/block-content.md
```diff
@@ -322,4 +322,98 @@ We can yield back multiple values as well, separated by spaces.
...
+
+```handlebars
+<Popover>
+  <:trigger as |open|>
+    <button type="button">Click to {{if open "close" "open"}}  the popover!</button>
+  </:trigger>
+  <:content>
+      This is what is showed when I'm opened!
```
elwayman02, line 358:
```suggestion
      This is what is shown when I'm opened!
```

guides/release/components/block-content.md
```diff
@@ -322,4 +322,98 @@ We can yield back multiple values as well, separated by spaces.
...
+
+The title being optional when you create a card, you can use the `(has-block)` helper with the named block by adding its name as a first parameter. That means you could also create this card:
+
+```handlebars
+<Card>
+  I don't want any title, and I only have a default content!
+</Card>
+```
+
+As you are not using named blocks, you can simply yield the content you would like to add.
```
elwayman02, line 415:
```suggestion
As you are not using named blocks, you can simply yield the content you would like to add, which becomes the default yield block.
```

guides/release/components/block-content.md
```diff
@@ -322,4 +322,98 @@ We can yield back multiple values as well, separated by spaces.
...
+  <div class="popover">
+    <div class="popover__trigger">
+      {{yield this.isOpen to="trigger"}}
+    </div>
+    {{#if this.isOpen}}
+      <div class="popover__content">
+        {{yield to="content"}}
+      </div>
+    {{/if}}
+  </div>
```
IgnaceMaes, line 345: The initial indent can be removed:

```suggestion
<div class="popover">
  <div class="popover__trigger">
    {{yield this.isOpen to="trigger"}}
  </div>
  {{#if this.isOpen}}
    <div class="popover__content">
      {{yield to="content"}}
    </div>
  {{/if}}
</div>
```

guides/release/components/block-content.md
```diff
@@ -322,4 +322,98 @@ We can yield back multiple values as well, separated by spaces.
...
+That would give this result:
+
+```html
+<!-- rendered -->
+<div class="popover">
+  <div class="popover__trigger">
+    <button type="button">Click to close the popover!</button>
+  </div>
+  <div class="popover__content">
+     This is what is showed when I'm opened!
```
IgnaceMaes, line 374: Two spaced indent for consistency:

```suggestion
    This is what is showed when I'm opened!
```

guides/release/components/block-content.md
```diff
@@ -322,4 +322,98 @@ We can yield back multiple values as well, separated by spaces.
...
+
+Here’s how we would call our named blocks as a consumer:
+
+```handlebars
+<Popover>
+  <:trigger as |open|>
+    <button type="button">Click to {{if open "close" "open"}}  the popover!</button>
+  </:trigger>
+  <:content>
+      This is what is shown when I'm opened!
```
IgnaceMaes, line 358: And here too:

```suggestion
     This is what is shown when I'm opened!
```

guides/release/components/block-content.md
```diff
@@ -322,4 +322,98 @@ We can yield back multiple values as well, separated by spaces.
...
+  </:content>
+</Popover>
+```
+
+We know the state of the popover because we passed it as an argument to the `yield`. To access its value, use the block parameters at the named block scope. It will not be accessible at the `Popover` level, so if you want the value to be available for all the blocks, you will have to pass it for each of them.
+
+That would give this result:
```
IgnaceMaes, line 365: The "That" reference confused me a bit with the explanation above. Making the reference exact would prevent this for me.

```suggestion
Rendering the previous code example would give this as result:
```

IgnaceMaes (commented):
This is GREAT!

I reviewed the other linked PR first but my suggestions there would basically mean writing something like this. The code examples make a lot of sense.

I've put a couple of suggestions. If those are resolved this is good to merge.

Sorry you have to wait so long for a review, that's not ideal.

IgnaceMaes (approved):
As the suggestions were mostly stylistic, I've taken the liberty to just apply them so we can get this through.

Thank you for this great contribution! 🎉

Shishouille:
No worries! And you're welcome 😃 I'm glad it got merged 👍

PR #1960, opened 2023-08-27
Add TypeScript guides

gitKrystan:
image

While we have [grand plans](https://github.com/orgs/emberjs/projects/9/views/1?pane=issue&itemId=37988245) to eventually have a js/ts toggle for all of the guides, we need _something_ to point to in the meantime. In this PR, I am adding initial TypeScript guides pulled from the various resources currently scattered across the interwebs to use as a starting point.

These pages currently will _not_ show up in the table of contents. You'll have to visit the link directly. The purpose of this is to allow us to merge this mega-PR and continue to iterate without being fully released. In the meantime, we can link people directly to these pages with the caveat that the pages are still under construction.

Sources:
* https://docs.ember-cli-typescript.com/
* https://blog.skylight.io/ts-extends-confidence-2-2022/

This will help us close out https://github.com/emberjs/rfcs/pull/952

## To review:
Visit deploy preview linked below and navigate to `./release/using-typescript/`
(e.g. https://deploy-preview-1960--ember-guides.netlify.app/release/using-typescript/ for the deploy that was current at the time of this writing)

Review for:
* Typos
* Mistakes
* Completeness

Please make liberal use of the GitHub suggestions feature because I am starting to get bleary-eyed looking at this for so long. 🙏


guides/release/using-typescript/configuration.md
```diff
@@ -0,0 +1,43 @@
...
+return require('@embroider/compat').compatBuild(app, Webpack, {
+  packagerOptions: {
+    webpackConfig: {
+      devtool: 'source-map'
+    }
+  }
+}
+```
+
+If you're updating from an older version of the addon, you may also need to update your `tsconfig.json`. (Current versions generate the correct config at installation.) Either run `ember generate ember-cli-typescript` or verify you have the same sourcemap settings in your `tscsonfig.json` that appear in [the blueprint](https://github.com/typed-ember/ember-cli-typescript/blob/master/blueprint-files/ember-cli-typescript/files/tsconfig.json).
```
chriskrycho, line 43:
```suggestion
If you're migrating from `ember-cli-typescript`, particularly an older version, to Ember's out-of-the-box TypeScript support, you may also need to update your `tsconfig.json`. Current versions of `ember-cli-typescript` generate the correct config at installation. You do *not* need to install `ember-cli-typescript` for new apps or adding.
```

guides/release/using-typescript/getting-started.md
```diff
@@ -0,0 +1,41 @@
...
+
+- `typescript`
+- `@types/ember`
+- `@types/ember-data`
+- `@types/ember__*` – `@types/ember__object` for `@ember/object` etc.
+- `@types/ember-data__*` – `@types/ember-data__model` for `@ember-data/model` etc.
+- `@types/qunit`
+- `@types/rsvp`
+
+<!-- FIXME: Zoey says note about preview types and how you can probably use them if you're not using EmberData -->
```
chriskrycho, line 27: This makes me extremely sad. 😂

guides/release/using-typescript/getting-started.md
```diff
@@ -0,0 +1,41 @@
...
+ember new my-typescript-app --typescript
+```
+
+All dependencies will be added to your `package.json`, and you're ready to roll!
+
+Using the `--typescript` flag changes the output of `ember new` in a few ways:
+
+- project files will be generated with `.ts` extensions instead of `.js`
+- packages will be installed and files will be generated to make TypeScript work in your project
+- TypeScript-specific config is set
```
chriskrycho, line 13:
```suggestion
- We configure Babel to transpile TypeScript files.
- We configure TypeScript to type check your project.
```

guides/release/using-typescript/getting-started.md
chriskrycho, line 27:
```suggestion

<div class="cta">
  <div class="cta-note">
    <div class="cta-note-body">
      <div class="cta-note-heading">Zoey says...</div>
      <div class="cta-note-message">
        Ember also includes its own types compiled directly from its source code, as described <a href="https://blog.emberjs.com/stable-typescript-types-in-ember-5-1/">in this blog post</a>. We still use the <code>@types</code> packages by default for the sake of compatibility with Ember Data, because Ember Data is not yet compatible with Ember’s native official types. However, if you only do not use Ember Data, we recommend following the instructions in that blog post to switch to the native types, which are guaranteed to always be 100% correct and 100% up to date!
      </div>
    </div>
    <img src="/images/mascots/zoey.png" role="presentation" alt="">
  </div>
</div>
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
+<!-- FIXME: This is copy-pasta from ember-cli-typescript docs and needs updates -->
+
+# TypeScript and Ember
+
```
chriskrycho, line 4:
```suggestion

**Note:** 🚧 This section is under construction! 🏗️ The content here may not be fully up to date!

```

guides/release/using-typescript/index.md
```diff
@@ -0,0 +1,41 @@
+This guide is designed to help you get up and running with TypeScript in an Ember app.
```
chriskrycho, line 1:
```suggestion
**Note:** 🚧 This section is under construction! 🏗️ The content here may not be fully up to date!

This guide is designed to help you get up and running with TypeScript in an Ember app.

```

guides/release/using-typescript/getting-started.md
```diff
@@ -0,0 +1,41 @@
+To start a new Ember project with TypeScript, you can add the `--typescript` flag when you run [`ember new`](../../getting-started/quick-start):
```
chriskrycho, line 1:
```suggestion
**Note:** 🚧 This section is under construction! 🏗️ The content here may not be fully up to date!

To start a new Ember project with TypeScript, you can add the `--typescript` flag when you run [`ember new`](../../getting-started/quick-start):
```

guides/release/using-typescript/configuration.md
```diff
@@ -0,0 +1,43 @@
+<!-- FIXME: This is copy-pasta from ember-cli-typescript docs and needs updates -->
```
chriskrycho, line 1:
```suggestion
**Note:** 🚧 This section is under construction! 🏗️ The content here may not be fully up to date!

```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
+<!-- FIXME: This is copy-pasta from ember-cli-typescript docs and needs updates -->
```
chriskrycho, line 1:
```suggestion
**Note:** 🚧 This section is under construction! 🏗️ The content here may not be fully up to date!
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+
+- Go more slowly, but write down at least minimally accurate types as you go. (This is easier if you follow the leaves-first strategy recommended above.) This is much slower going, and can feel harder because you can’t just skip over things. Once you complete the work for any given module, though, you can be confident that everything is solid and you won’t have to revisit it in the future.
+
+There is an inherent tradeoff between these two approaches; which works best will depend on your team and your app.
+
+## Install other types
+
+You'll want to use other type definitions as much as possible. The first thing you should do, for example, is install the types for your testing framework of choice: `@types/ember-mocha` or `@types/ember-qunit`. Beyond that, look for types from other addons: it will mean writing `any` a lot less and getting a lot more help both from your editor and from the compiler.
```
chriskrycho, line 70:
```suggestion
You'll want to use other type definitions as much as possible. Many packages ship their own type definitions, and many others have community-maintained definitions from [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), available in the `@types` name space. The first thing you should do is to look for types from other addons: it will mean writing `any` a lot less and getting a lot more help both from your editor and from the compiler.
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+
+_Where can I find types?_ Some addons will ship them with their packages, and work out of the box. For others, you can search for them on [Definitely Typed](https://github.com/DefinitelyTyped/DefinitelyTyped), or on npm under the `@types` namespace. (In the future we hope to maintain a list of known types; keep your eyes open!)
```
chriskrycho, line 72:
```suggestion
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+
```
chriskrycho, line 73:
```suggestion
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+We install this file because the actual `config/environment.js` is (a) not actually identical with the types as you inherit them in the content of an application, but rather a superset of what an application has access to, and (b) not in a the same location as the path at which you look it up. The actual `config/environment.js` file executes in Node during the build, and Ember CLI writes its result as `<my-app>/config/environment` into your build for consumption at runtime.
+
+## String-keyed lookups
+
+Ember makes heavy use of string-based APIs to allow for a high degree of dynamicism. With some limitations, you can nonetheless use TypeScript very effectively to get auto-complete/IntelliSense as well as to accurately type-check your applications.
+
+A few of the most common speed-bumps are listed here to help make this easier:
+
+### Nested keys in `get` or `set`
```
chriskrycho, line 102:
```suggestion
### Classic `get` or `set` methods
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+
+In general, `this.get` and `this.set` will work as you'd expect _if_ you're doing lookups only a single layer deep. Things like `this.get('a.b.c')` don't (and can't ever!) type-check; see the blog posts for a more detailed discussion of why.
```
chriskrycho, line 104:
```suggestion
In general, the `this.get` and `this.set` methods on `EmberObject` subclasses and the standalone `get` and `set` functions will work as you'd expect _if_ you're doing lookups only a single layer deep. We do not provide support for deep key lookups like `get(someObj, 'a.b.c')`, because normal property access can works correctly across the whole Ember ecosystem since at least Ember and Ember Data 3.28.
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+   ```
+
+2. **Using `// @ts-ignore`.** This will _not do any type-checking_, but is useful for the cases where you are intentionally checking a path which may be `null` or `undefined` anywhere long it.
+
+   ```typescript
+   // @ts-ignore
+   get(someObject, 'deeply.nested.key');
+   ```
+
+   It's usually best to include an explanation of _why_ you're ignoring a lookup!
```
chriskrycho, line 124: Since regular property access “just works”, and has for a very long time, you should migrate to using normal property access instead. TypeScript will help make this a smooth process by identifying where you need to handle null and undefined intermediate properties.

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+
+### Service and controller injections
+
+Ember does service and controller lookups with the `inject` functions at runtime, using the name of the service or controller being injected up as the default value—a clever bit of metaprogramming that makes for a nice developer experience. TypeScript cannot do this, because the name of the service or controller to inject isn't available at compile time in the same way.
+
+The officially supported method for injections with TypeScript uses _decorators_.
```
chriskrycho, line 130:
```suggestion
The officially supported method for injections with TypeScript uses _decorators_ along with a “type registry” which maps the string names of services to the corresponding type. For example, a session service which defines a `login` method might be defined like this—notice the `declare module '@ember/service`:
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+Then we can use the service as we usually would with a decorator, but adding a type annotation to it so TypeScript knows what it's looking at:
+
+```typescript
+// my-app/components/user-profile.ts
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+
+import MySession from 'my-app/services/my-session';
```
chriskrycho, line 157:
```suggestion
import type MySession from 'my-app/services/my-session';
```

guides/release/using-typescript/ts-and-ember.md
chriskrycho, line 130:
```suggestion
The officially supported method for injections with TypeScript uses _decorators_. Since decorators do not currently have access to enough information to produce an appropriate type, we need to import and name the type explicitly. For example, we might have `MySession` service which defines a `login` method, defined as usual:
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+  }
+}
+```
+
+Note that we need the `MySession` type annotation this way, but we _don't_ need the string lookup (unless we're giving the service a different name than the usual on the class, as in Ember injections in general). Without the type annotation, the type of `session` would just be `any`. This is because decorators are not allowed to modify the types of whatever they decorate. As a result, we wouldn't get any type-checking on that `session.login` call, and we wouldn't get any auto-completion either. Which would be really sad and take away a lot of the reason we're using TypeScript in the first place!
+
+Also notice [the `declare` property modifier](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier). This tells TypeScript that the property will be configured by something outside the class (in this case, the decorator), and guarantees it emits spec-compliant JavaScript.
+
+(This also holds true for all other service injections, computed property macros, and Ember Data model attributes and relationships.)
+
```
chriskrycho, line 173:
~~~suggestion

Finally, you may have noticed the `declare module` at the bottom of the example `MySession` definition:

```typescript
declare module '@ember/service' {
  interface Registry {
    'my-session': MySession;
  }
}
```

This definition allows for type-safe lookups with other Ember dependency injection APIs. For example, [the `Owner.lookup` method](https://api.emberjs.com/ember/5.2/classes/Owner#2-method) uses this "registration"—a mapping from the string `'my-session'` to the service type, `MySession`—to provide the correct type:

```typescript
function dynamicLookup(owner: Owner) {
  let mySession = owner.lookup('service:my-session');
  mySession.login("tom@example.com", "password123");
}
```

~~~

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+export default class UserProfile extends Component {
+  username?: string;
+}
+```
+
+When invoked via a template `{{user-profile username='example123'}}`, you would expect that `username` would have the value of `example123`, however prior to the native class feature released in Ember `3.6`, this will result in `username` being undefined.
+
+For users who remain on Ember versions below `3.6`, please use [https://github.com/pzuraq/ember-native-class-polyfill](https://github.com/pzuraq/ember-native-class-polyfill)
+
```
chriskrycho, line 193: BALETED

```suggestion
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+   ```typescript
+   // @ts-ignore
+   get(someObject, 'deeply.nested.key');
+   ```
+
+   It's usually best to include an explanation of _why_ you're ignoring a lookup!
+
+### Service and controller injections
+
+Ember does service and controller lookups with the `inject` functions at runtime, using the name of the service or controller being injected up as the default value—a clever bit of metaprogramming that makes for a nice developer experience. TypeScript cannot do this, because the name of the service or controller to inject isn't available at compile time in the same way.
```
chriskrycho, line 128:
```suggestion
Ember looks up services with the `@service` decorator at runtime, using the name of the service being injected up as the default value—a clever bit of metaprogramming that makes for a nice developer experience. TypeScript cannot do this, because the name of the service to inject isn't available at compile time in the same way. (These same considerations apply to controller injections using the `@inject` decorator from `@ember/controller`.)
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+}
+```
+
+When invoked via a template `{{user-profile username='example123'}}`, you would expect that `username` would have the value of `example123`, however prior to the native class feature released in Ember `3.6`, this will result in `username` being undefined.
+
+For users who remain on Ember versions below `3.6`, please use [https://github.com/pzuraq/ember-native-class-polyfill](https://github.com/pzuraq/ember-native-class-polyfill)
+
+### Ember Data lookups
+
+We use the same basic approach for Ember Data type lookups with string keys as we do for service or controller injections. As a result, once you add the module and interface definitions for each model, serializer, and adapter in your app, you will automatically get type-checking and autocompletion and the correct return types for functions like `findRecord`, `queryRecord`, `adapterFor`, `serializerFor`, etc. No need to try to write out those (admittedly kind of hairy!) types; just write your Ember Data calls like normal and everything _should_ just work.
```
chriskrycho, line 196:
```suggestion
We use the same basic approach for Ember Data type lookups with string keys as we do for service injections, but here we take advantage of the string "type registration" for the runtime code as well. As a result, once you add the module and interface definitions for each model, serializer, and adapter in your app, you will automatically get type-checking and autocompletion and the correct return types for functions like `findRecord`, `queryRecord`, `adapterFor`, `serializerFor`, etc. No need to try to write out those (admittedly kind of hairy!) types; just write your Ember Data calls like normal and everything _should_ just work.
```

guides/release/using-typescript/ts-and-ember.md
```diff
@@ -0,0 +1,317 @@
...
+- `Assertion Failed: Attempting to lookup an injected property on an object without a container, ensure that the object was instantiated via a container.` – You failed to pass `...arguments` when you called `super` in e.g. a component class `constructor`. Always do `super(...arguments)`, not just `super()`, in your `constructor`.
+
+## Type definitions outside `node_modules/@types`
+
+By default, the TypeScript compiler loads all type definitions found in `node_modules/@types`. If the type defs you need are not found there and are not supplied in the root of the package you're referencing, you can register a custom value in `paths` in the `tsconfig.json` file. See the [tsconfig.json docs](http://www.typescriptlang.org/docs/handbook/compiler-options.html#compiler-options) for details.
```
chriskrycho, line 317: DELETE (long since irrelevant!)

```suggestion
```

chriskrycho (commented):
Thanks for getting this started! I left a bunch of suggestions, and I think we want to keep these *out* of the top-level outline for now, but getting it started this way is awesome, and now we can iterate!

guides/release/using-typescript/signatures.md
```diff
@@ -0,0 +1,751 @@
...
+
+Because the signature set on the class, callers would still have to pass a single string argument, but we would need to change the behavior of the body to [narrow the type][narrowing] for the first item in the array.
+
+[narrowing]: http://www.typescriptlang.org/docs/handbook/2/narrowing.html
+
+Accordingly, the best practice is to keep the types matching.
+
+## Advanced signature techniques
+
+We can also define signatures in more complicated ways, using TypeScript features like generics and union types, to [make illegal states unrepresentable][illegal]. For example, a list component might need to yield back out instances of the same type it provides, and provide the appropriate element target based on a `type` argument.
```
chriskrycho, line 634:
```suggestion
We can also define signatures in more complicated ways using more advanced TypeScript features.
Nearly anything you can do with a “regular” TypeScript function or class, you can also do with signatures for Glimmer invokables.
We can make a component accept a generic type, or use union types.
With these tools at our disposal, we can even define our signatures to [make illegal states unrepresentable][illegal].

To see this in practice, consider a list component which yields back out instances of the same type it provides, and provide the appropriate element target based on a `type` argument.
Yielding back out the same type passed in will use generics, and providing an appropriate element target for `...attributes` can use a union type.
```

guides/release/using-typescript/signatures.md
```diff
@@ -0,0 +1,757 @@
...
+
+When this component is invoked, the `@type` argument will determine what kinds of modifiers are legal to apply to it. For example, if you defined a modifier `reverse` which required an `HTMLOListElement`, this invocation would be rejected:
+
+```handlebars
+<List @items={{array 1 2 3}} @type='unordered' {{reverse}} as |item|>
+  The item is
+  {{item}}.
+</List>
+```
+
```
chriskrycho, line 704:
```suggestion

The same approach with generics works for class-based helpers and class-based modifiers.
Function-based helpers as and modifiers can also use generics, but by using them on the function definition rather than via a signature.
One caveat: particularly complicated union types in signatures can sometimes become too complex for Glint/TypeScript to resolve when invoking in a template.
In those cases, your best bet is to find a simpler way to structure the types while preserving safety.

```

guides/release/using-typescript/signatures.md
```diff
@@ -38,25 +37,26 @@ interface InvokableSignature {
...
-1. In practice, while you _can_ write a full signature like this for any invokable, you never _need_ to. Different kinds of invokables care about different subsets of this set of features:
+[glint]: https://github.com/typed-ember/glint
 
-   - Helpers may have arguments and return values, but do not yield blocks and do not have an associated element.
-   - Components may have arguments, blocks, and an associated element, but never return values.
-   - Modifiers may have arguments and always have an associated element, but do not return values or have blocks.
+A few things to note about these signatures:
 
-   Accordingly, we supply simpler forms of signatures appropriate to each type of invokable.
+<!-- TODO: This was an ordered list, but nested list formatting is a bit funky currently, so I converted it to paragraphs. Would be nice to revert. -->
```
chriskrycho: Honestly, the "First… Second… Third" is probably better anyway!

guides/release/using-typescript/signatures.md
```diff
@@ -228,7 +228,7 @@ export default class AudioPlayer extends Component<AudioPlayerSignature> {
 <button type='button' {{on 'click' this.pause}}>Pause</button>
 ```
 
-Let’s go one step further and switch to supporting for two [named blocks][nb]: a `title` block for a caption for the audio element, and an `audio` block for the audio fallback where we previously used a `default` block.
+Let’s go one step further and switch to supporting for two [named blocks][nb]: a optional `title` block for a caption for the audio element, and a `fallback` block for the audio fallback where we previously used a `default` block.
```
chriskrycho: Grammar nit:

```suggestion
Let’s go one step further and switch to supporting for two [named blocks][nb]: an optional `title` block for a caption for the audio element, and a `fallback` block for the audio fallback where we previously used a `default` block.
```

guides/release/using-typescript/core-concepts/routing.md
```diff
@@ -44,10 +82,9 @@ How that works:
 
 - `Resolved<P>` says "if this is a promise, the type here is whatever the promise resolves to; otherwise, it's just the value"
```
chriskrycho, line 83: We should update this to use the `Awaited<T>` type that TS ships natively as of late in the 4.X series!

jenweber:
Wow, this is amazing and there is so much content here! Incredible work.

What kind of feedback do you want the most at this stage? When do you think this might be ready to merge? (Now, after more feedback, do you know of any blockers, etc)

gitKrystan:
IMO this is ready to merge, but no one else has read it.

MinThaMie (commented):
Discussed today during the learning meeting
- Remove the "under construction message"
- Add it to the table of content
  - Don't call them the same as in the current menu, but make it something like "TypeScript & EmberData"
 
 We would love to merge it sooner rather then later, because it looks great!

gitKrystan:
Thanks! Done!

image

guides/release/using-typescript/additional-resources/legacy.md
```diff
@@ -0,0 +1,204 @@
...
+}
+```
+
+<!-- Internal links -->
+
+[components]: ../../core-concepts/invokables/#toc_glimmer-components
+[global-types]: ../../additional-resources/faq/#toc_global-types-for-your-project
+[migrate-from-computed]: ../../../upgrading/current-edition/tracked-properties/
+[migrate-from-mixins]: ../../../upgrading/current-edition/native-classes/#toc_mixins
+[migrating-from-classic-classes]: http://localhost:4200/release/upgrading/current-edition/native-classes/
```
gitKrystan, line 197:
```suggestion
[migrating-from-classic-classes]: ../../../upgrading/current-edition/native-classes/
```

guides/release/pages.yml
```diff
@@ -249,6 +249,46 @@
   pages:
     - title: "Managing Dependencies"
       url: "index"
+- title: "Using TypeScript"
+  url: "using-typescript"
```
mansona, line 253: one thing that is important once we release something is the URLs, we need to always keep redirects and never break a link. This means that we need to be sure that we are happy with the URLs that we are generating.

Personally I think `using` in this URL is too much 🤔 i.e. `/release/using-typescript/` is not as good as `/release/typescript/`. What do you think?

guides/release/using-typescript/application-development/index.md
```diff
@@ -0,0 +1,3 @@
+---
+redirect: using-typescript/application-development/configuration
```
mansona, line 2: same here, why do we have this redirect?

guides/release/using-typescript/additional-resources/index.md
```diff
@@ -0,0 +1,3 @@
+---
+redirect: using-typescript/additional-resources/legacy
```
mansona, line 2: why do we have a redirect for a page that hasn't launched yet? What I said before about "never breaking a link" is true, but we don't need to do that for things that are not released yet 🤔

guides/release/using-typescript/application-development/converting-an-app.md
```diff
@@ -0,0 +1,105 @@
+These directions are for converting an _existing_ Ember app to TypeScript. If you are starting a new app, you can use the directions in [Getting Started].
+
+## Enable TypeScript Features
+
+### A Bit of a Hack
+
+Since `ember-cli` _currently_ has no flag to convert your project to TypeScript, we're going to use a bit of a hack and _temporarily_ install the legacy `ember-cli-typescript` addon to complete most of the migration:
```
gitKrystan, line 7: This approach kinda stinks but I experimented w/ different ways and this was the least obnoxious one. I have it on my TODO list to add an appropriate flag...probably to ember-cli-update.

mansona (commented):
Just some minor comments, I haven't reviewed the contents (I'm not sure I'm qualified 😂)

One other thing it might be worth considering is to split up the invokables docs into 2 pages because the diff was so big that github didn't want to show it at first 🙈

guides/release/using-typescript/application-development/index.md
gitKrystan: IIRC clicking the top-level header was broken w/o an "index.md". Will double-check.

mansona: ah yes that will probably be true, but there should be an index and it should have content instead of a redirect ;) in this case you would just need to rename the `configuration` file to be `index.md` and then that would be the first thing in the section 👍 the name of the page can still be `Configuration` it just changes the url structure

gitKrystan: I have some ideas to fix. Please hold.

guides/release/typescript/additional-resources/legacy.md
```diff
@@ -0,0 +1,204 @@
...
+  bio: computed('name', 'age', function() {
+  bio: computed('name', 'age', function(this: UserProfile) {
+    return `${this.get('name')} is `${this.get('age')}` years old!`;
+  }),
+})
+
+export default UserProfile;
+```
+
+The `this` type, tells TS to use `UserProfile` for `get` and `set` lookups; otherwise `this.get` would not know the types of `'name'` or `'age'` or even be able to suggest them for autocompletion.
```
gitKrystan, line 166:
```suggestion
The [`this` type][this], tells TS to use `UserProfile` for `get` and `set` lookups; otherwise `this.get` would not know the types of `'name'` or `'age'` or even be able to suggest them for autocompletion.
```

guides/release/typescript/application-development/addons.md
```diff
@@ -0,0 +1,174 @@
...
+
+<!-- External links -->
+
+[DefinitelyTyped]: https://github.com/DefinitelyTyped/DefinitelyTyped
+[dts]: https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html
+[ember-addon]: https://cli.emberjs.com/release/writing-addons/
+[glint]: https://typed-ember.gitbook.io/glint/
+[In-repo addons]: https://cli.emberjs.com/release/writing-addons/in-repo-addons/
+[npm-link]: https://docs.npmjs.com/cli/link
+[typesVersions]: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions
```
gitKrystan, line 173:
```suggestion
[this]: https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function
[typesVersions]: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions
```

guides/release/typescript/core-concepts/ember-data.md
```diff
@@ -0,0 +1,258 @@
...
+If you supply a custom transform, you will get back the type returned by your transform.
+
+So, for example, you might write a class like this:
+
+```typescript {data-filename="app/models/user.ts"}
+import Model, { attr } from '@ember-data/model';
+import CustomType from '../transforms/custom-transform';
+
+export default class User extends Model {
+  @attr()
```
runspired, line 31:
```suggestion
  @attr
```

guides/release/typescript/core-concepts/ember-data.md
```diff
@@ -0,0 +1,258 @@
...
+
+The _safest_ type you can write for an EmberData model, therefore, leaves every property optional: this is how models _actually_ behave. If you choose to mark properties as definitely present by leaving off the `?`, you should take care to guarantee that this is a guarantee your API upholds, and that ever time you create a record from within the app, _you_ uphold those guarantees.
+
+One way to make this safer is to supply a default value using the `defaultValue` on the options hash for the attribute:
+
+```typescript {data-filename="app/models/user.ts"}
+import Model, { attr } from '@ember-data/model';
+
+export default class User extends Model {
+  @attr()
```
lukasnys, line 63: Following runspired's suggestion for removing parentheses above, should they also be removed here?

```suggestion
  @attr
```

guides/release/typescript/core-concepts/ember-data.md
```diff
@@ -0,0 +1,258 @@
...
+    cursor: Cursor;
+  }
+}
+```
+
+Note that you should declare your own transform under [`TransformRegistry`][ED-registry] to make `@attr` to work with your transform.
+
+## Serializers and Adapters
+
+EmberData serializers and adapters are normal TypeScript classes. The only related gotcha is that you must [register][ED-registry] your them with a declaration:
```
lukasnys, line 213:
```suggestion
EmberData serializers and adapters are normal TypeScript classes. The only related gotcha is that you must [register][ED-registry] them with a declaration:
```

guides/release/typescript/core-concepts/invokables.md
```diff
@@ -0,0 +1,733 @@
...
+
+```handlebars
+<List @items={{array 1 2 3}} @type='unordered' {{reverse}} as |item|>
+  The item is
+  {{item}}.
+</List>
+```
+
+The same approach with generics works for class-based helpers and class-based modifiers.
+Function-based helpers as and modifiers can also use generics, but by using them on the function definition rather than via a signature.
```
lukasnys, line 711:
```suggestion
Function-based helpers and modifiers can also use generics, but by using them on the function definition rather than via a signature.
```

lukasnys (changes requested):

lukasnys:
Very excited to see this land! 🚀

guides/release/typescript/index.md
```diff
@@ -0,0 +1,43 @@
+This guide is designed to help you get up and running with TypeScript in an Ember app.
+
+This is _not_ an introduction to TypeScript _or_ Ember. Throughout this guide, we’ll link back to [the TypeScript docs][typescript-docs] and [the Ember Guides][ember-guides] when there are specific concepts that we will not explain here but which are important for understanding what we’re covering!
```
jaredgalanis, line 3: This makes it sound like we are linking to the Ember Guides as a separate resource, but I think we're actually making this a part of the Ember Guides. Maybe:

```suggestion
This is _not_ an introduction to TypeScript _or_ Ember. Throughout this guide, we’ll link back to [the TypeScript docs][typescript-docs] and to other sections of [the Ember Guides][ember-guides] when there are specific concepts that we will not explain here but which are important for understanding what we’re covering!
```

guides/release/typescript/core-concepts/index.md
```diff
@@ -0,0 +1,11 @@
+In the next sections, we will cover how to use TypeScript effectively with specific Ember.js APIs.
+
+We do _not_ cover general usage of Ember; instead, we assume that as background knowledge. Please see the [Ember Guides][ember-guides] and [API docs][api-docs]!
```
jaredgalanis, line 3: Same, concern here as https://github.com/ember-learn/guides-source/pull/1960/files#r1367496976. 

Maybe:

```suggestion
We do _not_ cover general usage of Ember; instead, we assume that as background knowledge. Please see other sections of the [Ember Guides][ember-guides] and [API docs][api-docs]!
```

guides/release/typescript/core-concepts/services.md
```diff
@@ -0,0 +1,115 @@
+Ember [Services] are global singleton classes that can be made available to different parts of an Ember application via dependency injection. Due to their global, shared nature, writing services in TypeScript gives you a build-time-enforceable API for some of the most central parts of your application.
+
+## A Basic Service
+
+Let's take this example from the [Ember Guide][example-location]:
```
jaredgalanis, line 5: Same concern. Maybe:

```suggestion
Let's take this example from elsewhere in the [Ember Guides][example-location]:
```

jaredgalanis (commented):
@gitKrystan, I've made a few comments that I wouldn't classify as blocking. These guides are incredible, thank you so much for doing this! 🔥🎉

guides/release/typescript/index.md
chriskrycho: Heh, that's because this content is all originally pulled from (though *massively* updated, thanks to @gitKrystan’s amazing efforts) https://docs.ember-cli-typescript.com (which can go away/redirect here once this lands!). So originally it *was* just pointing the the Ember Guides as a separate resource. 😅

gitKrystan:
Lint should be fixed now.

I played w/ moving the chapter up a level (per discussion on Discord) and it makes the initial TOC super long + pushes some of the TypeScript chapters to the same level as their top-level equivalent (e.g. "Getting Started" at the same level as "TypeScript: Getting Started") which I found a bit confusing.

PR #1963, opened 2023-09-26
Update quick-start.md

kalimdumas:
In the "diff" portion, it doesn't include deleting the h2; however, in the next paragraph, it says that the page should look identical. If you don't delete the h2, then it will display "List of Scientists" twice in 2 different h2s.


MinThaMie:
Thanks for bringing this to our attention, but with a recent update it's been fixed :) 

Screenshot 2023-10-09 at 16 08 19

PR #1966, opened 2023-10-15
Update rendering-values.md with example of how to return a component from a function

BryanCrotaz:
Add example of function to return a dynamic component

### DIFF
--- a/guides/release/in-depth-topics/rendering-values.md
+++ b/guides/release/in-depth-topics/rendering-values.md
@@ -71,6 +71,22 @@ or via a property on some object
 ```handlebars
 <this.someComponent />
 ```
+```ts
+import MyOtherComponent from 'my-app/components/my-other-component';
+import MySuccessComponent from 'my-app/components/my-success-component';
+
+export default class MyComponent extends Component {
+  @tracked status: string;
+
+  get someComponent(): typeof Component<any> {
+    if (this.status == 'success') {
+      return MySuccessComponent;
+    } else {
+      return MyOtherComponent;
+    }
+  }
+}
+```
 or via an argument passed to a component
 ```handlebars
 <@someComponent />


NullVoxPopuli (approved):
very nice!

MinThaMie (commented):
Thanks for this example, I'm wondering if it needs a little bit of text too. What this does :)

NullVoxPopuli:
It's about the same as all the other examples on the page. Can we merge this change, and then iterate on the page's tone and verbiage overall separately?

MinThaMie:
All the examples have a sentence that explains it. I would love to have that for this one too :)
Screenshot 2023-10-16 at 16 15 24

BryanCrotaz:
"or via a property on this object" only showed the template, and not how to define the property. I've filled that gap. It was only showing half of the how-to.

MinThaMie:
Aaah, since your component is called `myComponent` and the example uses `someComponent`, could we change that? And maybe add the sentence: "The implementation inside the component could look like this:" and then your example?

BryanCrotaz:
No, because that would be REALLY confusing.

guides/release/in-depth-topics/rendering-values.md
```diff
@@ -71,6 +71,25 @@ or via a property on some object
 ```handlebars
 <this.someComponent />
 ```
+```ts
```
mansona, line 74: There shouldn't be any TypeScript in the guides (yet). There is a whole process that is underway to add a typescript version of the guides but until that is done then you should write all your backing JS as JS 👍

guides/release/in-depth-topics/rendering-values.md
```diff
@@ -71,6 +71,25 @@ or via a property on some object
 ```handlebars
```
mansona, line 71: If we are trying to equate this file to the JS backing it then we should specify the filename for both. There is information in the styleguide on how to do this https://github.com/ember-learn/guides-source/blob/master/CONTRIBUTING.md#style-guide

mansona (changes requested):

PR #1978, opened 2023-11-20
WIP: Add Template Tag format guides

IgnaceMaes:
image

⚠️ Work in progress! ⚠️ 

While eventually all guide material will have to be updated to use the `<template>` tag authoring format, this PR serves as a first introduction to the format. This unblocks moving forward https://github.com/emberjs/rfcs/pull/871.

Sources:
- https://rfcs.emberjs.com/id/0779-first-class-component-templates/
- https://github.com/ember-template-imports/ember-template-imports

## To review

Review for:

- Typos
- Mistakes
- Completeness

Please make liberal use of the GitHub suggestions feature because I am starting to get bleary-eyed looking at this for so long. 🙏

### DIFF
--- a/guides/release/components/template-tag-format.md
+++ b/guides/release/components/template-tag-format.md
@@ -1,0 +1,39 @@
+The template tag component format allows defining a component's template using a `<template>` tag block. This block is a first-class participant in JavaScript and TypeScript with strict mode template semantics. For this new format, the extensions `.gjs` and `.gts` are used respectively.
+
+Template tag components address a number of pain points when defining a template as a separate file, and provide a number of new capabilities:
+
+- Accessing local JavaScript values requires no backing class, enabling much easier use of existing JavaScript ecosystem tools.
+- Authoring more than one component in a single file, where colocation makes sense.
+- Creating locally-scoped helpers, modifiers, and other JavaScript functionality *just works*.
+
+## Installation
+
+To use template tag components, you must install the [ember-template-imports](https://github.com/ember-template-imports/ember-template-imports) addon:
+
+```bash
+npm add --save-dev ember-template-imports
+```
+
+For integration with tools like Prettier and Glint, and information on version compatibility, refer to the addon's readme.
+
+## Using Template Tags and `.gjs`/`.gts` Files
+
+The new `<template>` tag format is available in `.gjs` and `.gts` files. These file extensions represent a new file format "GlimmerJS" and "GlimmerTS", which are supersets of standard JavaScript and TypeScript respectively. In this syntax, templates are defined in JavaScript files directly.
+
+This example defines a template-only component, which is the default export of hello.gjs:
+
+```text {data-filename="app/components/hello.gjs"}
+<template>
+  <span>Hello, {{@name}}!</span>
+</template>
+```
+
+You would be able to use this component in another component like so:
+
+```text {data-filename="app/components/world.gjs"}
+import Hello from 'example-app/components/hello';
+
+<template>
+  <Hello @name="world" />
+</template>
+```
--- a/guides/release/pages.yml
+++ b/guides/release/pages.yml
@@ -86,6 +86,8 @@
       url: template-lifecycle-dom-and-modifiers
     - title: "Built-in Components"
       url: "built-in-components"
+    - title: "Template Tag Format"
+      url: "template-tag-format"
 - title: "Routing"
   url: "routing"
   pages:


guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,171 @@
+The template tag component format allows defining a component's template using a `<template>` tag block. This block is a first-class participant in JavaScript and TypeScript with strict mode template semantics. For this new format, the extensions `.gjs` and `.gts` are used respectively.
```
NullVoxPopuli, line 1:
```suggestion
The template tag component format allows defining a component's template using a `<template>` tag block. This block is a first-class participant in JavaScript and TypeScript with strict mode template semantics. For this new format, the extensions `.gjs` and `.gts` are used for glimmer-flavored JavaScript and glimmer-flavored typescript respectively.
```

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,171 @@
...
+
+Template tag components address a number of pain points when defining a template as a separate file, and provide a number of new capabilities:
+
```
NullVoxPopuli, line 4: Most importantly: knowing where things come from and where to find them

ef4:
One of the biggest teaching points I think we need to hit here is how to locate the components/helpers/modifiers you're trying to import. There are generally-applicable, simple rules that work for essentially *all* addons. For example, an addon allowing you to importi a component from `your-app-name/components/the-component-name` is effectively mandatory.

ef4:
Another gotcha to document: to use an imported template, it needs to not be using the old, pre-template-colocation format where the templates is in `app/templates/components` with the JS in `app/components`. People should ideally codemod off that format before trying to use template tag.

If you must use it, you can use setComponentTemplate manually to stick the template and JS together, but probably that is not guides-level content.

guides/release/components/template-tag-format.md
IgnaceMaes: Adapted in rewrite.

guides/release/components/template-tag-format.md
IgnaceMaes: Adapted in rewrite.

IgnaceMaes:
Thanks @NullVoxPopuli @ef4 for the feedback.

I did a full rewrite from scratch instead of starting from the `ember-template-imports` README. It's restructured beginning at feature parity with the current format, only to introduce new capabilities afterwards. That makes most sense to me.

Marked as ready for review!

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+    </aside>
+  </template>
+}
+```
+
+## Importing components, helpers, and modifiers
+
+The template tag format uses strict template semantics. This means that you can only use components, helpers, and modifiers that are explicitly imported. This makes it easier to understand where things come from and what they do, as well as unlocks build optimizations.
+
+### Importing invokables from the own app
```
evoactivity, line 65: typo here should be

### Importing invokables from your own app

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+  <Avatar
+    @title="Picture of Zoey"
+    @initial="Zoey"
+  />
+</template>
+```
+
+#### Helpers and modifiers
+
+Importing helpers and modifiers from the own app also follows the same principle of using standard JavaScript import syntax. Instead of importing from `app/components`, the path to import from is `app/helpers` and `app/modifiers` respectively.
```
evoactivity, line 127: `from the own app` should be `from your own app`

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Importing from addons
+
+Just as with components, helpers, and modifiers from the own app, external invokables from addons also have to be imported. This is done using the same `import` statement, but with a path referencing the addon.
```
evoactivity, line 131: `from the own app` should be `from your own app`

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+      <template>
+        <Avatar @title="Picture of Zoey" @initial={{initial}} />
+      </template>
+    );
+    assert.dom().hasText(initial);
+  });
+});
+```
+
+Notice how the same semantics now apply to tests as well: local values in scope can be referenced directly, and invokables from the own app or addons need to be imported.
```
evoactivity, line 245: `from the own app` should be `from your own app`

evoactivity (commented):
Just a few typos, rest looks good to me

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+The example above demonstrates defining a "Message" template-only component. The import syntax for class-based components is the same.
+
+<div class="cta">
+  <div class="cta-note">
+    <div class="cta-note-body">
+      <div class="cta-note-heading">Zoey says...</div>
+      <div class="cta-note-message">
+        The components that are imported are not required to use the new template tag format. This is intentional, and very powerful, as it <strong>allows incremental adoption</strong> of the new format.
+        <br><br>
+        The only prerequisite is that the component is defined using the template-colocation structure instead of the separating the JavaScript and Glimmer template code into separate folders.
```
lukasnys, line 92:
```suggestion
        The only prerequisite is that the component is defined using the template-colocation structure instead of separating the JavaScript and Glimmer template code into separate folders.
```

guides/release/components/template-tag-format.md
IgnaceMaes: Thanks! Adapted them all.

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
+The template tag format is a powerful, new way to write components in Ember. It's a single-file format that combines the component's JavaScript and Glimmer template code. The `<template>` tag is used to keep a clear separation between the template language and the JavaScript around it.
```
NullVoxPopuli, line 1: In also unifies around the concept of ES Modules, which old components did not. Idk if that's too much no mention is the opening paragraph tho

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+
+Template tag components use the file extension `.gjs`. This abbreviation is short for "Glimmer JavaScript". The file extension `.gts` is also supported for TypeScript components.
+
+This new format is the official future of Ember's component authoring story, and is stable and usable today. We expect it to become the recommended way of authoring all Ember apps in the near future, once we are satisfied that we have sufficiently polished up all the corners of the implementation.
```
NullVoxPopuli, line 5:
```suggestion
This new format is the official future of Ember's component authoring story, and is stable and usable today. We expect it to become the recommended and default way of authoring all Ember apps in the near future, once we are satisfied that we have sufficiently polished up all the corners of the implementation.
```

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Writing template tag components
+
+Just like with separate JavaScript and Glimmer template files, the template tag format has the concept of template-only components and class-based components. Let's take a closer look at how these concepts compare between both component formats in the next section.
+
+### Template-only components
+
+The following template-only component was created in a previous section to extract an avatar layout into a reusable component.
```
NullVoxPopuli, line 13: Should/ can that previous section be linked? I don't know about it. Lol

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+
+```handlebars {data-filename="app/components/avatar.hbs"}
+<aside>
+  <div class="avatar" title={{@title}}>{{@initial}}</div>
+</aside>
+```
+
+This layout can be turned into a template tag component by wrapping the code in a `<template>` tag and changing the file extension to `.gjs`.
+
+```text {data-filename="app/components/avatar.gjs"}
```
NullVoxPopuli, line 23:
```suggestion
```gjs {data-filename="app/components/avatar.gjs"}
```
Perhaps?

Though, last i tried this, showdown-prism didn't support gjs in prerender/fastboot mode.

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+The example above demonstrates defining a "Message" template-only component. The import syntax for class-based components is the same.
+
+<div class="cta">
+  <div class="cta-note">
+    <div class="cta-note-body">
+      <div class="cta-note-heading">Zoey says...</div>
+      <div class="cta-note-message">
+        The components that are imported are not required to use the new template tag format. This is intentional, and very powerful, as it <strong>allows incremental adoption</strong> of the new format.
+        <br><br>
+        The only prerequisite is that the component is defined using the template-colocation structure instead of splitting up the JavaScript and Glimmer template files into separate folders.
```
NullVoxPopuli, line 92: Does it make sense to link to the rfc trying to deprecate the old stuff?

https://github.com/emberjs/rfcs/pull/995

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+<Messages::Avatar
+  @title="Picture of Zoey"
+  @initial="Zoey"
+/>
+```
+
+This quirk is no longer necessary with the template tag format. Instead, importing now works the same as importing any other JavaScript module.
+
+```text {data-filename="app/components/avatar-usage.gjs"}
+import Avatar from 'app/components/messages/avatar';
```
NullVoxPopuli, line 115: Relate better since the file is sibling

```suggestion
import Avatar from './messages/avatar';
```

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+### Locally-scoped values
+
+The template tag format follows JavaScript module syntax. Any value that isn't exported is only available locally within the file. This is useful for defining helper functions that are only used within the component, or for defining constants that are used multiple times within the template.
+
+In the following example, a "Square" component is defined that calculates the square of a number. The `value` constant is defined locally, and the `square` helper function is only available within the component.
+
+```text {data-filename="app/components/square.gjs"}
+const value = 2;
+
+const square = (number) => {
```
NullVoxPopuli, line 178: May as well use `function` if using curlies, imo

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+<template>
+  The square of {{value}} equals {{square value}}
+</template>
+```
+
+This will render to `The square of 2 equals 4`.
+
+### Multiple components per file
+
+The template tag format allows defining multiple components within a single file. This is useful for defining components that are closely related to each other, but are not used in other parts of the app.
```
NullVoxPopuli, line 191: Yaaaaas!

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+npm add --save-dev ember-template-imports
+```
+
+### Integration with external tooling
+
+You may need to upgrade dependency versions or install additional plugins to have proper integration with external tools. The following commonly-used tools are supported:
+
+- [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint): Versions 5.8.0 and up.
+- [eslint-plugin-ember](https://github.com/ember-cli/eslint-plugin-ember): Versions 11.6.0 and up.
+- [Prettier](https://github.com/prettier/prettier): Versions 3.0.0 and up. This requires installing the [prettier-plugin-ember-template-tag](https://github.com/gitKrystan/prettier-plugin-ember-template-tag).
```
NullVoxPopuli, line 261:
```suggestion
- [Prettier](https://github.com/prettier/prettier): Versions 3.1.0 and up. This requires installing the [prettier-plugin-ember-template-tag](https://github.com/gitKrystan/prettier-plugin-ember-template-tag).
```

3.0 had issues with detecting plugins :(

NullVoxPopuli (commented):
Very nice work! This is super exciting!

guides/release/components/template-tag-format.md
IgnaceMaes: This crashes the app at runtime:

```
Uncaught (in promise) Error: Assertion Failed: Language "gjs" not found. Have you configured Prism correctly?
```

So it would require changes in https://github.com/ember-learn/guidemaker-ember-template. Seems okay to leave as is until highlighting is properly supported?

guides/release/components/template-tag-format.md
IgnaceMaes: Good point! Adapted.

guides/release/components/template-tag-format.md
NullVoxPopuli: Ya

guides/release/components/template-tag-format.md
IgnaceMaes: I fear that goes a bit too deep. It's super interesting to know and realize, but it also requires you to have prior understanding of what the concept of ES Modules are. And I don't want people not being familiar with that being scared off.

A separate section/article about the more technical concepts or reasoning behind the syntax would be more suiting. But I think that could be a separate PR.

guides/release/components/template-tag-format.md
IgnaceMaes: Added!

guides/release/components/template-tag-format.md
IgnaceMaes: Hmm, I'm not sure linking to an open (not yet accepted) RFC is the best idea.

I've added a link to the one introducing co-location instead: https://rfcs.emberjs.com/id/0481-component-templates-co-location

guides/release/components/template-tag-format.md
IgnaceMaes: Could you elaborate on "better"?

I'm kind of split on this. Absolute paths might be easier to understand in docs as there is no visible file explorer, and thus less "magic" going on. On the other hand, V2 addons mostly use relative imports everywhere already.

NullVoxPopuli: in real apps, too, folks often assume "absolute paths only", and then their imports get massively long.

> there is no visible file explorer,

however, we do specify the path
image

IgnaceMaes: I'm slightly in favour for absolute paths for readability in guides. But no strong opinion to block this so I've adjusted the imports.

Guess we can always revise later what makes most sense, especially when more and more docs will be using template tags.

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
+The template tag format is a powerful, new way to write components in Ember. It's a single-file format that combines the component's JavaScript and Glimmer template code. The `<template>` tag is used to keep a clear separation between the template language and the JavaScript around it.
+
+Template tag components use the file extension `.gjs`. This abbreviation is short for "Glimmer JavaScript". The file extension `.gts` is also supported for TypeScript components.
+
+This new format is the official future of Ember's component authoring story, and is stable and usable today. We expect it to become the recommended and default way of authoring all Ember apps in the near future, once we are satisfied that we have sufficiently polished up all the corners of the implementation.
```
mansona, line 5: Since we're talking about this being "official" it might be worth linking to the RFC here and mentioning what stage it's in? thoughts?

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+
+```text {data-filename="app/components/avatar.gjs"}
+<template>
+  <aside>
+    <div class="avatar" title={{@title}}>{{@initial}}</div>
+  </aside>
+</template>
+```
+
+The top-level template tag is exported as default component from the file. You *can* write this export explicitly, but it's not necessary. The following example is equivalent to the previous one.
```
mansona, line 31:
```suggestion
The top-level template tag is exported as the default component from the file. You *can* write this export explicitly, but it's not necessary. The following example is equivalent to the previous one.
```

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+    </aside>
+  </template>
+}
+```
+
+## Importing components, helpers, and modifiers
+
+The template tag format uses strict template semantics. This means that you can only use components, helpers, and modifiers that are explicitly imported. This makes it easier to understand where things come from and what they do, as well as unlocks build optimizations.
+
+### Importing invokables from your own app
```
mansona, line 65: `invokables` isn't defined before using it in this title. We need to define it in the previous paragraph **explicitly** 

Sure we are talking about invokables when saying `components, helpers, and modifiers` but we need to say something like "we call them invokables"

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+
+#### Helpers and modifiers
+
+Importing helpers and modifiers from your own app also follows the same principle of using standard JavaScript import syntax. Instead of importing from `app/components`, the path to import from is `app/helpers` and `app/modifiers` respectively.
+
+### Importing from addons
+
+Just as with components, helpers, and modifiers from your own app, external invokables from addons also have to be imported. This is done using the same `import` statement, but with a path referencing the addon.
+
+The structure of files within Ember addons is mostly standardized. This means that the path to import from can be derived from the addon's name. For example, an addon that is named `ember-foo` will have its components, helpers, and modifiers available as default import from `ember-foo/components/<component-name>`, `ember-foo/helpers/<helper-name>` and `ember-foo/modifiers/<modifier-name>` respectively.
```
mansona, line 133:
```suggestion
The structure of files within Ember addons is mostly standardized. This means that the path to import from can be derived from the addon's name. For example, an addon that is named `ember-foo` will likely have its components, helpers, and modifiers available as default import from `ember-foo/components/<component-name>`, `ember-foo/helpers/<helper-name>` and `ember-foo/modifiers/<modifier-name>` respectively.
```

We can't know for sure that this is true so we need to add uncertainty in the docs 🙈 I also think we need to add a sentence like "Refer to the addon's docs for more information"

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+
+Some addons may choose to re-export their invokables from the root index as named exports. (This is known as the barrel file pattern) Usually addons will document this usage in their README, which may look like `import { ComponentName } from 'ember-foo';`, if supported.
```
mansona, line 135:
```suggestion
Some addons may choose to re-export their invokables from the root index as named exports. Usually addons will document this usage in their README, which may look like `import { ComponentName } from 'ember-foo';`, if supported.
```

We can just remove that line 👍

mansona: also can we move the example of how this import looks to a full triple-tick code example? it's easy to miss this example when it's inline like this

guides/release/components/template-tag-format.md
```diff
@@ -0,0 +1,285 @@
...
+      </template>
+    );
+    assert.dom().hasText(initial);
+  });
+});
+```
+
+Notice how the same semantics now apply to tests as well: local values in scope can be referenced directly, and invokables from your own app or addons need to be imported.
+
+## Installation
```
mansona, line 247: I think that the installation being so low in the page feels a bit strange, especially if we want people to be able to follow along with examples. 

That being said, I feel like the tooling section belongs down here 😂 🙈 

What do you think?

mansona (commented):
The learning team did a review on this PR. These are minor changes but I think they should help get this perfectly ready to merge and launch 🎉 Let me know if you have any questions about my comments 👍

guides/release/components/template-tag-format.md
IgnaceMaes: I've included a link to the RFC. Repeating the stage here makes me concerned it might get forgotten when it moves to a next stage.

mansona: we did think about that, but the next stage is "ready for release" and the one after that is "recommended". To move to "recommended" we will be deleting this page (probably?) and all of the docs will be converted to GJS 👍 So I doubt it will be missed 😂

guides/release/components/template-tag-format.md
IgnaceMaes: Good catch.

Reworded this section based on https://guides.emberjs.com/release/typescript/core-concepts/invokables/.

guides/release/components/template-tag-format.md
IgnaceMaes: Adjusted! Better indeed.

Added a clearer example of the standard import as well.

guides/release/components/template-tag-format.md
IgnaceMaes: Added info about the RFC stage.

I do see value in these docs after reaching "Recommended". While the format will be adopted in all other pages, we'll need some place to link to explaining what is `.gjs`/ `.gts` and why Ember uses a custom format.

Other frameworks have this too:
- https://react.dev/learn/writing-markup-with-jsx
- https://svelte.dev/docs/svelte-components
- https://vuejs.org/api/sfc-spec

guides/release/components/template-tag-format.md
IgnaceMaes: I believe installation instructions in the guides in general, wherever on the page, feel strange as you'd expect to be able to use a component authoring format out of the box.

It was up at the top in the initial write up, but that felt "bloated" because there is quite a lot to it with the addon installation and editor setup.

I think having it down here is the best of both worlds. That way it's also easier for people not familiar with Ember, or not being around for a bit, to be able to get the concept first.

IgnaceMaes:
All comments should be addressed now, thanks for all the great input!

I've added another short section on importing conventions for helpers and modifiers. They change from kebab-case to camelCase as it now follows ES import standards. I've seen some confusion and questions on Discord about this so I believe was needed.

guides/release/components/template-tag-format.md
bertdeblock: I think the callout idea is good, but I didn't notice it at first 😬 Maybe the callout design is just a bit too stealthy.

IgnaceMaes: There's some minor usability tweaks I'd like to make to the docs.

But as that's out of scope for this PR I'm going to revolve this thread. The original comment has been addresses. If changes are wanted, feel free to open a new one! 😄

mansona (approved):
LGTM 💪 🎉 🚀

PR #1977, opened 2023-11-05
🐹 Migrate to ember 4

bartocc:
Migrate to ember 4

Ember 3.x deprecations had to be removed to make this possible:

- use ember-metrics@1.5.2
- use ember-cli-fastboot@3+
- use a version of guidemaker without ember 3.x deprecations, from https://github.com/empress/guidemaker/pull/94

The unit test `Unit | Service | head data: title with currentPage and currentSelect details` is failing because of the ember 3.x deprecation `computed-property.override`.

I believe that:

1. the test should be re-written with a mock of `service:page` instead of looking up for it and overriding its `currentSection` and `currentPage` CPs
2. the test should belong to https://github.com/empress/guidemaker, not to https://github.com/ember-learn/guides-source, since the `head-data` service is defined there…

PR #1980, opened 2023-11-27
✅ 🔥 Delete "Unit | Service | head data" test file

bartocc:
As discussed with @mansona on 2023-11-27 during the ember-learning team weekly meeting:

- testing meta data should be done with an acceptance test, not a unit test
- this test does not belong to the guides-source project, but rather to `empress/guidemaker`.

Issue https://github.com/empress/guidemaker/issues/96 on `empress/guidemaker` lets us remember that we need to write this acceptance test, with inspiration from https://github.com/empress/empress-blog/blob/master/tests/acceptance/meta-test.js

unblocks https://github.com/ember-learn/guides-source/pull/1977


mansona (approved):
Looks great 🎉

PR #1985, opened 2023-12-15
ember-cli-update v4.0.1...v4.4.1

bartocc:
ember-template-lint@4+ exists with status code 1 when there are no .hbs files to check, while ember-template-lint@3 exited with status code 0.

Since this repo does not contain any .hbs, let's just remove ember-template-lint and the related lint:hbs script


mansona (approved):
Looks good 🎉

PR #1986, opened 2023-12-17
ember-cli-update v4.4.1...v4.8.1

bartocc:
Needs: #1985 
Needs: https://github.com/empress/guidemaker/pull/100


mansona:
The guidemaker PR is merged so you can rebase and update and this would be ready to review. Also you will need to add `crypto` to the fastbootDependencies because of the error in the netlify build 🙈 this is annoying but it's currently the only workaround available

PR #1991, opened 2024-01-06
feat: Use Shikiji as syntax highlighter [WIP]

IgnaceMaes:
Proof of concept in using Shikiji as syntax highlighter for the Ember guides.

PR #1994, opened 2024-01-14
ci: upgrade Node to v20

IgnaceMaes:
Extracting parts of the [Shikiji prototype](https://github.com/ember-learn/guides-source/pull/1991) in reviewable bits.

This PR upgrades Node to the latest LTS release.


.node-version
```diff
@@ -0,0 +1 @@
+20.11.0
```
IgnaceMaes, line 1: This is used to force the Node version on Netlify. I'm assuming it's currently configured in the UI itself and not from code.

mansona (approved):
Looks good to me 👍

PR #2000, opened 2024-01-29
Fix tutorial

mansona:
This PR adds a little script that I used to update the tutorial and I ran it to get the diffs for us.

Things to do before we merge this: 

- [x] either manually or automatically update the external links from each version to the correct ember/ember-data versions
- [x] generally double check this works 🤷 
- [x] any other things?


update-tutorial.mjs
```diff
@@ -0,0 +1,59 @@
...
+  'v5.3.0': 'db073816ad373993cf3505a701f26e3c9c47ac2d',
+  'v5.4.0': '3a459944b20d7070830f5ab5f3de92a443dc3883',
+  'v5.5.0': 'a7aec376ea0b77c51196540c4d21a1e39e287754',
+};
+
+for (let version in versionShas) {
+  console.log(`Fixing version ${version}`);
+  const sha = versionShas[version];
+  await execa('git', ['checkout', sha], {
+    cwd: '/Users/mansona/temp/guides-source-temp-tutorial',
```
IgnaceMaes, line 23: Should this be a user independent path? e.g. could we use `/tmp`?

update-tutorial.mjs
```diff
@@ -0,0 +1,59 @@
...
+      '-r',
+      `/Users/mansona/temp/guides-source-temp-tutorial/guides/release/tutorial`,
+      `./guides/${version}/tutorial`,
+    ],
+    { stdio: 'inherit' }
+  );
+
+  let emberDataVersion = version.replace(/^v/, '');
+
+  if (['5.4.0', '5.5.0'].includes(emberDataVersion)) {
```
IgnaceMaes, line 41: Why these specific Ember Data versions? What happens e.g. if there's a `5.6.0`?

IgnaceMaes:
Could we have a short description and example usage of the script, if it gets included in source control?

mansona:
sorry something wasn't clear with this PR 🙈 the actual script isn't intended to be used in the future, I only included it in the PR to make sure that we didn't lose the code. I've removed it now as the last commit 👍

package.json
```diff
@@ -120,5 +120,8 @@
   "volta": {
     "node": "20.11.0",
     "npm": "9.6.3"
+  },
+  "dependencies": {
+    "execa": "^8.0.1"
```
IgnaceMaes, line 125: Should this dependency be removed as well then, now that the script is gone?

mansona: done!

IgnaceMaes (approved):
Did some quit spot checks on links and LGTM.

PR #2002, opened 2024-02-01
fix: allow EmberData version to be set on release:minor script

jaredgalanis:
- prompt user to enter EmberData current version and pass it to update links script
- also handle different ember data version when updating links


mansona (approved):
Looks good 👍

PR #2010, opened 2024-02-17
Update more-about-components.md - Rental::Image component test

bitxplora:
The test for `<Rental::Image />` component failed as a result of two different invocations of the image component. The first one was not passed the attributes, so the test failed with this result. 

`Result: "Element .image img does not have attribute \"src\"" `

The fix: Removing the first component invocation, leaving the second invocation which has attributes passed in, made the test successful.


bitxplora:
Ember Discord's post on this fix

[Ember Discord Post](https://discord.com/channels/480462759797063690/1208434266883756042)

PR #2012, opened 2024-02-19
chore: normalize h1 headers in content to h2

IgnaceMaes:
As part of https://github.com/ember-learn/guidemaker-ember-template/issues/179 I took a look at our current header levels used. It turns out we have a few H1s inside of the guide contents. In general there should only be one h1 on a page, which here is the page title, so having H1s inside the content is odd.

This PR replaces the usages (found via regex search) with H2s.

### DIFF
--- a/guides/release/components/index.md
+++ b/guides/release/components/index.md
@@ -186,7 +186,7 @@ know HTML and CSS, you know how to build a basic Ember application!
 You can even use SVG or web components without any changes. As long as your HTML
 is valid, Ember will render it.
 
-# Self-Closing Tags
+## Self-Closing Tags
 
 In addition to normal HTML syntax, Ember allows you to use self-closing syntax
 (`<div />`) as a shorthand for an opening and closing tag (`<div></div>`).
@@ -206,7 +206,7 @@ In addition to normal HTML syntax, Ember allows you to use self-closing syntax
   </div>
 </div>
 
-# Supported Features
+## Supported Features
 
 This means that all of the following HTML features work as-is:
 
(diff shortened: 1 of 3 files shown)


IgnaceMaes:
@mansona Did you close this on purpose? 🤔

mansona:
ha ha no 🙃 
Screenshot 2024-02-22 at 14 47 42

This happened because I merged your other PR, and you had accidentally used the magic encantation `fix <link to pr>` and it tied that PR's fate to this one 🙃

mansona:
so... regarding this PR, I was 90% sure that we had a lint rule that made sure that we didn't have any H1 elements in our markdowns here 🤔 we should find a way to lint this

IgnaceMaes:
Ahhh, that `fix <id>` explains 😂 

I had a quick look in [the remark config](https://github.com/ember-learn/guides-source/blob/master/.remarkrc.js) but don't see any rules related to the usage of h1's. TBF I'm not that familiar with it so I also wouldn't know what the rule is called.

As a way forward we could create a bug issue to set up such kind of rule?

mansona:
So we already have a node script that checks all markdown files in CI, and we could easily add something to just check that there are no H1s here https://github.com/ember-learn/guides-source/blob/master/node-tests/link-checker.js#L29 

We could then open an issue to make sure that header nesting is correct i.e. no H4s in a H2 section etc. 

On the flip side I’m not explicitly asking to block this PR unless this CI work is done, I just know our track record and if we don’t do it now it’s unlikely to get done 😂

IgnaceMaes:
I've opened an issue as I don't have time to look into automating this in the near future.

mansona (approved):
LGTM, I'll try to take a look at the linting issue soon 👍

PR #2013, opened 2024-02-21
Correct @cached example

kategengler:
The example for `@cached` was using code that wasn't in the example. Seems like it was missed when the conversion from caching primitives to `@cached` happened in https://github.com/ember-learn/guides-source/pull/1887/files


mansona (approved):
Looks good to me 🎉

PR #2030, opened 2024-05-31
Update `finding Record` section in ember-data guides to teach request service

MehulKChaudhari:
This is to make changes to this [page](https://guides.emberjs.com/release/models/finding-records/)

Changes related to ember-data request service paradigm [RFC](https://github.com/emberjs/rfcs/pull/964)


guides/release/models/finding-records.md
```diff
@@ -38,7 +40,7 @@ Use [`store.peekAll()`](https://api.emberjs.com/ember-data/release/classes/Store
 let blogPosts = this.store.peekAll('blog-post'); // => no network request
 ```
 
-`store.findAll()` returns a `PromiseArray` that fulfills to a `RecordArray` and `store.peekAll` directly returns a `RecordArray`.
+`findRecord()` returns a `PromiseArray` that fulfills to a `RecordArray` and `store.peekAll` directly returns a `RecordArray`.
```
Baltazore, line 43: This is not really correct here. As `findRecord` is request builder and returns options for the request to make. And then later `store.request` returns Future or StructuredDocument. Please checkout documentation for `@ember-data/request`

guides/release/models/finding-records.md
```diff
@@ -47,28 +49,29 @@ the `[]` notation will not work--you'll have to use `objectAt(index)` instead.
 ### Querying for Multiple Records
 
 EmberData provides the ability to query for records that meet certain criteria.
-Calling [`store.query()`](https://api.emberjs.com/ember-data/release/classes/Store/methods/query?anchor=query) will make a `GET` request with the passed object serialized as query params.
-This method returns a `PromiseArray` in the same way as `findAll`.
+Calling [`query()`](https://api.emberjs.com/ember-data/5.3/functions/@ember-data%2Fjson-api%2Frequest/query) will make a `GET` request with the passed object serialized as query params.
+This method returns a respone from the server in the same way as `findRecord`.
```
Baltazore, line 53: Similarly here

guides/release/models/finding-records.md
```diff
@@ -47,28 +49,29 @@ the `[]` notation will not work--you'll have to use `objectAt(index)` instead.
...
+}));
+const person = result.content.data;
 ```
 
 ### Querying for A Single Record
 
-If you are using an adapter that supports server requests capable of returning a single model object,
-EmberData provides a convenience method [`store.queryRecord()`](https://api.emberjs.com/ember-data/release/classes/Store/methods/queryRecord?anchor=queryRecord) that will return a promise that resolves with that single record.
-The request is made via a method `queryRecord()` defined by the adapter.
+If you are using an builder that supports server requests capable of returning a single model object,
```
Baltazore, line 72: This whole paragraph I think should be rewritten. Something like if you need to fetch single record you should use `findRecord` builder for your request. Maybe some native speaker can write it better

guides/release/models/finding-records.md
```diff
@@ -82,43 +85,43 @@ For example, if your server API provides an endpoint for the currently logged in
...
-export default class UserAdapter extends Adapter {
-  queryRecord(store, type, query) {
-    return fetch('/api/current_user');
-  }
+And if the builders for the `User` model defines a `queryData()` method that targets that endpoint:
+
+```javascript {data-filename=app/builders/user.js}
+export function queryData() {
+    return {
+        url: `/api/current_user`,
```
Baltazore, line 93: I feel like the approach here to use custom builder is correct one, but this exact builder implementation missing some key options in return. Let me find more info for you

Baltazore (commented):
Great job overall, still have some questions to resolve.

guides/release/models/finding-records.md
```diff
@@ -2,34 +2,36 @@ The EmberData store provides an interface for retrieving records of a single typ
...
-this.store.findRecord('blog-post', 1)  // => GET /blog-posts/1
-  .then(function(blogPost) {
-      // Do something with `blogPost`
-  });
+// GET /blog-post/1
+import { service } from '@ember/service';
+import { findRecord } from '@ember-data/json-api/request';
+
+// somewhere in the app
+const result = await this.store.request(findRecord('blog-post', '1'));
```
Baltazore, line 14:
```suggestion
const result = await store.request(findRecord('blog-post', '1')); 
```

I think @runspired mentioned some time ago that we need to move from referencing `this.store` to just `store`

guides/release/models/finding-records.md
```diff
@@ -82,43 +85,44 @@ For example, if your server API provides an endpoint for the currently logged in
...
+Then, calling `queryData()` will retrieve that object from the server:
 
 ```javascript
-store.queryRecord('user', {}).then(function(user) {
-  let username = user.get('username');
-  console.log(`Currently logged in as ${username}`);
-});
+import { queryData } from './builders';
+
+const user = await this.requestManager.request(queryData());
```
Baltazore, line 107:
```suggestion
const user = await requestManager.request(queryData());
```

Baltazore (commented):
code parts looks correct, added minor change requests for consistency.

someone please review is linguistically all sounds good

guides/release/models/finding-records.md
MehulKChaudhari: Thanks for pointing out fixed it

guides/release/models/finding-records.md
MehulKChaudhari: Oh, My bad I might have missed it. Thanks for pointing it out. 

 Fixed it now.

guides/release/models/finding-records.md
```diff
@@ -2,34 +2,36 @@ The EmberData store provides an interface for retrieving records of a single typ
 
 ### Retrieving a Single Record
 
-Use [`store.findRecord()`](https://api.emberjs.com/ember-data/release/classes/Store/methods/findRecord?anchor=findRecord) to retrieve a record by its type and ID.
-This will return a promise that fulfills with the requested record:
+Use [`findRecord()`](https://api.emberjs.com/ember-data/5.3/functions/@ember-data%2Fjson-api%2Frequest/findRecord) to retrieve a record by its type and ID.
```
runspired, line 5: I think this should likely still be `release` vs `5.3` here

guides/release/models/finding-records.md
```diff
@@ -2,34 +2,36 @@ The EmberData store provides an interface for retrieving records of a single typ
...
+This will return a single `record` or a `array` depends upon the response from server:
```
runspired, line 6: we should probably introduce builders / the concept of the request first, because this function won't do this on its own and this particular one builds a json:api specific request.

guides/release/models/finding-records.md
MehulKChaudhari: Yes. Changed it.

guides/release/models/finding-records.md
MehulKChaudhari: Umm, that Makes sense. But the builder or the concept of the request will be in a separate document, right? Since this is specific to finding records.

runspired: there's a chicken/egg problem, unsure if its better in a separate file or not because its likely you want to introduce them together

PR #2038, opened 2024-06-25
Replace not existing `removeObject` method

wozny1989:
New Ember provides `EXTEND_PROTOTYPES: false`, then method like `removeObject` doesn't exist which can be confusing for new developers. Feel free to suggest how write it more pretty 🙏


IgnaceMaes (approved):
Good catch, thanks for the contribution 🎉

PR #2040, opened 2024-06-26
deps: upgrade ember-styleguide to latest patch

IgnaceMaes:
Upgrades to the latest `ember-styleguide` patch, and removes the temporarily CSS override which now comes from the package.


mansona (approved):
LGTM

PR #2053, opened 2024-08-26
add heading about configuring netlify URL handlers

nikomatsakis:
make it easier to see that configuration is required.

I missed this paragraph even when I was directly looking for it.


mansona (approved):
I'm going to merge this but I think we should review the "headings" under the Netlify section 🙈 it's better if they are real heading for lots of different reasons, but this change is at least an improvement 👍

PR #2062, opened 2024-10-09
Remove discussion of old debug handler polyfill, it's been a minute

acorncom:
As part of discussing https://github.com/emberjs/rfcs/pull/1009, I took another look at the deprecation workflows page. I don't think we need to recommend installing the pre 2.1 deprecation workflow polyfill any more ... 😉 

Not least because it's archived and likely won't even working in the last few versions of Ember: https://github.com/ember-polyfills/ember-debug-handlers-polyfill


NullVoxPopuli (approved):

mansona (approved):
Looks good 👍

PR #2069, opened 2024-10-30
Template Tag Feature Toggle

ef4:
Starting work on being able to swap content between hbs and gjs.


guides/release/getting-started/quick-start.md
```diff
@@ -83,6 +83,8 @@ Congratulations! You just created and booted your first Ember app.
 
 ## Write some HTML in a template
 
+<div data-template-tag="off" markdown="1">
```
ef4, line 86: `markdown="1"` is required here to get `showdown` to continue parsing inside this tag. Without it you lose all support for formatting, etc.

PR #2071, opened 2024-11-06
Switch to pnpm

ef4:
ember.js, ember-cli, embroider, etc are all using pnpm now. It's the last bad NPM client, and having it here will help will upcoming updates.


package.json
```diff
@@ -41,6 +41,9 @@
   "devDependencies": {
     "@ember/optional-features": "^2.0.0",
     "@ember/test-helpers": "^2.7.0",
+    "@ember-data/adapter": "^4.4.3",
+    "@ember-data/model": "^4.4.3",
+    "@ember-data/serializer": "^4.4.3",
```
ef4, line 46: These are peerDependencies of guidemaker that were previously being sloppily provided by ember-data's transitive dependencies. pnpm will more strictly do what you say, and by not having these we were letting guidemaker's "*" dependencies auto-install the very latest `@ember-data/*` packages instead, which don't work.

.github/workflows/ci.yml
```diff
@@ -58,19 +39,10 @@ jobs:
   test-node:
     name: Test node-tests
     runs-on: ubuntu-latest
-    timeout-minutes: 7
+    timeout-minutes: 14
```
ef4, line 42: Hang on, there's an extreme performance regression here. Apparently the node:tests on main take only 6 seconds, and on this branch I'm not even sure this 14 minutes is long enough.

ef4: (I don't blame pnpm, I blame the fact that this re-rolled the whole lockfile so we got some upgrade that pwned performance.)

ef4:
This is ready to go. 

I'm pretty sure there is a pre-existing flaky test in `Acceptance | side bar links: release links go to correct page` but my attempts to observe it made it go away again.

kategengler (approved):

PR #2092, opened 2025-02-04
Fix improperly-applied diff highlighting

GeneralSpace:
This fixes several diffs which were marked as additions when they should have been deletions (or vice versa).

### DIFF
--- a/guides/release/tutorial/part-1/component-basics.md
+++ b/guides/release/tutorial/part-1/component-basics.md
@@ -58,7 +58,7 @@ When invoking a component, Ember will replace the component tag with the content
 
 Let's try it out by editing the index template:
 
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
Got a screenshot of the before and after for this area of the docs?

GeneralSpace:
> Got a screenshot of the before and after for this area of the docs?

Sure! Here's one of the occurrences that are fixed by this PR. Note the diff highlighting applied to the last two HTML tags in each.

old #1

And here's the fixed version.

new #1

The before/after for the other occurrences are here.
old #2
new #2
old #3
new #3

lem0n4id:
+1
Noticed the improperly applied diff highlighting while going through the docs

PR #2102, opened 2025-03-14
Remove old code line from component-state-and-actions.md

yagni:
Looks like this line was leftover from the previous example; leaving it in prevents the new line from ever being called.


IgnaceMaes:
The highlights indicate this line should be removed. So this looks to be correct.

image

https://guides.emberjs.com/release/components/component-state-and-actions/#toc_combining-arguments-and-state

PR #2105, opened 2025-03-27
Fix self-closing tag syntax in Profile component test

wozny1989:
The only change made was adding a closing `/>` to properly terminate the self-closing <Profile> component tag.


locks:
Appreciate!

PR #2108, opened 2025-04-25
Remove outdated / incorrect information on the initializers page

NullVoxPopuli:
implicit injections were deprecated long ago

### DIFF
--- a/guides/release/applications/initializers.md
+++ b/guides/release/applications/initializers.md
@@ -18,47 +18,6 @@ Any asynchronous loading conditions (e.g. user authorization) are almost always
 better handled in your application route's hooks,
 which allows for DOM interaction while waiting for conditions to resolve.
 
-## Application Initializers
-
-Application initializers can be created with Ember CLI's `initializer` generator:
-
-```bash
-ember generate initializer shopping-cart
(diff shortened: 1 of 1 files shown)


johanrd:
@NullVoxPopuli the implicit injection part of this is outdated, perhaps, but are `Application Initializers` and `Application Instance Initializers` themselves deprecated? 

I'm asking since I'm reporting [ember deprecations to @sentry/ember](https://github.com/getsentry/sentry-javascript/issues?q=is%3Aissue%20state%3Aopen%20ember), and there I also see they are using an [instance initializer for sentry-performance](https://github.com/getsentry/sentry-javascript/blob/e0400020d9acddc4b9291d88006edf1c77311bc0/packages/ember/addon/instance-initializers/sentry-performance.ts).

Thanks,

MehulKChaudhari (approved):

IgnaceMaes:
While this does remove the outdated service injections, it makes the page incomplete about initializers itself. We could replace the injection bit with a different setup logic as example instead?

NullVoxPopuli:
sure, but I don't have the energy to write that rn. <3

PR #2116, opened 2025-05-30
Merge 'origin/master' into gjs branch

mansona:


mansona:
Actually it looks like nothing has been merged into the GJS branch yet so I'm just going to reset it so there is a simpler history 👍

PR #2123, opened 2025-07-11
Update template-tag-format.md to include info about the JS representation of the runtime compiler

NullVoxPopuli:


guides/release/components/template-tag-format.md
```diff
@@ -259,6 +259,18 @@ export default CustomSelect;
 
 This can be a powerful refactoring technique to break up large components into smaller ones. (where it makes sense!)
 
+## Runtime compiler
```
locks: I think this section is missing some explanation for context?

NullVoxPopuli: aye, we need to copy a bunch of stuff from here: https://github.com/emberjs/rfcs/pull/931

guides/release/components/template-tag-format.md
```diff
@@ -259,6 +259,54 @@ export default CustomSelect;
...
+```
+
+If in an environment with compilation, omitting the `/runtime` at the end of the import allows ahead-of-time compilation to occur on components created with `template()` for better runtime performance.
+
+Without specifying `/runtime`, there are additional restrictions required for the argument passed to `template()`:
+- it must be a string literal
+
+With the `/runtime`, the argument passed to `template()` can be an extpression, for example:
+```js
+/* someValue could be f rom anywhere */
```
NullVoxPopuli:
```suggestion
/* someValue could be from anywhere */ 
```

guides/release/components/template-tag-format.md
```diff
@@ -259,6 +259,54 @@ export default CustomSelect;
...
+
+}
+```
+
+If in an environment with compilation, omitting the `/runtime` at the end of the import allows ahead-of-time compilation to occur on components created with `template()` for better runtime performance.
+
+Without specifying `/runtime`, there are additional restrictions required for the argument passed to `template()`:
+- it must be a string literal
+
+With the `/runtime`, the argument passed to `template()` can be an expression, for example:
```
sukima: When you say _expression_ does that also include support for interpolation?

```js
export default template`<div>${someValue}</div>`;
```

guides/release/components/template-tag-format.md
```diff
@@ -259,6 +259,54 @@ export default CustomSelect;
...
+```js
+/* someValue could be from anywhere */ 
+export default default template(someValue);
+```
+
```
sukima, line 371: I think it would help (**me**) to see example use cases for the different possibilities. Maybe when to precompile during a `<script>` tag in the HTML `<head>` versus dynamically from a `<textarea>`. The difference between `scope:` and `${stringInterpolation}`. Use of this in a build tool versus use of this in an HTML file. And how this might look used in a QUnit test or Vitest.

guides/release/components/template-tag-format.md
NullVoxPopuli: nay -- interpolation is a hazard because it can't be reactive as far as I know -- ideally we'd even add a lint that prevents that.

if reactivity is not needed, then I don't see a reason why this wouldn't work (I expect it to work -- just not be reactive on changes to someValue)
```js
export default template(`<div>${someValue}</div>`);
```

ef4: This is not quite correct, even reactivity should in theory work, if for example you put the `template` call in a getter and access reactive state. That's probably a fairly expensive thing to do, but would be possible.

PR #2146, opened 2025-09-13
minor changes to runloop

mansona:
There was only one template in here but there are 2 other changes that I made to this file. I am targeting gjs to prevent confusion 👍 

Changes: 

- make the warning at the top a "Zoey Says" to make it more noticeable
- change the one hbs template to be gjs and wrap it in a `<template/>` (this is probably overkill but it's also just fine) 
- stop using the Ember.run global in the JS examples

### DIFF
--- a/guides/release/applications/run-loop.md
+++ b/guides/release/applications/run-loop.md
@@ -1,6 +1,22 @@
-**Note:**
-* _For basic Ember app development scenarios, you don't need to understand the run loop or use it directly. All common paths are paved nicely for you and don't require working with the run loop._
-* _However, the run loop will be helpful to understand the internals of Ember and to assist in customized performance tuning by manually batching costly work._
+<div class="cta">
+  <div class="cta-note">
+    <div class="cta-note-body">
+      <div class="cta-note-heading" data-test-es-note-heading="">Zoey says...</div>
+      <div class="cta-note-message">
+      
+        <p>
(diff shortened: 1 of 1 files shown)


guides/release/applications/run-loop.md
```diff
@@ -83,9 +99,11 @@ class Image {
 
 and a template to display its attributes:
 
-```handlebars
-{{this.width}}
-{{this.aspectRatio}}
+```gjs
+<template>
+  {{this.width}}
```
kategengler: This example was weird to begin with -- the Image class isn't a component or controller that could back a template.

guides/release/applications/run-loop.md
```diff
@@ -174,8 +192,10 @@ In this example, `Ember.run` is used to handle an online
 event (browser gains internet access) and run some Ember code.
 
 ```javascript
+import { run } from '@ember/runloop';
```
kategengler, line 197: A few lines above references `Ember.run`

guides/release/applications/run-loop.md
mansona: I fixed it in a way that will allow us to move forward for now 👍 I think this whole page is questionable but I don't want it to be a blocker of merging the GJS branch

guides/release/applications/run-loop.md
mansona: got them thanks 👍

IgnaceMaes (approved):

PR #2175, opened 2025-10-23
use gjs in Linking Between Routes guide

mcfiredrill:
Changed the templates from .hbs to .gjs, and added relevant imports.

I didn't add the import or <template> when it looked like the example was only showing part of a template.

### DIFF
--- a/guides/release/routing/linking-between-routes.md
+++ b/guides/release/routing/linking-between-routes.md
@@ -193,12 +193,19 @@ the currently active one, you can pass an array of model objects using the
 
 ```gjs {data-filename=app/templates/photos.gjs}
 import { LinkTo } from '@ember/routing';
+import { array } from '@ember/helper';
+
+export default class {
+  get latestComments() {
+    return this.model.sort((a, b) => new Date(b.date) - new Date(a.date)).slice(0, 5);
+  }
+}
 
 <template>
   <h1>Latest Comments</h1>
 
   <ul>
-    {{#each @controller.latestComments as |comment|}}
+    {{#each this.latestComments as |comment|}}
       <li>
         <LinkTo @route="photos.photo.comment" @models={{array comment.photo comment}}>
           {{excerpt comment.body}}...


guides/release/routing/linking-between-routes.md
```diff
@@ -182,18 +191,21 @@ To solve this problem, or maybe to cross-link comments from photos other than
...
-    </li>
-  {{/each}}
-</ul>
+```handlebars {data-filename=app/templates/photos.gjs}
+import { LinkTo } from '@ember/routing';
+<template>
+  <h1>Latest Comments</h1>
+  
+  <ul>
+    {{#each this.latestComments as |comment|}}
```
SanderKnauff: Does `this.latestComments` still work in route templates?
Shouldn't this be `@controller.latestComments`?

guides/release/routing/linking-between-routes.md
```diff
@@ -16,14 +16,17 @@ Router.map(function() {
...
 
-```handlebars {data-filename=app/templates/photos.hbs}
-<ul>
-  {{#each this.photos as |p|}}
-    <li>
-      <LinkTo @route="photos.edit" @model={{p}}>{{p.title}}</LinkTo>
-    </li>
-  {{/each}}
-</ul>
+```handlebars {data-filename=app/templates/photos.gjs}
```
mansona: this needs to be gjs in all of your examples 👍 
```suggestion
```gjs {data-filename=app/templates/photos.gjs}
```

mansona (changes requested):
thanks for this! can you update the language on the code blocks that you changed too?

guides/release/routing/linking-between-routes.md
mansona: very good point 👍 yes that would need to be `@controller`

SanderKnauff: In that case, it probably applies to the other references to `this` as well so those also should be updated

tcjr: Yes, I was just going to suggest that you look at all the `this.` references in the newly converted gjs files.  Some may come from the model and some from a controller.

Additionally, you will need to add import statements for _everything_ you use in the template, like `array` and `excerpt` used below.

mansona: well just to be clear here, until you explicitly added the file name this template wasn't technically wrong. You had "plausable denyability" that you were dealing with a component or something (although there was no backing class so the argument is very weak). 

If you want to be 100% correct there should be a backing class that shows some `latestComments` and you can see what the template is referring to without having to guess that there is a controller that happens to have the `latestComments` field on it. 

I hope all that makes sense

guides/release/routing/linking-between-routes.md
mcfiredrill: Thanks, I updated the language on all the code blocks

guides/release/routing/linking-between-routes.md
mcfiredrill: Thanks for the feedback! I think I know what to do, I'm working on addressing this now then I'll ask for more feedback.

mcfiredrill: Would it be better to write a backing class in the template or assume that latestComments came from a controller and just use `@controller.latestComments` ?

NullVoxPopuli: I hate controllers, so I'm a fan of using the stateful component pattern instead of controller lol

mcfiredrill: Yeah I guess we don't necessarily need to indrectly encourage controller use here

mcfiredrill: Added a backing class version, although we had some discussion about longer term direction in the discord. https://discord.com/channels/480462759797063690/480777444203429888/1432054589661778003

mcfiredrill:
@mansona done!

PR #2177, opened 2025-11-03
Update @cached decorator documentation link

acorncom:
Noticed the `@cached` link here was bad and we had some outdated info here

Updated the link for the @cached decorator to the official Ember API documentation and removed outdated information about the ember-cached-decorator-polyfill.


NullVoxPopuli:
Tyty

PR #2193, opened 2026-02-17
Add context7.json with URL and public key

NullVoxPopuli:
So we can get stats on https://context7.com/ember-learn/guides-source/


NullVoxPopuli:
AI is good enough now where we don't need context7 to exist, I think

PR #2208, opened 2026-04-14
Update Glint information in code editors guide

evoactivity:
Now part of the extension pack so removed Zoe's note. Also updated description as it was not accurate.


IgnaceMaes (changes requested):
Nice one!

Can you update in latest, instead of a specific version?

Otherwise this change will get lost and wont propagate to future version "forks".

PR #2216, opened 2026-06-09
Updating template tag page

ef4:
Updating the template tag page to reflect that it became default at ember 6.8.


kategengler (approved):

kategengler:
Preview looks good but is not reporting to github.

PR #2219, opened 2026-06-11
Reactivity

NullVoxPopuli:
To go out with / before ember-source 7.3

Advancement PR: 
- https://github.com/emberjs/rfcs/pull/1202

Merged RFC PR:
- https://github.com/emberjs/rfcs/pull/1071

Docs Inspirations:
- https://github.com/NullVoxPopuli/ember-resources/tree/main/docs
- https://github.com/NullVoxPopuli/limber/tree/main/apps/tutorial/public/docs
- https://starbeam-lake.vercel.app/start/introduction/
  - https://starbeam-lake.vercel.app/reference/reactive-primitives/
  - https://starbeam-lake.vercel.app/frameworks/ember/
- https://signalium.dev/core/signals-and-reactive-functions
  - https://signalium.dev/advanced/reactive-function-behaviors 

Read Order:

image
image


guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,406 @@
...
+state. A getter that nothing currently displays costs nothing, no matter how
+often its inputs change. Ten getters reading the same tracked property add no
+overhead to writes. Work happens at read time, driven by what the page
+actually needs.
+
+One consequence is that you should never rely on when, or whether, a getter
+runs. A derivation may run once, many times, or never; it may run later than
+you expect or more often than you expect. If you find yourself wanting "run
+this code when X changes," you are looking for something other than derived
+state - see [Reactivity and the Outside World](../outside-world/).
```
johanrd: First impressions while I read:  

The "outside world" bag/layer feels a bit forced to me. 

Maybe just simplify/unabstract to "Inputs and outputs" instead of "Outside world"?

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,406 @@
...
+universal across reactive systems, and Ember enforces it: writing to a
+tracked value that has already been read during the current render throws a
+development-mode error:
+
+```text
+Error: You attempted to update `count`, but it had already been used
+previously in the same computation.
+```
+
+This is sometimes called the _backtracking assertion_: render evaluates your
```
johanrd:
```diff
-This is sometimes called the _backtracking assertion_: render evaluates your …
+This backtracking assertion evaluates your …
```

guides/release/in-depth-topics/reactivity/index.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Where to Go from Here
+
+The rest of this section works through each layer of the model:
+
+- [Root State](./root-state/) - what should (and should not) be root state,
+  and how to design it.
+- [Derived State](./derived-state/) - laziness, purity, caching, composition,
+  and deferring consumption.
+- [Reactivity and the Outside World](./outside-world/) - outputs, side
```
johanrd: testing the consequence of the idea above:
```diff
-[Reactivity and the Outside World](./outside-world/) - outputs, side effects, async, and the edges of the graph.
+[Inputs and outputs](./inputs-and-outputs/) - inputs, outputs, side effects, async, and the edges of the graph.
```

guides/release/in-depth-topics/reactivity/root-state.md
```diff
@@ -0,0 +1,459 @@
...
+```js
+this.count = this.count; // still invalidates everything that consumed `count`
+```
+
+Consumers re-evaluate, and the renderer re-checks the DOM it produced. The
+DOM itself won't change if the final values are equal, but the recomputation
+happens.
+
+When a write that changes nothing shouldn't dirty anything, the decorator
+accepts an options form, `@tracked({ equals })`, that skips invalidation
```
johanrd, line 113: Very cool! 

Depends on release of https://github.com/emberjs/ember.js/pull/21471, right? starting with 7.3.0?

guides/release/in-depth-topics/reactivity/root-state.md
```diff
@@ -0,0 +1,459 @@
...
+  @tracked({ equals: (a, b) => a === b }) score = 0;
+
+  reset = () => {
+    this.score = 0; // if score was already 0, nothing invalidates
+  };
+}
+```
+
+Without options, `@tracked` keeps its historical always-dirty behavior, so
+existing code is unaffected.
```
johanrd: this section is temporal, which can feel off in normal prose og guides. Maybe it can be simplified to:

"Without options, `@tracked` has always-dirty behavior."

guides/release/in-depth-topics/reactivity/root-state.md
```diff
@@ -0,0 +1,459 @@
...
+```js
+import { tracked } from '@glimmer/tracking';
+
+const count = tracked(0);
+
+count.value;     // reading consumes, like any tracked property
+count.value = 1; // writing invalidates consumers
+```
+
+Unlike the decorator, a standalone value checks equality by default (using
```
johanrd, line 380: oh:/ missed this part of the RFC. The difference between `@tracked count = 0` and `const count = tracked(0)` is definitely a potential foot gun. 

maybe `root-state.md:369-373` needs a callout with them side by side or something?

johanrd: Extra need to spell it out in the guides is that this part of the RFC read in isolation can be misunderstood:
> The above example, and **the default value**, would use the "always dirty" behavior of `() => false`.

even if the drawbacks section clarifies the intent

johanrd:
@NullVoxPopuli this was a great read – thanks! Also very exited to learn about the overloaded tracked of 7.3.

guides/release/in-depth-topics/reactivity/root-state.md
NullVoxPopuli: yea, tho I think these docs are required to release in 7.3

@kategengler knows the task<->dependency chain.

guides/release/in-depth-topics/reactivity/derived-state.md
NullVoxPopuli: tyty

guides/release/in-depth-topics/reactivity/index.md
NullVoxPopuli: tyty

guides/release/in-depth-topics/reactivity/root-state.md
NullVoxPopuli: tyty

guides/release/in-depth-topics/reactivity/root-state.md
NullVoxPopuli: tyty

guides/release/in-depth-topics/reactivity/root-state.md
```diff
@@ -0,0 +1,488 @@
...
+
+class Draft {
+  @tracked title = '';
+  @tracked body = '';
+}
+```
+
+A write to a tracked property is the only way anything changes in a reactive
+application. Every update you see on screen traces back to some event
+handler, timer, or response callback assigning to root state.
```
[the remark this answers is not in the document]
NullVoxPopuli: @copilot -- you misunderstand -- these are all the same concept

guides/release/in-depth-topics/reactivity/root-state.md
```diff
@@ -0,0 +1,488 @@
...
+  associateDestroyableChild,
+  registerDestructor,
+} from '@ember/destroyable';
+
+class Poller {
+  @tracked lastReading = null;
+
+  constructor() {
+    let timer = setInterval(() => this.poll(), 5000);
+    registerDestructor(this, () => clearInterval(timer));
```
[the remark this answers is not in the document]
NullVoxPopuli: no it cannot. constructors are not normal methods that can run multiple times.

guides/release/in-depth-topics/reactivity/root-state.md
```diff
@@ -0,0 +1,488 @@
...
+have it accept functions rather than values, so that it reads the current
+state on every use instead of a snapshot from construction time. See
+[Deferring Consumption](../derived-state/#toc_deferring-consumption) for the
+full pattern.
+
+## Reactive Values Without Classes
+
+`tracked` also works as a plain function. Called with a value instead of
+applied as a decorator, it returns a standalone reactive value - root state
+that isn't attached to any class:
```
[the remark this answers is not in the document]
NullVoxPopuli: while true, I'm not sure how beneficial it is to mention in these docs.

This is implementation caveats that are very edge casey.

guides/release/in-depth-topics/reactivity/root-state.md
```diff
@@ -0,0 +1,488 @@
...
+
+  addItem(item) {
+    this.items.push(item); // 🛑 not tracked - nothing updates
+  }
+}
+```
+
+The right tool here is a tracked collection from
+[`@ember/reactive/collections`](https://api.emberjs.com/ember/release/modules/@ember%2Freactive%2Fcollections),
+which tracks reads and writes of its contents at fine granularity:
```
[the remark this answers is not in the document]
NullVoxPopuli: simpler is better, sure

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,406 @@
...
+
+Beyond raw cost, there are two more good reasons to reach for `@cached`:
+
+- Stable identity. An uncached getter that returns a fresh array or object on
+  every read can defeat downstream `===` checks and cause child components to
+  see "new" values that are deep-equal to the old ones. Caching makes the
+  derivation return the same object until its inputs actually change.
+- Once-per-change semantics. If a derivation must observably run at most once
+  per change (because it allocates, logs, or is just very hot), `@cached`
+  guarantees that.
```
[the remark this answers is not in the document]
NullVoxPopuli: 🤷 I don't know how to best phrase this. sounds good

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,373 @@
+Derived state is everything derived from [root state](../root-state/).
```
rtablada, line 1:
```suggestion
Derived state is any state that can be computed from the [root state](../root-state/).
```

This is self definition. Especially for ESL saying `derived` is defined by it being derived.

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,373 @@
...
+In a healthy Ember application, most and nearly all of your state is derived.
+Derived state requires no special API -- any usage of an ordinary getter, an ordinary function, or an ordinary template expression that reads tracked state is derived state.
+Derived state stays up to date on its own.
```
rtablada, line 4:
```suggestion
Derived state in Ember applications can be declared as native JavaScript getters, arguments or references in functions, or by using properties or values in template expressions.
Ember automatically tracks these pieces of state and keeps them up to date.
```

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,373 @@
...
+
+```js
```
rtablada, line 6:
```suggestion
Below is an example of using a native JavaScript property getter [get](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get).

```js
```

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,373 @@
...
+
+  get summary() {
+    return this.hasQuery
+      ? `${this.visibleResults.length} results for “${this.query}”`
+      : 'Type to search';
+  }
+}
+```
+
+The getters are solely native JavaScript, [get](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get).
```
rtablada, line 29:
```suggestion
```

Remove and move above as suggested

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,373 @@
...
+
+Because `summary` reads from `hasQuery` and `visibleResults`, and because those eventually access root state, any `{{ }}` regions in the template that access summary will _become entangled with_ or _consume_ `query` and `results`, and updates will happen automatically.
```
rtablada, line 31: This sentence is pretty confusing to track. Start with answering what or why people care.
What is the problem that Ember's auto-reactivity solves, why is this hard?

Maybe start with something like

```
Even though `summary` does not directly reference `query` or `results`, Ember is able to setup reactivity...
```

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,373 @@
...
+
+## Derivations Are Lazy
+
+The most important fact about derived state in Ember is that changing root state does not run your getters.
+A write to `@tracked` state only marks the things that consumed it as out of date.
+The getter runs again when something reads it, and only then.
+If nothing reads it, it never runs.
```
rtablada, line 38:
```suggestion
In Ember, changing root state does not directly cause getters to re-run.
Changing a property that is decorated with `@tracked` only marks the things that consumed it as out of date.
Using a getter in a template or other reactive system means that those systems listen for this "out of date" mark and then rerun the getter.
Since reactive getters are plain JavaScript properties, they can be used outside of the reactivity system without side-effects (this is referred to as "pull-based" reactivity) in [Thinking in Reactivity](../).

Below is an example of pull based reactivity where no getters are automatically reassessed.
```

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,373 @@
...
+
+search.query = 'Hello';
+search.query = 'Hello, world';
+search.query = 'Hello, world!';
+// ...nothing is logged. No computation has happened at all.
+
+search.normalizedQuery; // logs "computing!" - exactly once
+```
+
+This is pull-based reactivity, described in [Thinking in Reactivity](../).
```
rtablada, line 60:
```suggestion
```

Moves this to above the example

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,373 @@
...
+It is why you can be generous with derived state.
+A getter that nothing currently displays costs nothing, no matter how often its inputs change.
+Ten getters that read the same tracked property add no overhead to writes.
+Work happens at read time, driven by what the page needs.
+
+One consequence is that you cannot rely on when, or whether, a getter runs.
+A derivation can run once, many times, or never.
+It can run later than you expect, or more often than you expect.
+If you want to "run this code when X changes," you are looking for something other than derived state.
+See [Inputs and Outputs](../inputs-and-outputs/).
```
rtablada, line 70:
```suggestion
Unlike pushed-based reactivity, this means that you cannot rely on when, or whether, a getter runs.
It can run later than you expect, or more often than you expect.
Do not use derived state to trigger side effects ("run this code when X changes"), if you do need these type of side effects see [Inputs and Outputs](../inputs-and-outputs/).
```

rtablada (changes requested):

guides/release/in-depth-topics/reactivity/derived-state.md
NullVoxPopuli: computed is quite overloaded tho -- and in ember specifically means.... `@computed` 🙈

rtablada: ~calculated~ See below

rtablada: Actually on second thought... `computed` is the right term.

I don't think we should stray for simplicity and readability because of Ember 1.x pattern terminology.

`@computed` is marked in lints or deprecations and shouldn't be mentioned as part of learning new modern Ember, so confusion with this deprecated (and usually not necessary decorator) shouldn't tarnish the simplest/clearest language here.

guides/release/in-depth-topics/reactivity/derived-state.md
```diff
@@ -0,0 +1,373 @@
...
+
+## Derivations Must Be Pure
+
+The job of a derivation is to compute a value from its inputs.
+It must not change anything, and above all, it must not write to tracked state.
+This rule is universal across reactive systems, and Ember enforces it.
+A write to a tracked value that was already read during the current render throws a development-mode error:
```
rtablada, line 77: Be clearer: Derived state should not introduce side-effects or directly trigger changes in root-state

I would maybe show using a template function to do subscribe to pull reactivity? 

There are times with metrics or similar cases where people will want to do some observer like behavior on a computed state reacting to changes to trigger an async side-effect (and clearly mention that subscribing _is_ async).

---

From teaching, having a "don't do this thing" without clear information of how to do it when you actually needs to, often leads to unnecessary fear and then doing the wrong pattern anyway.

PR #2231, opened 2026-07-27
Fix Quick Start diff description and restore +/- markers (#2090)

### LINKED ISSUE #2090: Description of diff style is incorrect in Quick Start
https://guides.emberjs.com/release/getting-started/quick-start/  [here in code] (https://github.com/ember-learn/guides-source/blob/master/guides/release/getting-started/quick-start.md?plain=1#L257 ) describes the diff style: 

> In the rest of the code examples in this tutorial, whenever we add or remove code, we will show a "diff." The lines you need to remove have a minus sign in front of them, and the lines you should add have a plus sign. If you are using a screen reader while you go through the Guides, we recommend using Firefox and NVDA or Safari and VoiceOver for the best experience.

However, the diffs display only with colors: 

Image

There are no pluses or minuses in the elements either. 

I suspect this may not be accessible to those with colorblindness.

olenderhub:
Fixes #2090

After the Shiki migration, code diffs were highlighted with color only. The Quick Start guide still said added/removed lines would show `+` / `-`, but those markers were gone because the old Prism CSS no longer matched Shiki’s classes.

This PR restores `+` / `-` next to line numbers on all Shiki diff blocks (using the existing Shiki add/remove colors), keeps a fixed-width gutter so indentation stays aligned, and updates the Quick Start wording to match.

**Text change** (Quick Start only):
https://guides.emberjs.com/release/getting-started/quick-start/

**Visual `+`/`-` markers** (every page with diffs, including Quick Start), e.g.:
- https://guides.emberjs.com/release/getting-started/quick-start/
- https://guides.emberjs.com/release/tutorial/part-1/component-basics/
- https://guides.emberjs.com/release/tutorial/part-1/interactive-components/
- https://guides.emberjs.com/release/components/introducing-components/
- https://guides.emberjs.com/release/components/component-state-and-actions/
- https://guides.emberjs.com/release/tutorial/part-1/working-with-data/

The same CSS also applies under other guide versions that use Shiki diffs (e.g. `/v7.0.0/...`), since it ships in shared `/assets/`.

### DIFF
--- a/app/styles/app.css
+++ b/app/styles/app.css
@@ -113,40 +113,35 @@ code {
   z-index: 10;
 }
 
-/* Fix diff formatting */
+/*
+  Diff formatting for Shiki (ember-showdown-shiki).
+  Show +/- beside line numbers so diffs are not color-only (#2090).
+  Colors match .shiki .line.add / .line.remove in ember-showdown-shiki.
 
-code[data-diff] .diff-deletion {
-  background-color: rgb(144 84 84 / 70%);
+  Use a fixed-width gutter for every line in a diff block so code stays
+  aligned whether the line is added, removed, or unchanged.
+*/
+pre.has-diff code.line-numbers .line::before {
+  width: 5ch;
+  margin-right: 1rem;
+  text-align: right;
+  font-variant-numeric: tabular-nums;
 }
 
-code[data-diff] .diff-insertion {
-  background-color: rgb(93 125 93 / 50%);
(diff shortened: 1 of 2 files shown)


kategengler:
Thanks for doing this! The styling here doesn't look right when the line numbers increase in digits:

CleanShot 2026-07-27 at 13 32 26@2x

olenderhub:
@kategengler thank you for reply

I think better proposition about is https://github.com/ember-learn/guides-source/issues/2090#issuecomment-5094811828

PR #2232, opened 2026-08-11
remove imports that aren't needed on ember >= 7.1

ef4:
- remove imports that are built-in starting at ember 7.1
- cleanup some vestigial use of "component's template"
- cleanup a remaining use of hbs-backtick instead of template tag


kategengler (approved):

ef4:
(force pushed to remove accidental edit of tutorial output)

PR #2237, opened 2026-09-10
Application State

NullVoxPopuli:
This is necessarily entangled with 
- https://github.com/ember-learn/guides-source/pull/2219

image

NOTE: these things make dealing with state easier than ember's docs currently describe (and remove some boilerplate in this PR):
- https://github.com/universal-ember/reactiveweb/blob/9ef52e6ac15f72e9fe260a53dc68f6cf74cd32cd/reactiveweb/src/link.ts#L95
  - https://github.com/emberjs/rfcs/issues/905
  - https://github.com/emberjs/rfcs/pull/1067 
- https://github.com/emberjs/rfcs/pull/502
  - Note that this RFC I think could be closed, because everything described in this "Application State" guides PR gets us everything we want from explicit services, all without using ember's services. The only action item I think for ember.js here would be to potentially overload teh `@service` decorator to give us the behavior from the below-linked code
- https://github.com/universal-ember/ember-primitives/blob/8cbb7dbb7a23d14e9163ed99a11380247eb6d28b/ember-primitives/src/service.ts#L42  
- https://github.com/universal-ember/ember-primitives/blob/8cbb7dbb7a23d14e9163ed99a11380247eb6d28b/ember-primitives/src/store.ts#L49
  
  
I can, ofc, pursue shipping the RFCs above, but I would like more engagement from the core teams, and the RFCs will ofc need to be accepted


guides/release/application-state/services.md
NullVoxPopuli, line 1: I'm sad this doesn't show up as a `mv` :(

guides/release/application-state/application-wide-state.md
```diff
@@ -0,0 +1,189 @@
...
+The signed-in user, the contents of a shopping cart, and an open WebSocket connection are examples.
+Every component that shows or changes this state needs the same instance.
+
+The application instance, called the owner, has the longest lifetime in an Ember application.
+State tied to the owner is created once, shared everywhere, and destroyed when the application is destroyed.
+In tests, that happens at the end of every test.
+
+Ember gives you two ways to tie state to the owner: a plain class, described on this page, and a [Service](../services/).
+Prefer the plain class because plain classes will participate in the import graph and naturally fit in to automatic bundle splitting (unlike services).
+Use a Service only for the small, minimally required, set of state that the application needs to boot.
```
johanrd, line 11: Thanks for the read (again). This section is very interesting to me – As long time ember user, I have never considered plain es module class state as a good alternative to Services (except for a few cases of caches I use). This also begs the question for me in this read: _Why should I not replace all services with plain classes?_ 

What is special about services that makes them better fit for "the state that the application needs to boot.", than plain classes? (Isn't there a way to make es class es available on app boot?).

NullVoxPopuli: > why should I not replace all services with plain classes

if they are not needed for boot, I argue you absolutely should use plain classes. The closer we get to vanilla javascript the better, and more at-home developers will feel.

> What is special about services that makes them better fit for "the state that the application needs to boot.", than plain classes? (Isn't there a way to make es class es available on app boot?).

if your vanilla class adheres to the same static api interface, you don't need to use `Service` from `@ember/service` at all, your intuition is correct here. (all you need is a static create function).

So in that regard, there is _almost_ an argument  for ditching `Service` entirely -- however the convenience of the interface abstraction is nice.

I drew the line at "minimally required state", such as the router service, because:
- existing convention (we don't really have a convention for application state outside of this)
- libraries may want to access the service, and given the worries brought up in the explicit service injection RFC, I think we can _just teach_ to use the "string-service" system if someone can't get a hold of their dependency graph, or if they truely have no way to get a hold of a static reference (that isn't a string)

johanrd: thanks for the explanation. makes sense. I like that it shows a future of simple vanilla javascript, where Ember is not shy to get out of the way and embraces the platform. 

There are some teaching downsides in explaining two concepts for application state, though – at least at the same level. The proposed text for the class based approach jumps straight into  `Owner`, `singleton`, (and `WeakMap`), which may quite a steep learning curve?

We should maybe also run the thought: How should plain class state be taught when / if `Service` ever become more plain (i.e. keep most of the ergonimics, but get the benefits of cutting the cord of string based DI, etc.)

PR #2236, opened 2026-09-10
Documented how to enable syntax highlighting and language server protocol in Sublime Text

ijlee2:
## Background

I created [LSP-ember](https://github.com/ijlee2/LSP-ember) to help write code in Sublime Text. The package depends on Glint v2 and the packages for syntax highlighting, [GJS and GTS](https://github.com/ijlee2/sublime-syntax-definition-template-tag).

## What changed?

I documented how to enable syntax highlighting and language server protocol for Sublime Text. The instructions are valid once [package_control_channel#9559](https://github.com/sublimehq/package_control_channel/pull/9559#issuecomment-5615075897) and [repository#173](https://github.com/sublimelsp/repository/pull/173) are approved and merged (i.e. once the packages become available on [Package Control](https://packagecontrol.io)).

I removed [ember-cli-sublime-snippets](https://github.com/terminalvelocity/ember-cli-sublime-snippets), because it shows code patterns from Ember 2.x (removed in 2.16 in the Guides). I also removed [ember-component-template-split-view](https://github.com/mmitchellgarcia/ember-component-template-split-view), because the repo no longer exists.

PR #2238, opened 2026-09-11
Patched #2236

ijlee2:
## Background

After #2236 was merged, I received and implemented the review suggestion https://github.com/sublimehq/package_control_channel/pull/9559#issuecomment-5622237049 that a single package (Template Tag) would suffice.

Related:

- https://github.com/ijlee2/sublime-syntax-definition-template-tag/pull/19
- https://github.com/ijlee2/LSP-ember/pull/5


NullVoxPopuli:
fwiw, in the tree-sitter and lezer/codemirror worlds, all languages are separated by extension. Ecmascript (a foundation!) ts, js, jsx, gjs, gts, etc (tho, lezer mistakenly built in jsx into their js). i am a bit surprised, but also slightly not, that the sublime  maintainer was surprised by split by extension.

ijlee2:
> i am a bit surprised, but also slightly not, that the sublime maintainer was surprised by split by extension.

I did see that, in VS Code, we seem to be treating things separately as Glimmer JS and Glimmer TS.

The reviewer is correct that the actual code (syntax definition) for GJS and GTS is identical.

Having two packages might make more sense if the implementation of the syntax definition differs. But, out of convenience, I had used `TypeScript.sublime-syntax` and its `ts-type-parameter-list` rule to syntax-highlight `*.gjs` files, too. We might try to use `JavaScript.sublime-syntax` and one of its rules to syntax-highlight `*.gjs` files (just in case `*.js` files are highlighted differently from `*.ts`), but I couldn't easily find the right rule to use.

Update: Pursued a two-package approach again.

